PR #21, opened 2023-06-23
Signals field (recordId)

jenningsanderson:
Moving the places update to this repo

### DIFF
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -66,6 +66,7 @@ description: Common schema definitions shared by all themes
         that referenced property will be specified in the overture list of
         approved sources from the Overture Data Working Group that contains
         the relevant metadata for that dataset including license source organization.
+        The recordid refers to the specific record within the dataset that was used.
       type: object
       required: [property, dataset]
       unevaluatedProperties: false
@@ -73,6 +74,8 @@ description: Common schema definitions shared by all themes
         property:
           type: string
         dataset:
+          type: string
+        recordid:
           type: string
     theme:
       description: Top-level Overture theme this feature belongs to
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -66,7 +66,6 @@ description: Common schema definitions shared by all themes
         that referenced property will be specified in the overture list of
         approved sources from the Overture Data Working Group that contains
         the relevant metadata for that dataset including license source organization.
-        The recordid refers to the specific record within the dataset that was used.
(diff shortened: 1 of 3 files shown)


schema/defs.yaml
```diff
@@ -66,6 +66,7 @@ description: Common schema definitions shared by all themes
         that referenced property will be specified in the overture list of
         approved sources from the Overture Data Working Group that contains
         the relevant metadata for that dataset including license source organization.
+        The recordid refers to the specific record within the dataset that was used.
```
vcschapp, line 69: It's IMO preferable to move field-specific documentation into the field itself. A `description` property could be added into the `record[I]d` property.

vcschapp (changes requested):

danbeeee (commented):
address Vic's comments, assuming I'm doing this right

schema/defs.yaml
```diff
@@ -74,6 +74,9 @@ description: Common schema definitions shared by all themes
           type: string
         dataset:
           type: string
+        recordId:
```
varapmsft, line 77: This approach would result in a lot of duplication. For example, in Admin dataset for M5 TOM is providing most of the data and one name, ESRI providing bunch of names. To populate the sources we will need to repeat the ESRI provider id dozens of times.

Another approach would be to create a separate property to list provider name to provider id mappings once.

TristanDiet-TomTom (approved):

schema/defs.yaml
vcschapp: That's a good point Vara. I'd be willing to just ship what we have to see how it goes down, and then if (as is perhaps likely) we feel things are way over-duplicated, we could move to a system with more "indirection" like you're suggesting.

vcschapp (approved):
```
                  ~.
           Ya...___|__..aab     .   .
            Y88a  Y88o  Y88a   (     )
             Y88b  Y88b  Y88b   `.oo'
             :888  :888  :888  ( (`-'
    .---.    d88P  d88P  d88P   `.`.
   / .-._)  d8P'"""|"""'-Y8P      `.`.
  ( (`._) .-.  .-. |.-.  .-.  .-.   ) )
   \ `---( O )( O )( O )( O )( O )-' /
    `.    `-'  `-'  `-'  `-'  `-'  .' CJ
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #41, opened 2023-08-14
Updated admin_level min max values

sasastanojkov:
Definitions have property minValue and maxValue (see property "version" or "linearlyReferencedPosition") while admin_level has properties "minimum" and "maximum". Changing to align with general overture defs.yaml

### DIFF
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -38,7 +38,7 @@ description: Common schema definitions shared by all themes
         release where the geometry or attributes of this feature
         changed.
       type: integer
-      minValue: 0
+      minimum: 0
       "$comment": >-
         It might be reasonable to combine "updateTime" and "version"
         in a single "updateVersion" field which gives the last Overture
@@ -87,8 +87,8 @@ description: Common schema definitions shared by all themes
         the distance along a path such as a road segment or a river
         center-line segment.
       type: number
-      minValue: 0
-      maxValue: 1
+      minimum: 0
+      maxiumum: 1
       "$comment": >-
         One possible advantage to using percentages over absolute
         distances is being able to trivially validate that the position
--- a/schema/places/place.yaml
+++ b/schema/places/place.yaml
@@ -37,8 +37,8 @@ properties:
       confidence:
         description: The confidence of the existence of the place. It's a number between 0 and 1. 0 means that we're sure that the place doesn't exist (anymore). 1 means that we're sure that the place exists. If there's no value for the confidence, it means that we don't have any confidence information.
         type: number
-        minValue: 0
-        maxValue: 1
+        minimum: 0
+        maximum: 1
       websites:
         description: The websites of the place.
         type: array
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -311,7 +311,7 @@ properties:
       prefixItems:
         - description: Number of speed units
           type: integer
-          minValue: 20
+          minimum: 20
         - description: One speed unit
           type: string
           enum: [ "km/h", "mph" ]


varapmsft (approved):

BoZivk:
I'm a bit confused with this PR: [Updated admin_level min max values by sasastanojkov · Pull Request #41 · OvertureMaps/schema (github.com)](https://github.com/OvertureMaps/schema/pull/41). The PR changes how are the min and max constraints for adminLevel expressed in json schema. Currently we're using minimum and maximum keywords, and the change is replacing them with minValue and maxValue. I'm not expert in json schema, but I know that the current schema (with minimum and maximum keywords is passing validation). Take a look at this this counterexample: [schema/counterexamples/admins/admin-lower-adminLevel-no-context.json at main · OvertureMaps/schema (github.com)](https://github.com/OvertureMaps/schema/blob/main/counterexamples/admins/admin-lower-adminLevel-no-context.json). That counterexample represents an entity with adminLevel=3 and it is missing context relationship. That is not allowed based on the current schema becasuse of this constraint:

            if: {properties: {adminLevel: { minimum: 3 }}}
            then:
              required: [context]

Counterexamples are set to test incorrect conditions, and that is why above counter example is passing validation - we expect that there is something wrong with the counterexample and since that is really the case the validation succeeds for this counterexample. However, I tried to change adminLevel in the counterexample to 2, which is now a valid condition (schema allows missing context for adminLevel=2). When I run the validation when adminLevel=2, the validation fails (because validation expects the incorrect data four counterexample but everything is actually correct).

What we can deduct from the above experiment is that minimum and maximum keywords work with the schema, at least with the current validation tool. On the other hand, when I apply this change (minimum -> minValue, maximum -> maxValue), the validation fails for this example: [schema/examples/admins/locality/admin-country-no-context.yaml at main · OvertureMaps/schema (github.com)](https://github.com/OvertureMaps/schema/blob/main/examples/admins/locality/admin-country-no-context.yaml), while the expectation is that it should succeed.

BoZivk (changes requested):
I'm a bit confused with this PR. The PR changes how are the min and max constraints for adminLevel expressed in json schema. Currently we're using minimum and maximum keywords, and the change is replacing them with minValue and maxValue. I'm not expert in json schema, but I know that the current schema (with minimum and maximum keywords is passing validation). Take a look at this this counterexample: [schema/counterexamples/admins/admin-lower-adminLevel-no-context.json at main · OvertureMaps/schema (github.com)](https://github.com/OvertureMaps/schema/blob/main/counterexamples/admins/admin-lower-adminLevel-no-context.json). That counterexample represents an entity with adminLevel=3 and it is missing context relationship. That condition is not allowed based on the current schema because of this constraint:

            if: {properties: {adminLevel: { minimum: 3 }}}
            then:
              required: [context]

Counterexamples are set to test incorrect conditions, and that is why above counter example is passing validation - we expect that there is something wrong with the counterexample and since that is really the case the validation succeeds for this counterexample. However, I tried to change adminLevel in the counterexample to 2, which is now a valid condition (schema allows missing context for adminLevel=2). When I run the validation when adminLevel=2, the validation fails (because validation expects the incorrect data for counterexample but everything is actually correct).

What we can deduct from the above experiment is that minimum and maximum keywords work with the schema, at least with the current validation tool. On the other hand, when I apply this change (minimum -> minValue, maximum -> maxValue), the validation fails for this example: [schema/examples/admins/locality/admin-country-no-context.yaml at main · OvertureMaps/schema (github.com)](https://github.com/OvertureMaps/schema/blob/main/examples/admins/locality/admin-country-no-context.yaml), while the expectation is that it should succeed.r to the conversation

varapmsft (changes requested):
Sorry, I did not notice earlier, but the validation is failing, please fix it.

BoZivk:
@varapmsft I actually thing that no change is needed at all. With disclaimer that I'm not an expert on json schema, but based on the validation results I think that the current state of the schema is correct.

vcschapp:
> @varapmsft I actually think that no change is needed at all. With disclaimer that I'm not an expert on json schema, but based on the validation results I think that the current state of the schema is correct.

I mostly agree with Bojan, but I would go a bit farther. Change is needed, but this is not the right place. According to [the JSON Schema docs](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-validation-keywords-for-num), `minimum` and `maximum` seem to be the correct validation properties. So this suggests that the problem is not here, but actually in the main `defs.yaml`...

@sasastanojkov , would you mind updating the PR accordingly, and fix the validation problems at the same time?

Ideally you should not only have all the tests passing, but also add new test cases that prevent the problem from recurring.

BoZivk:
> I mostly agree with Bojan, but I would go a bit farther. Change is needed, but this is not the right place. According to [the JSON Schema docs](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-validation-keywords-for-num), `minimum` and `maximum` seem to be the correct validation properties. So this suggests that the problem is not here, but actually in the main `defs.yaml`...

Right, I agree. I think that there are three places where minValue and maxValue are referenced: (global) defs.yaml, place.yaml and segment.yaml. I believe that in all three files they should be replaced with minimum and maximum.

vcschapp:
> > I mostly agree with Bojan, but I would go a bit farther. Change is needed, but this is not the right place. According to [the JSON Schema docs](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-validation-keywords-for-num), `minimum` and `maximum` seem to be the correct validation properties. So this suggests that the problem is not here, but actually in the main `defs.yaml`...
> 
> Right, I agree. I think that there are three places where minValue and maxValue are referenced: (global) defs.yaml, place.yaml and segment.yaml. I believe that in all three files they should be replaced with minimum and maximum.

Thanks for finding those locations! @sasastanojkov, could we use this PR as an opportunity to fix the problem everywhere?

sasastanojkov:
> > > I mostly agree with Bojan, but I would go a bit farther. Change is needed, but this is not the right place. According to [the JSON Schema docs](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-validation-keywords-for-num), `minimum` and `maximum` seem to be the correct validation properties. So this suggests that the problem is not here, but actually in the main `defs.yaml`...
> > 
> > 
> > Right, I agree. I think that there are three places where minValue and maxValue are referenced: (global) defs.yaml, place.yaml and segment.yaml. I believe that in all three files they should be replaced with minimum and maximum.
> 
> Thanks for finding those locations! @sasastanojkov, could we use this PR as an opportunity to fix the problem everywhere?

Yea sure. I personally don't care if it is minimum or minValue, I just want them to be the same everywhere. I will make changes.

BoZivk (approved):

varapmsft (approved):

vcschapp:
> Yea sure. I personally don't care if it is minimum or minValue, I just want them to be the same everywhere. I will make changes.

Consistency is a high goal, but correctness is a higher one. You should care, because one is correct and has meaning in JSON Schema, and one has no meaning and is ignored by JSON Schema!

vcschapp:
@sasastanojkov , thanks.

From a Git hygiene perspective, would you mind squashing the four commits currently attached to this PR into one commit that just describes the business value achieved, *i.e.* something along the lines of "Fix incorrect JSON Schema validation constraints"?

-----

Basically in this case there's one atomic unit of business value (fixing a particular incorrectness in the schema) and that's what we want the Git history to reflect, rather than the messy process of getting there.

vcschapp:
Thanks @sasastanojkov !

sasastanojkov:
> Thanks @sasastanojkov !

You are welcome :D

PR #42, opened 2023-08-17
Context theme

jenningsanderson:
Opening a standing PR here to get this branch started and create the docs.

### DIFF
--- a/docusaurus/sidebars.js
+++ b/docusaurus/sidebars.js
@@ -142,6 +142,18 @@ const sidebars = {
             'reference/transportation/segment',
           ]
         },
+        {
+          type: 'category',
+          label: 'context layers',
+          // link: {
+          //   type: 'generated-index',
+          // },
+          collapsed: false,
+          items: [
+              'reference/context/water',
+              // 'reference/context/land',
+          ]
+        },
       ]
     }
   ],
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 .gitignore
 .vscode
-
+.watchman*
--- a/docusaurus/docs/reference/context/water.mdx
+++ b/docusaurus/docs/reference/context/water.mdx
@@ -1,0 +1,43 @@
+---
+title: water
+---
+
+import CodeBlock from '@theme/CodeBlock';
+import JSONSchemaViewer from "@theme/JSONSchemaViewer";
+import generateResolverOptions from "@site/src/components/shared-libs/generateResolverOptions"
+import yamlLoad from "@site/src/components/yamlLoad"
+import WaterSchema from "!!raw-loader!@site/docs/_schema/context/water.yaml";
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Water
+
+Buildings are human-made structures with roofs or interior spaces that are permanently or semi-permanently in one place ([OSM building definition](https://wiki.openstreetmap.org/wiki/Key:building)).
+
+<table>
+  <tr>
+    <th>Geometry Type</th>
+    <td><code>Polygon</code> or <code>MultiPolygon</code></td>
+  </tr>
+  <tr>
+    <th>Theme</th>
+    <td><code>context</code></td>
+  </tr>
+  <tr>
+    <th>Type</th>
+    <td><code>water</code></td>
+  </tr>
+</table>
+
+
+## Schema
+
+<Tabs>
+  <TabItem value="browsable" label="Browsable" default>
+    <JSONSchemaViewer schema={ yamlLoad(WaterSchema) } resolverOptions={ generateResolverOptions({remote: true, yamlBasePath: '/context'})}/>
+  </TabItem>
+  <TabItem value="yaml" label="YAML" default>
+    <CodeBlock language="jsx">{WaterSchema}</CodeBlock>
+  </TabItem>
+</Tabs>
--- a/docusaurus/docs/themes/context/water.mdx
+++ b/docusaurus/docs/themes/context/water.mdx
@@ -1,0 +1,7 @@
+---
+title: Water
+---
+
+# Water
+
+[View the Water Schema](/reference/context/water)
--- a/docusaurus/src/YAML_FILE_TREE.js
+++ b/docusaurus/src/YAML_FILE_TREE.js
@@ -13,6 +13,8 @@ const _default = {
     '/transportation/connector.yaml': require('!!raw-loader!@site/docs/_schema/transportation/connector.yaml'),
     '/transportation/segment.yaml': require('!!raw-loader!@site/docs/_schema/transportation/segment.yaml'),
 
+    '/context/water.yaml': require('!!raw-loader!@site/docs/_schema/context/water.yaml'),
+
     '/defs.yaml': require('!!raw-loader!@site/docs/_schema/defs.yaml'),
     '/schema.yaml': require('!!raw-loader!@site/docs/_schema/schema.yaml')
 
--- a/schema/context/water.yaml
+++ b/schema/context/water.yaml
@@ -1,0 +1,40 @@
+---
+"$schema": https://json-schema.org/draft/2020-12/schema
+title: Water Schema
+description: Bodies of water including oceans, lakes, and rivers
+type: object
+properties:
+  geometry:
+    description: >-
+      A regular building's geometry is defined as it's footprint or roofprint (if traced from aerial/satellite imagery). These MUST be a Polygon as defined by GeoJSON
+      schema.
+    unevaluatedProperties: false
+    oneOf:
+      - "$ref": https://geojson.org/schema/LineString.json
+      - "$ref": https://geojson.org/schema/MultiLineString.json
+      - "$ref": https://geojson.org/schema/Polygon.json
+      - "$ref": https://geojson.org/schema/MultiPolygon.json
+  properties:
+    unevaluatedProperties: false
+    allOf:
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/levelContainer
+    properties:
+      names: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/names" }
+      subType:
+        description: >-
+          A broad category of the building type / purpose. When the current use of the building does not match the built purpose, the class should
+          be set to represent the current use of the building.
+        type: string
+        enum:
+          - ocean
+          - sea
+          - lake
+          - pond
+          - river
+          - stream
+          - human_made
+        salt:
+        description: Is it salt water or not
+        type: boolean
+        exclusiveMinimum: 0
--- a/schema/context/water.yaml
+++ b/schema/context/water.yaml
@@ -34,7 +34,7 @@ properties:
           - river
           - stream
           - human_made
-        salt:
+      salt:
         description: Is it salt water or not
         type: boolean
         exclusiveMinimum: 0


vcschapp:
Stopping by to say that I love these standing-PRs-with-docs!

docusaurus/docs/themes/base/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+title: Base
+---
+
+# Base Theme
+
+Features from OpenStreetMap that provide additional context such as **land** and **water** that let users render a more complete basemap.
+
+## Key Schema Design Choices
+
+1. Basic classificaiton of features: type, class, and subclass where appropriate.
```
batinic-m, line 11: Typo.

vcschapp (changes requested):
I'm excited about this new theme!!

Thoughts and comments below.

schema/base/land.yaml
```diff
@@ -0,0 +1,74 @@
...
+    allOf:
+      - "$ref": ./defs.yaml#/$defs/propertyContainers/osmPropertiesContainer
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/levelContainer
+    required:
+      - class
+      - subclass
+    properties:
+      class:
+        description: Broad class of landcover, such as forest, glacier, grass, or a physical feature.
```
AleksandraDebicka-TomTom, line 26: What is an example of physical feature?

schema/base/land.yaml
```diff
@@ -0,0 +1,74 @@
...
+           - grass
+           - land
+           - physical
+           - reef
+           - rock
+           - sand
+           - shrub
+           - tree
+           - wetland
+      subclass:
```
AleksandraDebicka-TomTom, line 41: Not sure how accurate we want to be in validation at this stage of Base Theme introduction (but our schema is about providing some guarantee in space of properties validity vs osm). Some of subclass values have a sense only for specific classes - no?. It could be then beneficial to have schema for subclass constructed through conditional values of subclass depending on value of class.

docusaurus/sidebars.js
```diff
@@ -55,6 +55,7 @@ const sidebars = {
           label: 'Admins'
         },
         'themes/buildings/building',
+        'themes/base/index',
```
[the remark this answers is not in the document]
jenningsanderson: Yes it should... I had renamed it from `context` :)

examples/base/water-river-with-wikidata.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - common:
+      - value: Rio Santana
+        language: local
+  sourceTags:
+    waterway: river
+  sources:
+    - recordID: w353912370@4
+      property: ""
+      dataset: OpenStreetMap
+  wikidata: Q7420045
```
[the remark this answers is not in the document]
jenningsanderson: This is a GREAT idea @vcschapp that I agree, we should surface.  We do have the recordId in sources, but that is _distinctly different_.

schema/base/land.yaml
jenningsanderson: Like a mountain peak from OSM, which I've added

schema/base/defs.yaml
```diff
@@ -0,0 +1,20 @@
...
+title: Context layer properties
+description: Common schema definitions the context theme (primarily from OSM)
+"$defs":
+  propertyDefinitions:
+    wikidata:
+      description: A wikidata ID if available
+      type: string
+    sourceTags:
+      description: Any attributes/tags from the original source data that should be passed through.
+      type: object
```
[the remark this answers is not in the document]
jenningsanderson: I agree, let's kick off a much larger discussion about this.

Keeping this in the `base` theme (as it is), it _more or less_ implies `OSMTags` because we don't see other data sources coming into the base theme.

@jwass had a good point that if we elevate `sourceTags` to an Overture level property, we kind of permit and encourage all tags to pass through, and the defined schema loses value. 

To that end, we always have the `ext` option, so for buildings, for example, if we wanted to pass some source tags from OSM (such as addr: tags), we could always include them as an `extSourceTags`. 

We do have the ability in `sources` to denote specific sources for source tags, should that need come up, but I find myself in agreement that if we get into the position where we're conflating source tags within a specific theme, then the content of the tags is likely important enough that whatever info we are conflating should be it's own top-level attribute.

If we needed to track multiple sources through source tags, we can:
```json
[{"property":"properties/extSourceTags", "dataset":"OpenStreetMap"}]

[
  {"property":"properties/extSourceTags/addr:street", "dataset":"OpenStreetMap"},
  {"property":"properties/extSourceTags/material", "dataset":"Esri"}
]

```

schema/base/landuse.yaml
```diff
@@ -0,0 +1,159 @@
...
+        - landfill
+        - medical
+        - military
+        - park
+        - public
+        - protected
+        - recreation
+        - religious
+        - residential
+        - resource_extraction
```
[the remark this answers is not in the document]
jenningsanderson: The majority of these values are all pass through from OSM `landuse`= or `natural`= tags, which are snake cased. I think there is value in leaving them untouched from the source, but I could be convinced otherwise.

examples/base/land-sand-example.yaml
```diff
@@ -0,0 +1,23 @@
+---
+type: Feature
+geometry:
+  type: Polygon
+  coordinates: [[[34.4379197, 28.7592689], [34.4380109, 28.7595816], [34.4380173, 28.7601587], [34.4380538, 28.7606186],[34.4381540, 28.7609863], [34.4380735, 28.7612459], [34.4377346, 28.7608396], [34.4377158, 28.7606139],[34.4377105, 28.7605527], [34.4376300, 28.7604258], [34.4376407, 28.7603599], [34.4376488, 28.7602494],[34.4376863, 28.7599673], [34.4376675, 28.7597980], [34.4376890, 28.7596216], [34.4379197, 28.7592689]]]
+properties:
+  theme: base
+  type: land
+  class: sand
+  subclass: dune
```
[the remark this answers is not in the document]
jwass: @vcschapp We need to be careful here. "type" is used as a partition for the Parquet data. I wouldn't think we want to partition landuse or buildings based on their class tags.

examples/base/water-river-with-wikidata.yaml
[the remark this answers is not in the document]
jwass: This is really a much broader GERS discussion in my opinion. When you figure out two different datasets contain the same object, how to reflect that? Should this field always contain every possible dataset that can match?

examples/base/land-sand-example.yaml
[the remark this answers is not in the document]
jwass: I also think there's an argument that subclass is a single word and doesn't need camel case: https://www.merriam-webster.com/dictionary/subclass

schema/base/land.yaml
jwass: My guess is it's tough to get in for the upcoming release but I think this is right since some subclasses should only work with some classes and the schema should capture that.

docusaurus/docs/themes/base/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+1. Basic classification of features into type, class, and subclass where appropriate.
+2. Parsing and normalization of common OSM tags, such as `height` and `ele`.
+4. Extraction of consistent OSM tags, such as wikidata to top-level properties.
+3. Direct pass-thru of remaining relevant OSM tags.
+
+## Current Types:
+
+- [land](/reference/base/landuse)
+- [landuse](/reference/base/landuse)
```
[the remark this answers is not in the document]
vcschapp: > I wouldn't oppose making it camel case

The schema standard is camel case.

schema/base/landuse.yaml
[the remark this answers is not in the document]
vcschapp: The schema standard is camel case. We're also already translating all the snake case things like this to camel case in transportation. This introduces a conflict and inconsistency....

Ideally the schema will present to Overture consumers as one consistent whole that's easy to understand, so IMO we need to resolve the conflict and inconsistency....

examples/base/water-river-with-wikidata.yaml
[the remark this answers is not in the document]
vcschapp: Love the discussion. We do have the option of holding this idea back until we have stabilized the GERS discussion.

examples/base/land-sand-example.yaml
```diff
@@ -0,0 +1,23 @@
...
+geometry:
+  type: Polygon
+  coordinates: [[[34.4379197, 28.7592689], [34.4380109, 28.7595816], [34.4380173, 28.7601587], [34.4380538, 28.7606186],[34.4381540, 28.7609863], [34.4380735, 28.7612459], [34.4377346, 28.7608396], [34.4377158, 28.7606139],[34.4377105, 28.7605527], [34.4376300, 28.7604258], [34.4376407, 28.7603599], [34.4376488, 28.7602494],[34.4376863, 28.7599673], [34.4376675, 28.7597980], [34.4376890, 28.7596216], [34.4379197, 28.7592689]]]
+properties:
+  theme: base
+  type: land
+  class: sand
+  subClass: dune
+  names:
+    - common:
```
ibnt1, line 12: this '-' should be removed, `names` is object, not array (applies to the other two examples as well)

schema/base/defs.yaml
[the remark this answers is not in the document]
vcschapp: I agree that the `sources` property schema is well-designed and resilient to source tracking as you illustrated.

My concern is more subtle, and has to do with tag name collisions.

Let's say you have two sources, OSM and Who Knows What, they are both sources used for the feature, and they both have a `"foo"` tag. In my mind the semantics of `sourceTags` implies that both the tags from OSM and the tags from Who Knows What would be preserved given they are both sources. But with the schema as written, only one of them can "win" and be chosen.

Given that the whole point of `sourceTags` is to pass through tag values that Overture doesn't really attach meaning to, this is a weird and unexpected outcome. If Overture has no opinion on whether OSM's `foo` tag is more important or less important than Who Knows What's `foo` tag, how can it choose between them? This is why I think some kind of system for making clear that the tags are passed through from OpenStreetMap is probably desirable.

------

We can certainly do it, but I don't love the idea of our own data release including extension properties. If it's important enough to include, it should be important enough to model in the schema, especially since people will, with absolute certainty, start to depend on it like it's part of the core schema. To my mind extension properties should only be to support extension of the core schema by non-Overture actors.

examples/base/land-sand-example.yaml
```diff
@@ -0,0 +1,23 @@
+---
+type: Feature
+geometry:
+  type: Polygon
+  coordinates: [[[34.4379197, 28.7592689], [34.4380109, 28.7595816], [34.4380173, 28.7601587], [34.4380538, 28.7606186],[34.4381540, 28.7609863], [34.4380735, 28.7612459], [34.4377346, 28.7608396], [34.4377158, 28.7606139],[34.4377105, 28.7605527], [34.4376300, 28.7604258], [34.4376407, 28.7603599], [34.4376488, 28.7602494],[34.4376863, 28.7599673], [34.4376675, 28.7597980], [34.4376890, 28.7596216], [34.4379197, 28.7592689]]]
+properties:
+  theme: base
+  type: land
+  class: sand
+  subClass: dune
```
vcschapp, line 10: One of my more important comments from last week seems to have gotten clobbered, or at any rate I can't find it anymore.

So far the schema property "hierarchy" implied by the other layers is: `type > subType > class`. Here we have `type > class > subClass`.

There's an exception to this in `buildings` where we go immediately to `type > class` but I think we should strive to normalize this so that customers and consumers can look at and query a predictable set of columns rather than an *ad hoc* set that varies by theme and feature type.

My solution would be:

- Transportation: Promote `road.class` up to main feature properties.
- Buildings: Rename `class` to `subType`.
- Base: Rename `class` to `subType` and `subClass` to `class`.

Obviously this will be a bit less aligned with the Earth tables than previously, but Overture is a different product and we want to present a coherent, structured, internally consistent picture across all the themes and types.

vcschapp (changes requested):

schema/base/defs.yaml
```diff
@@ -0,0 +1,21 @@
...
+"$schema": https://json-schema.org/draft/2020-12/schema
+title: Base layer properties
+description: Common schema definitions the base theme (primarily from OSM)
+"$defs":
+  propertyDefinitions:
+    wikidata:
+      description: A wikidata ID if available
+      type: string
+      pattern: ^Q\d+
+    sourceTags:
```
ibnt1, line 11: this is the first property that has type `object` but no fields explicit defined in the schema. i assume parquet type for this would be map<string, string>?

jenningsanderson: Yes

examples/base/land-sand-example.yaml
jenningsanderson: @vcschapp I mostly agree with you, but there's something about the hierarchy that feels off to me for some things, check this out: 

1. type=water, subType = 'manMade', class= 'pool' ✅ 
2. type=land, subType = 'grass', class='grass' ✅ 
3. type=landuse, subType = 'agriculture', class='farmland' ✅ 
4. type=building, subType = 'residential' ?, class=None ?? 

I think the class -> subType makes sense for the base theme, but the building classes (and this opens up some of the original buildings argument) is that `class` applies to the _use_ of the building, not a subType of the building. I could imagine `subType` being "shed" or "garage" or something, but `class` is about the current use.

schema/base/landuse.yaml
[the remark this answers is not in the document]
jenningsanderson: fair

examples/base/water-river-with-wikidata.yaml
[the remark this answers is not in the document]
jenningsanderson: Yes. 
Proposal: Keep it as top-level feature in theme=base for now, and revisit later when we have a clear understanding of how Wikidata IDs and GERS IDs might be interconnected? (Because they should be to a degree).

schema/base/defs.yaml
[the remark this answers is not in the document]
jenningsanderson: Proposal for M1: Keep this as `sourceTags` for now and it's only present in this theme.

We can add: "Elevate Wikidata and sourceTags" to the list of schema improvements for consideration in December?

docusaurus/docs/themes/base/index.mdx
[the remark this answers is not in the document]
vcschapp: I'm not sure if @RobSoetewey-TomTom had a chance to post his comment, but I think he has a comment to the effect that, agreeing with @jwass, "land use" is a well-established and well-understood term.

Rob and Jake's reasoning makes sense to me.

In that case, let's go with `landUse` in the camelCase if that makes sense to @jenningsanderson also.

examples/base/land-sand-example.yaml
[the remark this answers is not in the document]
vcschapp: > @vcschapp We need to be careful here. "type" is used as a partition for the Parquet data. I wouldn't think we want to partition landuse or buildings based on their class tags.

This sounds important. Why not?

examples/base/land-sand-example.yaml
vcschapp: > I think the class -> subType makes sense for the base theme, but the building classes (and this opens up some of the original buildings argument) is that `class` applies to the *use* of the building, not a subType of the building.

Isn't this mostly a question of definitions?

The interesting thing is if you just take the words *type* and *class* in the abstract, I think most people would have a hard time differentiating between them. They're more or less synonyms. (I guess what I'm saying could be an argument for not allowing both `class` and `type`/`subType` in the top-level schema.)

I think if we define `subType` for `building` to mean current use, and make that clear, people would understand.

examples/base/land-sand-example.yaml
```diff
@@ -0,0 +1,23 @@
...
+  subclass: dune
+  names:
+    common:
+      value: Hadeida
+      language: local
+  sourceTags:
+    natural: dune
+    surface: sand
+  sources:
+    - recordID: w407753930@3
```
vcschapp, line 19: `recordID` should be `recordId`...

Might be an error in the example, or an error in the schema...

Is it possible to fix the schema (if needed) and the example?

examples/base/water-river-with-wikidata.yaml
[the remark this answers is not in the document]
vcschapp: Talked about this in Schema TF meeting today, memorialized at [Schema WG issue 196](https://github.com/OvertureMaps/schema-wg/issues/196).

We agreed the `externalIds` bag-of-IDs property has merit but we lack time to align on it and implement it for now, so the outcome was:

1. For the future [Schema Roadmap](https://wiki.overturemaps.org/x/BQEjAQ), the`externalIds` property is added for later prioritization and discusison.
2. For the current PR (see Rob's notes and minutes from issue 196):
    1. [ ] `wikidata` should remain a top-level property
    2. [ ] But shouldn't be linked to OSM.
    3. [ ] As part of this PR, a shared `wikidata` schema definition should be factored out and placed in `schema/defs.yaml` and reused both in:
       - [ ] here, in the Base theme; and
       - [ ] retro-fitted into the Places theme's `brand` object.

examples/base/land-sand-example.yaml
jenningsanderson: Nope that was me, I fixed them in the examples.

examples/base/land-sand-example.yaml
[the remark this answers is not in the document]
jenningsanderson: To be clear, the change is from `type/class/subclass` --> `type/subType/class`, so type remains unchanged and valid for the partitioning. We aren't partitioning on subType

examples/base/land-sand-example.yaml
```diff
@@ -0,0 +1,23 @@
...
+  type: Polygon
+  coordinates: [[[34.4379197, 28.7592689], [34.4380109, 28.7595816], [34.4380173, 28.7601587], [34.4380538, 28.7606186],[34.4381540, 28.7609863], [34.4380735, 28.7612459], [34.4377346, 28.7608396], [34.4377158, 28.7606139],[34.4377105, 28.7605527], [34.4376300, 28.7604258], [34.4376407, 28.7603599], [34.4376488, 28.7602494],[34.4376863, 28.7599673], [34.4376675, 28.7597980], [34.4376890, 28.7596216], [34.4379197, 28.7592689]]]
+properties:
+  theme: base
+  type: land
+  subType: sand
+  class: dune
+  names:
+    common:
+      value: Hadeida
```
vcschapp, line 13: This could be causing a problem for validation. Below common the `value`/`language` pairs should be a list.

schema/base/landuse.yaml
```diff
@@ -0,0 +1,159 @@
+---
```
vcschapp, line 1: Looks like the current diff has both `landuse.yaml` *and* `landUse.yaml`?

schema/schema.yaml
```diff
@@ -58,3 +58,27 @@ oneOf:
...
+            theme: { enum: [base] }
+            type: { enum: [land] }
+    then: { "$ref": base/land.yaml }
+    else: { propertyNames: false }
+  - if:
+      properties:
+        properties:
+          properties:
+            theme: { enum: [base] }
+            type: { enum: [landuse] }
```
vcschapp, line 74: Maybe needs to be `landUse`?

schema/base/landUse.yaml
```diff
@@ -0,0 +1,159 @@
...
+          - vineyard
+          - waterHazard
+          - waterPark
+          - wildernessArea
+          - winterSports
+          - zoo
+      names: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/names" }
+      surface:
+        description: Surface material
+        type: string
```
vcschapp, line 159: Is there any enumeration or validation here?

examples/base/water-river-example.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - value: Root River
+        language: local
+  sourceTags:
+    waterway: river
+  sources:
+    - recordId: w54320702@1
+      property: ""
+      dataset: OpenStreetMap
+  isSalt: false
+  isIntermittent: false
```
vcschapp, line 22: For post M1, we could consider following the transportation segment model and having a flags array here. Would keep the column count down a bit.

schema/base/defs.yaml
```diff
@@ -0,0 +1,17 @@
...
+      description: Any attributes/tags from the original source data that should be passed through.
+      type: object
+  propertyContainers:
+    osmPropertiesContainer:
+      title: "OSM Properties"
+      description: OSM Properties
+      type: object
+      properties:
+        sourceTags: { "$ref": "#/$defs/propertyDefinitions/sourceTags" }
+        wikidata: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/wikidata" }
```
vcschapp, line 17: Move `wikidata` outside of `osmPropertiesContainer`?

vcschapp (commented):
```
                  ~.
           Ya...___|__..aab     .   .
            Y88a  Y88o  Y88a   (     )
             Y88b  Y88b  Y88b   `.oo'
             :888  :888  :888  ( (`-'
    .---.    d88P  d88P  d88P   `.`.
   / .-._)  d8P'"""|"""'-Y8P      `.`.
  ( (`._) .-.  .-. |.-.  .-.  .-.   ) )
   \ `---( O )( O )( O )( O )( O )-' /
    `.    `-'  `-'  `-'  `-'  `-'  .' CJ
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

vcschapp (approved):

docusaurus/docs/themes/base/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+title: Base
+---
+
+# Base Theme
+
+The Overture Base Theme provides additional context such as land and water that are critical to rendering a complete basemap. These features are currently derived from the [Daylight Earth Tables](https://daylightmap.org/earth/) schema and include the [Daylight Coastlines](https://daylightmap.org/coastlines.html).
+
+## Key Schema Design Choices
+
+1. Basic classification of features into type, class, and subclass where appropriate.
```
DavidKarlas, line 11: Probably want to update to type, subtype and class.

schema/base/landUse.yaml
```diff
@@ -0,0 +1,159 @@
...
+          - villageGreen
+          - vineyard
+          - waterHazard
+          - waterPark
+          - wildernessArea
+          - winterSports
+          - zoo
+      names: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/names" }
+      surface:
+        description: Surface material
```
DavidKarlas, line 158: Wondering if this should mention if it is coming directly from some OSM tag as-is, otherwise it should probably have enum.

schema/base/land.yaml
```diff
@@ -0,0 +1,74 @@
+---
+"$schema": https://json-schema.org/draft/2020-12/schema
+title: Land Schema
+description: Land
```
DavidKarlas, line 4: I wonder if there would be added value in copy-pasting texts from docusaurus/docs/themes/base/index.mdx#Current Types to land.yaml, water.yaml and landUse.yaml?

DavidKarlas (approved):

schema/base/landUse.yaml
jenningsanderson: Good point, we do _some_ normalization to this tag.

schema/base/land.yaml
jenningsanderson: Good call, done.

RobSoetewey-TomTom (approved):

PR #44, opened 2023-08-22
confidence proposal for buildings theme

jenningsanderson:
The Google open buildings dataset has a `confidence` score that should likely be included in the next data release.

Creating a `confidence` value at the top level in this theme matches the `places` theme and gives us flexibility to include a sort of rating for all future ml-derived data in the buildings theme.

Further heuristics can be documented making this a valuable attribute, for example, ml-derived data is always < 0.85, OSM and other "human-validated" sources could always be > 0.85, etc. etc. 

While the ms-ml buildings don't have a `confidence` score, we could introduce some basic metrics: buildings with (very) large areas get lower confidence, buildings with many vertices and sharp angles could also get lower confidence.

GERS-based conflation could also leverage this `confidence` attribute downstream.


jenningsanderson:
https://dfhx9f55j8eg5.cloudfront.net/pr/44/reference/buildings/building

jenningsanderson:
Other considerations: 
1. If `confidence` exists on within the `sources`, then it's just _more difficult_ to use, but it might make more sense for some types of features?

skmoore:
Also, see this note from the [Google buildings FAQ](https://sites.research.google/open-buildings/#faq):
Screenshot 2023-08-23 at 10 20 46 AM
 
 Maybe I'm misinterpreting it but from that note it seems like a .9 confidence score doesn't mean the same thing even across the whole Google dataset.

PR #51, opened 2023-09-18
Introduce building parts

jwass:
Introduce the building part schema. Parts and main footprints share shape information.


jwass:
Still todo: 
- More/better examples
- Is the building reference from part to building a string or something else?
- Should the `hasParts` boolean be on the footprint if it can be determined by inspecting the data?

jenningsanderson:
View the change in the documentation: https://dfhx9f55j8eg5.cloudfront.net/pr/51/reference/buildings/building

skmoore:
@jwass Looking at the PR, I can't quite tell where the building ID that the part belongs to should go. Is it just an attribute underneath "properties" like this?

"properties": {
		"building": "1234",
		"extFoo": "I am a customer user property.",
		"extBar": "Me too!",
		"theme": "buildings",
		"type": "part",
		"version": 1,
		"level": 1,
		"updateTime": "2023-06-06T10:30:00-08:00",
		"numFloors": 8,
		"sources": [{
			"property": "",
			"dataset": "OpenStreetMap"
		}]
	}

skmoore:
@jwass Is the PR updated to show where the building ID value should go? Not sure if I missed it somewhere.

jwass:
> @jwass Is the PR updated to show where the building ID value should go? Not sure if I missed it somewhere.

@skmoore Yes I just made the change in 051d72f0df0fea672fd2dc3ec1b0edfda08407a7. Good catch - I made it so the building id is now a required field and the example now contains the `building` field

skmoore:
@jwass How should I submit the sample data I put together?

jwass:
> @jwass How should I submit the sample data I put together?

@skmoore Maybe make a new directory for each one under the examples/buildings directory in the repo?

examples/buildings/osm/outline.json
```diff
@@ -0,0 +1,70 @@
...
+				"value": "Valentina by Alta",
+				"language": "local"
+			}]
+		},
+		"numFloors": 8,
+		"class": "commercial",
+		"sources": [{
+			"property": "",
+			"dataset": "OpenStreetMap"
+		}]
```
jwass, line 68: Add hasParts set to True

jwass:
Outstanding issue: https://github.com/OvertureMaps/tf-buildings/issues/24 before merging

LukaszMichalski-TomTom:
Should we include `roof:material` ([link](https://wiki.openstreetmap.org/wiki/Key:roof:material))  and `building:material`([link](https://wiki.openstreetmap.org/wiki/Key:building:material)) from [OSM Simple 3D Buildings](https://wiki.openstreetmap.org/wiki/Simple_3D_Buildings)?

jwass:
> Should we include roof:material ([link](https://wiki.openstreetmap.org/wiki/Key:roof:material)) and building:material([link](https://wiki.openstreetmap.org/wiki/Key:building:material)) from [OSM Simple 3D Buildings](https://wiki.openstreetmap.org/wiki/Simple_3D_Buildings)?

@LukaszMichalski-TomTom I think we agreed at the meeting this is a good idea. Want to take a stab at adding it here?

LukaszMichalski-TomTom:
@jwass I don't have permission to commit to this repo. I prepared the change, but I cannot push it.

jwass:
> @jwass I don't have permission to commit to this repo. I prepared the change, but I cannot push it.

Can you make a pull request to this branch? Or can't even do that?

jwass:
> @jwass I don't have permission to commit to this repo. I prepared the change, but I cannot push it.

Can you try again? You should have permissions now

schema/buildings/defs.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - flat
+        - gable
+        - gambrel
+        - half-hip
+        - hip
+        - mansard
+        - onion
+        - pyramid
+        - round
+        - spherical
```
LukaszMichalski-TomTom, line 37: What is the difference between `spherical` and `dome`? In osm there are only 25 instances of `roof:shape=spherical` and I think they are incorrectly tagged and can be replaced with `dome`.  Example: [way:945021143](https://www.openstreetmap.org/way/945021143)

varapmsft (approved):

mojodna (approved):

schema/buildings/defs.yaml
jwass: As discussed in the meeting, dome is something that is half a sphere. But sphere is needed to model something like Epcot or the new Sphere in Vegas. So we decided to keep it.

jenningsanderson (approved):

LukaszMichalski-TomTom (approved):

PR #52, opened 2023-09-25
Users/ibantea/transportation refactor

ibnt1:
Draft proposal to have static data types. Intended to benefit:
- can have corresponding parquet data structures and can automatically derive a parquet data types for all properties in json schema, including complex nested properties; 
- data structures that are easier to understand and consume programmatically.

Investigating downside that a strongly type parquet schema for nested complex properties is not being converted properly by ogr2ogr.


vcschapp:
We're currently doing the work contemplated in this PR, so I think it's safe to close this one.

PR #56, opened 2023-10-04
Update turn restriction schema as per the discussion in (https://gith…

sanjibdutta1:
…ub.com/OvertureMaps/schema-wg/issues/131)

Other impact:
Renamed direction container to heading container. As a result access restriction need to be heading: instead of direction: However, I guess this should be changed to rules container instead so that it becomes when {heading: forward} instead of just heading:forward.

Removed during container from rules container. There is no reason for during container to be part of rule container.

Next Step: Fix combobulator to take of these changes

---

**UPDATE** (from Vic): This is the staging docs website for this pull request: https://dfhx9f55j8eg5.cloudfront.net/pr/56.

### DIFF
--- a/examples/transportation/docusaurus/turn-restriction-01-source.yaml
+++ b/examples/transportation/docusaurus/turn-restriction-01-source.yaml
@@ -23,4 +23,4 @@ properties:
            - segmentId: overture:transportation:example:simple-turn-restriction-target
              connectorId: overture:transportation:example:simple-turn-restriction-connector2
            finalHeading: forward
-           when: {heading : forward} 
+           when: {heading: forward} 
--- a/examples/transportation/docusaurus/turn-restriction-01-source.yaml
+++ b/examples/transportation/docusaurus/turn-restriction-01-source.yaml
@@ -18,8 +18,9 @@ properties:
     - overture:transportation:example:simple-turn-restriction-connector2
   road:
     restrictions:
-      turns:
-        - segmentId: overture:transportation:example:simple-turn-restriction-target
-          connectorId: overture:transportation:example:simple-turn-restriction-connector2
-          direction: forward
-          toDirection: forward
+      prohibitedTransitions:
+         - sequence:  
+           - segmentId: overture:transportation:example:simple-turn-restriction-target
+             connectorId: overture:transportation:example:simple-turn-restriction-connector2
+           finalHeading: forward
+           when: {heading : forward} 
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -226,12 +226,12 @@ properties:
           "$comment": >-
             Restrictions define two kinds of rerstriction on this segemnt
             Travel Restrictions defines how certain modes of transportation
-            may be limited on this segement, for example trucks are not
+            may be limited on this segment, for example trucks are not
             permitted from 7AM to 7PM
 
             Turn restrictions defines restrcited tunrs that orignates from
-            this segement, for example against driving side turn from this
-            segement to another connected segement is blocked.
+            this segment, for example against driving side turn from this
+            segment to another connected segment is blocked.
           type: object
           unevaluatedProperties: false
           properties:
@@ -258,15 +258,15 @@ properties:
                      default:
                        enum: [legal]
                    sequence:
-                     description: Sequence of ordered segements that needs to be traversed in order for the relation to be effective
+                     description: Ordered sequence of segments that needs to be traversed in order for the relation to be effective
                      type: array
                      items:
-                       description: GERS ID of the connected segments and the relevent connector
+                       description: Pair of segment and connector IDs along the sequence 
                        "$ref": "#/$defs/propertyDefinitions/relationTarget"
                      minItems: 1
                      uniqueItems: true
                    finalHeading:
-                     description: direction of travel on the target that pertains to the relation
+                     description: Direction of travel that is prohibited on the destination segment of the sequence.
                      "$ref": "#/$defs/propertyDefinitions/direction"
 
     roadFlag:
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -237,15 +237,14 @@ properties:
           properties:
             speedLimits: { "$ref": "#/$defs/propertyContainers/speedLimitsContainer" }
             access: { "$ref": "#/$defs/propertyContainers/accessContainer" }
-            turns:
+            prohibitedTransitions:
               description: Turn restriction from this segment to other segments.
               type: array
               items:
-                 allOf:
-                   - "$ref": "#/$defs/propertyContainers/modesContainer"
-                   - "$ref": "#/$defs/propertyContainers/duringContainer"
-                   - "$ref": "#/$defs/propertyDefinitions/relationTarget"
                  unevaluatedProperties: false
+                 allOf:
+                   - "$ref": "#/$defs/propertyContainers/ruleContainer"
+                   - "$ref": "#/$defs/propertyContainers/duringContainer"
                  type: object
                  properties:
                    reason:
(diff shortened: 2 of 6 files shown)


vcschapp:
Can you do a quick pass over the Docusaurus section on transition restrictions?

I think you can make just make some very limited point edits, like a word here and a word there, to bring it in line with the schema changes.

Screenshot of Docusaurus section on turn restrictions

examples/transportation/docusaurus/turn-restriction-01-source.yaml
```diff
@@ -18,8 +18,9 @@ properties:
...
-        - segmentId: overture:transportation:example:simple-turn-restriction-target
-          connectorId: overture:transportation:example:simple-turn-restriction-connector2
-          direction: forward
-          toDirection: forward
+      prohibitedTransitions:
+         - sequence:  
+           - segmentId: overture:transportation:example:simple-turn-restriction-target
+             connectorId: overture:transportation:example:simple-turn-restriction-connector2
+           finalHeading: forward
+           when: {heading : forward}
```
vcschapp, line 26: Nit but the space before the colon (` :`) should be deleted to keep roughly the same style throughout.

(Also it will make it consistent with the below `turn-restriction-02-source.yaml` at line 38.)

schema/transportation/segment.yaml
```diff
@@ -258,17 +257,18 @@ properties:
...
-                   direction:
-                     description: direction of travel on the source that pertains to the relation
-                     "$ref": "#/$defs/propertyDefinitions/direction"
-                   via:
+                   sequence:
                      description: Sequence of ordered segements that needs to be traversed in order for the relation to be effective
                      type: array
                      items:
-                       description: GERS ID of the connected segments
+                       description: GERS ID of the connected segments and the relevent connector
```
vcschapp, line 264: nit: the last vowel in relev*a*nt is an *a* not an *e*.

----

I think that where we're landing on relationships is that this ID will not be the GERS ID, it will be the "technical" or "feature" ID.

I would be tempted to just phrase this as:

> Pair of IDs for the segment being traversed and the connector used to access that segment

schema/transportation/segment.yaml
```diff
@@ -258,17 +257,18 @@ properties:
                        - gatedKey
                      default:
                        enum: [legal]
-                   direction:
-                     description: direction of travel on the source that pertains to the relation
-                     "$ref": "#/$defs/propertyDefinitions/direction"
-                   via:
+                   sequence:
                      description: Sequence of ordered segements that needs to be traversed in order for the relation to be effective
```
vcschapp, line 261: nits:
1. Segment has two e's, not three. (Not seg*e*ment...)
2. It's an *ordered sequence* of segments, not a sequence of *ordered segments*.

----
I would be tempted to write this as:

> Ordered sequence of segments that may not be traversed, where the last segment in the list is the ultimate destination of the transition.

schema/transportation/segment.yaml
```diff
@@ -258,17 +257,18 @@ properties:
...
                      type: array
                      items:
-                       description: GERS ID of the connected segments
+                       description: GERS ID of the connected segments and the relevent connector
                        "$ref": "#/$defs/propertyDefinitions/relationTarget"
                      minItems: 1
                      uniqueItems: true
+                   finalHeading:
+                     description: direction of travel on the target that pertains to the relation
```
vcschapp, line 269: Reword as?

> Direction of travel that is prohibited on the destination segment of the sequence.

schema/transportation/segment.yaml
```diff
@@ -258,17 +257,18 @@ properties:
...
+                     "$ref": "#/$defs/propertyDefinitions/direction"
```
vcschapp, line 270: Shouldn't this type be renamed from `direction` to `heading`?

Also, `direction` is [currently defined to contain](https://github.com/OvertureMaps/schema/blob/3a5e5b62e43a19ae1d24b9cddb3af149003295f7/schema/transportation/segment.yaml#L96) `forward`, `backward`, and `both`. Should the `heading` type just be `forward` and `backward` and should we drop the `both`? Or do we need `both`? Maybe we do...?

vcschapp:
We should rename the `relationTarget` type to something more directly tied to what it is used for and update the documentation there. Although it doesn't surface in the data, it makes it easier to understand the schema itself.

I suggest `transitionSequenceStep` instead of `relationTarget`.

Suggesting the following documentation changes for `transitionSequenceStep`:

```yaml
segmentId:
  description: >-
    Segment ID of the segment being traveled on in this step.
  type: string
connectorId:
  description: Connector ID of the connector used to access the segment for this step.
  "$comment": >-
    A source segment can connect to another segment at multiple connectors.
    Therefore, for each step in a transition across segments, an explicit
    connector is necessary to uniquely identify the point of transition.
```

vcschapp (changes requested):
Found a few small issues here and there.

examples/transportation/docusaurus/turn-restriction-01-source.yaml
sanjibdutta1: Agree

schema/transportation/segment.yaml
sanjibdutta1: Interesting thing about segement - found many instances and not all of them are mine. I guess it has something to do with muscle memory - g followed by e :)

schema/transportation/segment.yaml
sanjibdutta1: We do not have a heading enum, we have heading container, which hosts the direction enum. finalHeading is of type direction because otherwise it becomes heading: <direction>

sanjibdutta1:
> Can you do a quick pass over the Docusaurus section on transition restrictions?
> 
> I think you can make just make some very limited point edits, like a word here and a word there, to bring it in line with the schema changes.
> 
> Screenshot of Docusaurus section on turn restrictions

I can make changes but I have no way to test it. But since the changes are benign in nature, I guess its ok

sanjibdutta1:
> We should rename the `relationTarget` type to something more directly tied to what it is used for and update the documentation there. Although it doesn't surface in the data, it makes it easier to understand the schema itself.
> 
> I suggest `transitionSequenceStep` instead of `relationTarget`.
> 
> Suggesting the following documentation changes for `transitionSequenceStep`:
> 
> ```yaml
> segmentId:
>   description: >-
>     Segment ID of the segment being traveled on in this step.
>   type: string
> connectorId:
>   description: Connector ID of the connector used to access the segment for this step.
>   "$comment": >-
>     A source segment can connect to another segment at multiple connectors.
>     Therefore, for each step in a transition across segments, an explicit
>     connector is necessary to uniquely identify the point of transition.
> ```

I agree that relationTarget is not a good name, but I would keep it as is for now. The reason, I initially created another type for this in the first place is that it can be used also for signposts which is also a relation, otherwise a type would not be necessary. Lets keep this for now and revisit with the signpost schema

sanjibdutta1:
> > Can you do a quick pass over the Docusaurus section on transition restrictions?
> > I think you can make just make some very limited point edits, like a word here and a word there, to bring it in line with the schema changes.
> > Screenshot of Docusaurus section on turn restrictions
> 
> I can make changes but I have no way to test it. But since the changes are benign in nature, I guess its ok

Actually on a second thought, I would like to discuss further about turn vs. transition, I do not have any particular affinity, but then to be consistent we need to replace turn with transition everywhere. I was about to do that when I realized changing u-turn to u-transition would be funny and so is left turn vs. left transition. Why do you think transition is a better option over turn? Even in our spoken language, we never call it left transition or u transition, even OSM nomenclature is turn not transition

I think, we should rather change the property name from prohibitedTranstion to prohibitedTurnSequence - agree?

> 
image

examples/transportation/docusaurus/turn-restriction-01-source.yaml
sanjibdutta1: Fixed

schema/transportation/segment.yaml
sanjibdutta1: Fixed

schema/transportation/segment.yaml
sanjibdutta1: Fixed

schema/transportation/segment.yaml
sanjibdutta1: Fixed

PR #57, opened 2023-10-05
Add signpost schema

sanjibdutta1:
This is the first draft for signpost schema.

### DIFF
--- a/examples/transportation/docusaurus/signposts.yaml
+++ b/examples/transportation/docusaurus/signposts.yaml
@@ -1,0 +1,28 @@
+---
+id: overture:transportation:example:simple-signpost-source
+type: Feature
+geometry:
+  type: LineString
+  coordinates:
+    - [-113.57822030759499, 50.01868388494378]
+    - [-113.57831482025354, 50.018860947117304]
+    - [-113.57851814418316, 50.01923724443006]
+properties:
+  theme: transportation
+  type: segment
+  version: 4
+  updateTime: "2023-08-21T16:23:13-06:00"
+  subType: road
+  connectors:
+    - overture:transportation:example:simple-signpost-connector1
+    - overture:transportation:example:simple-signpost-connector2
+  road:
+    signposts:
+      - direction: forward
+        destination:
+          segmentId: overture:transportation:example:simple-signposts-dest
+          connectorId: overture:transportation:example:simple-turn-restriction-connector2
+          signDirection: North
+          signTextType: Branch
+          signText: "M 21"
+          direction: forward
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -221,6 +221,31 @@ properties:
                     items: { "$ref": "#/$defs/propertyDefinitions/lane" }
                     minItems: 1
               minItems: 1
+        signposts:
+          description: Describes signpost along the a segment
+          type: array
+          items:
+             allOf:
+               - "$ref": "#/$defs/propertyContainers/directionContainer"
+             unevaluatedProperties: false
+             type: object
+             properties:
+               destination:
+                 type: object
+                 required: [segmentId, signTextType, signText]
+                 allOf:
+                   - "$ref": "#/$defs/propertyContainers/directionContainer"
+                 properties:
+                   segmentId: 
+                     type: string
+                   connectorId: 
+                     type: string # Connector id is optional if the segments are not connected or if the the connector can be unambiguously identified 
+                   signDirection:
+                     "$ref": "#/$defs/propertyDefinitions/compassDirection"
+                   signTextType:
+                     "$ref": "#/$defs/propertyDefinitions/signTextType"
+                   signText: 
+                     type: string            
         restrictions:
           description: Restrictions on how the road may be used
           "$comment": >-
@@ -392,6 +417,27 @@ properties:
           description: direction of travel on the target that pertains to the relation
           "$ref": "#/$defs/propertyDefinitions/direction"
       required: [segmentId]
+    compassDirection:
+      description: >-
+        Four cardinal and four ordinal direction
+      type: string
+      enum:
+        - East
+        - West 
+        - North
+        - South 
+        - NorthEast
+        - NorthWest 
+        - SouthEast
+        - SouthWest 
+    signTextType:
+      description: >-
+        Text type for a sign post. Gives general direction as to how to use the sign text 
+      type: string
+      enum:
+        - Text
+        - Branch 
+        - Route
   propertyContainers:
     atRangeContainer:
       description: >-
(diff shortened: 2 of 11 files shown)


examples/transportation/docusaurus/signposts.yaml
```diff
@@ -0,0 +1,28 @@
...
+  subType: road
+  connectors:
+    - overture:transportation:example:simple-signpost-connector1
+    - overture:transportation:example:simple-signpost-connector2
+  road:
+    signposts:
+      - direction: forward
+        destination:
+          segmentId: overture:transportation:example:simple-signposts-dest
+          connectorId: overture:transportation:example:simple-turn-restriction-connector2
```
vbogdana, line 24: probably needs to be renamed

schema/transportation/segment.yaml
```diff
@@ -221,6 +221,31 @@ properties:
...
+               destination:
+                 type: object
+                 required: [segmentId, signTextType, signText]
+                 allOf:
+                   - "$ref": "#/$defs/propertyContainers/directionContainer"
+                 properties:
+                   segmentId: 
+                     type: string
+                   connectorId: 
+                     type: string # Connector id is optional if the segments are not connected or if the the connector can be unambiguously identified
```
brad-richardson, line 242:
```suggestion
                     type: string # Connector id is optional if the segments are not connected or if the connector can be unambiguously identified 
```

schema/transportation/segment.yaml
```diff
@@ -221,6 +221,31 @@ properties:
                     items: { "$ref": "#/$defs/propertyDefinitions/lane" }
                     minItems: 1
               minItems: 1
+        signposts:
+          description: Describes signpost along the a segment
```
brad-richardson, line 225:
```suggestion
          description: Describes signpost along the segment
```

schema/transportation/segment.yaml
```diff
@@ -392,6 +417,27 @@ properties:
...
+        - West 
+        - North
+        - South 
+        - NorthEast
+        - NorthWest 
+        - SouthEast
+        - SouthWest 
+    signTextType:
+      description: >-
+        Text type for a sign post. Gives general direction as to how to use the sign text
```
brad-richardson, line 435: It's not clear to me what these enum values mean by your description here. Could you add some additional detail?

schema/transportation/segment.yaml
```diff
@@ -221,6 +221,31 @@ properties:
...
+               - "$ref": "#/$defs/propertyContainers/directionContainer"
+             unevaluatedProperties: false
+             type: object
+             properties:
+               destination:
+                 type: object
+                 required: [segmentId, signTextType, signText]
+                 allOf:
+                   - "$ref": "#/$defs/propertyContainers/directionContainer"
+                 properties:
```
brad-richardson, line 238: How do we know at what point along the segment the signs exist?

examples/transportation/docusaurus/signposts.yaml
```diff
@@ -0,0 +1,28 @@
...
+          signDirection: North
+          signTextType: Branch
+          signText: "M 21"
+          direction: forward
```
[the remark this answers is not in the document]
vcschapp: > If there are 2 twoway segments connected with a connector in the middle of those two segments (this would be a simple "cross" (+) intersection), if we want to model a signpost we need to have the direction information about both of these segments, so that we know at which part of the "from" segment the signpost is visible and also to which part of the "target" segment is the signpost pointing. Does the schema provide support for this?

Agree, as to the first one. (At which part of the "from" segment the sign~~post~~ is visible.) The second one probably isn't necessary because there's already a reference to a connector, which has a physical location (Point geometry).

The schema does support this. We would require geometric scoping of the sign location itself (on the segment where the sign is located).

schema/transportation/segment.yaml
vcschapp: @vbogdana had a similar question. Agree this is a gap.

schema/transportation/segment.yaml
```diff
@@ -392,6 +417,27 @@ properties:
...
+      type: string
+      enum:
+        - East
+        - West 
+        - North
+        - South 
+        - NorthEast
+        - NorthWest 
+        - SouthEast
+        - SouthWest
```
vcschapp, line 432: The schema naming standard is `camelCase`, so this is `northEast`, `northWest`, *etc.*

Incidentally this probably isn't the first time I've made a comment like this....

schema/transportation/segment.yaml
```diff
@@ -392,6 +417,27 @@ properties:
...
+        - SouthEast
+        - SouthWest 
+    signTextType:
+      description: >-
+        Text type for a sign post. Gives general direction as to how to use the sign text 
+      type: string
+      enum:
+        - Text
+        - Branch 
+        - Route
```
vcschapp, line 440: `camelCase` please.

schema/transportation/segment.yaml
```diff
@@ -221,6 +221,31 @@ properties:
...
+                 properties:
+                   segmentId: 
+                     type: string
+                   connectorId: 
+                     type: string # Connector id is optional if the segments are not connected or if the the connector can be unambiguously identified 
+                   signDirection:
+                     "$ref": "#/$defs/propertyDefinitions/compassDirection"
+                   signTextType:
+                     "$ref": "#/$defs/propertyDefinitions/signTextType"
+                   signText:
```
vcschapp, line 247: - The "stutter naming" (repetition of the prefix word "sign") doesn't seem necessary here. Can't `signDirection` be `direction`, *etc.*?
- I notice this doesn't say `signpostDirection`, *etc.* This seems to support my contention that the important thing is the sign, not the post holding it up :)
- Where are we going to get the data for this? How much of it exists in OSM, in which tags?

vcschapp (changes requested):
My tippy top-level questions are:

1. How are we going to get the data for this?
   - Can you share the OSM tag mappings?
   - We need a decent amount of data in OSM, IMO, before modeling this into the schema becomes a high priority.
3. I'd love to see a diagram explaining it.
   - The best way to resolve this would be to write the *Signs* section of the [Roads](https://docs.overturemaps.org/themes/transportation/roads) docusaurus page.
5. I think this is about Signs, not Posts.

vcschapp:
A question I have is what usage are the sign~~post~~s data being put to that is separate from the network itself?

For example @sanjibdutta1, you do a lot of work out routing engines. How does your routing engine use signs data, separately from its knowledge about the connectivity of the roads? Does the generated route reference signs, like "you will see a big green sign with exit 2A written on it coming up on the left"? Or does it only need the exit numbers?

This is an area I don't know a lot about, and I don't fully still understand the purpose.

vcschapp:
Closing this PR as nothing has happened in two months. We have access to it from the feature issue here: https://github.com/OvertureMaps/schema-wg/issues/188.

PR #60, opened 2023-10-10
Change prohibitedTransitions to prohibitedTurns in the schema. Also c…

sanjibdutta1:
…hange the documentation so that it refers to the resatrcition as turn restriction but still call transition to describe the act. This will also require change in the combobulator which will be done once this is approved


vcschapp:
I'm not super-aligned with the executive decision to switch from `prohibitedTransitions` to `prohibitedTurns`, as this is something that did argue over at length between yourself, myself, and @AleksandraDebicka-TomTom and now this seems like an eleventh-hour decision to undo the previous decision.

The arguments mentioned in the meeting this morning about seeing things called "turn restrictions" in other APIs *etc.* don't resonate hugely with me because we are not them, and their method of modeling is not our method of modeling. (Also, OpenStreetMap doesn't even call them turn restrictions. That's an informal phrase. They're just `restriction` relations and not every `type` references a "turn". Just like us.)

vcschapp:
As a Git nit, the following isn't an ideal commit stream:

```
1) Change prohibitedTransitions to prohibitedTurns in the schema. Also c…
2) Missed test fixes in the first
```

This creates a bloated commit history full of low-value commits. In 1 month or 3 years, nobody wants to have to navigate through someone's local development process (one commit with most of the change, a second commit to fix the problems from the first commit), The ideal is each commit represents one business unit of change. In this case, one could consider a "fixup" merge of commit no. 2 into commit no. 1 leaving one commit that creates a single atomic working change....

sanjibdutta1:
I am not sure I understand your second comment - can you explain? I forgot to fix the tests during my first commit, so what should be the process? Abandon my first commit and redo the whole thing in a single commit?

sanjibdutta1:
As for the turn vs. transition, it seems you are the only one who does not like turn :) I am not sure how the decision making should work here, but if you have the final say for the task force, then so be it, I will force myself to change it, otherwise, I will leave it as is.... As I was saying every other provider calls it turns (including OSM) how exactly are we different from every other map data providers on the planet?

sanjibdutta1:
Aha! I get your git-nit comment, yeah thats a better idea...but the deed is already done :)

PR #62, opened 2023-10-17
Revert addition of entityId

ibnt1:
Reverting 90f8e5fef7187fce2d5e919f99257517c4c096b2: we will not use `entityId` property for stable ids.


jenningsanderson (approved):
entityid -> id ✅

varapmsft (approved):

examples/admins/locality/admin-named-locality.yaml
```diff
@@ -1,11 +1,9 @@
-id: capitolHillFeatureId
```
jwass, line 1: Why did the id disappear here?

ibnt1: it's just a git revert of my old commit, where i had added it then to highlight that there were supposed to have both id and entityId.

AleksandraDebicka-TomTom (approved):

vcschapp (approved):

PR #64, opened 2023-10-18
Fix tr documentation

sanjibdutta1:


jenningsanderson:
@RobSoetewey-TomTom Seems like this should be merged and then dev merged immediately with `main` to update the schema to match what is being delivered tomorrow. @sanjibdutta1 is that correct?

sanjibdutta1:
> @RobSoetewey-TomTom Seems like this should be merged and then dev merged immediately with `main` to update the schema to match what is being delivered tomorrow. @sanjibdutta1 is that correct?

Yes this is correct. I am awaiting @vcschapp  to approve the change.

vcschapp (approved):

PR #68, opened 2023-10-25
Fix 113: [Defect] Inconsistent country and region code property names across addresses and admins themes

DavidKarlas:
Simply moved definition of properties to root defs.yaml and updated places.yaml to reference it + updated examples

### DIFF
--- a/counterexamples/places/bad-address-invalid-property.yaml
+++ b/counterexamples/places/bad-address-invalid-property.yaml
@@ -12,8 +12,8 @@ properties:
       locality: "New York"
     - freeform: "770 Brodway, #802"
       locality: "New York"
-      region2: "NY"
-      country: "US"
+      isoSubCountryCode2: "NY"
+      isoCountryCodeAlpha2: "US"
   # Overture properties
   theme: places
   type: place
--- a/counterexamples/places/bad-address-missing-required-property.yaml
+++ b/counterexamples/places/bad-address-missing-required-property.yaml
@@ -11,8 +11,8 @@ properties:
     - freeform: "770 Broadway, Floor 8"
       locality: "New York"
     - locality: "New York"
-      region: "NY"
-      country: "US"
+      isoSubCountryCode: "NY"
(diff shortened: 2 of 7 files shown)


danbeeee:
Would it make sense to instead change the names used in the admin layer to "region" and "country"?
If not, why not?

I went searching for example address schemas used by other groups, and I found this:
https://developers.google.com/search/docs/appearance/structured-data/local-business
which refers to the fields as "addressRegion" and "addressCountry".

danbeeee (changes requested):
I think the existing names "region" and "country" are good, and I suggest that other layers should change their schema to "region" and "country".

jwass:
My guess is that addresses can't be enforced to use ISO codes for country and region since they're always messy and have lots of exceptions. Maybe we leave this as-is for now and see if the address tf has a strong opinion as they form a better addressing schema?

RobSoetewey-TomTom:
> My guess is that addresses can't be enforced to use ISO codes for country and region since they're always messy and have lots of exceptions. Maybe we leave this as-is for now and see if the address tf has a strong opinion as they form a better addressing schema?

Right. There's some notion that iso 3166-1 alpha 2 'can' be used but then you see that followed with examples using "USA". I think we should try to standardize to a single country reference in Overture. Most logical would be to use iso-3166-1 alpha 2 throughout, and call it like that.
Fixing that in the schema will become harder in the future though.

danbeeee:
Places is already using iso 3166-1 codes for the country field, and iso 3166-2 codes for the region field, where available.  I agree that using standard codes is good.

However I think you can find a better way to communicate the format of the field, rather than encoding the format in the field name.  Perhaps you could add this as a comment, instead?

Part of the reason I am asking is that we have only just now (last week) gotten Microsoft to conform to the existing places schema.  If we ask them to change their format for the data they send us, it will create a lot of work for both me and them.  It sounds like the changes would not even be permanent since there is a separate "address group" that wants to rewrite this?

PR #69, opened 2023-10-25
[WIP] add sourceTags attribute?

jenningsanderson:
## Context

The `base` theme currently has a `sourceTags` attribute to capture tags that come directly from the source, primarily passing through OSM tags. 

This works for `base` because there is only one source (OSM). Is there value in extending this attribute to other themes?

If so, how is it captured? When data could be coming from multiple sources, how would sourceTags be organized?

```yaml
sourceTags: 
  osm: 
    building: garage
  esri:
    "addr:number": 129
```

Or the order of `sources` is `[osm,esri]` and `sourceTags` looks like: 

```yaml
sourceTags: 
  - building:gararge
  - "addr:number": 129
```

Scenarios:
1. **Buildings**: buildings from either OSM or Esri might have addresses or other tags associated that users may want. If, for example, our current list of classes are insufficient, a user might be interested in knowing what the original `building` tag was.

### DIFF
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -244,6 +244,10 @@ description: Common schema definitions shared by all themes
         country:
           description: ISO 3166 alpha2 country code
           type: string
+    sourceTags:
+      description: Attributes that pass directly through from the source data
+      type: object
+
     wikidata:
       description: A wikidata ID if available, as found on https://www.wikidata.org/.
       type: string


schema/defs.yaml
```diff
@@ -244,6 +244,10 @@ description: Common schema definitions shared by all themes
         country:
           description: ISO 3166 alpha2 country code
           type: string
+    sourceTags:
+      description: Attributes that pass directly through from the source data
+      type: object
```
vcschapp, line 249: If it's truly a tags array, it shouldn't be *just* an `object`: should have validation constraints to ensure the keys and values must be strings.

You need something along the lines of https://stackoverflow.com/a/36237572/1911388.

vcschapp (commented):
Two high-level comments.

1. There probably isn't a screaming rush to add this, versus other high-priority needs, but it is interesting.
2. If we do it, recall that the current structure of the `sources` tag is based on *dataset*. There's no concept of a "source" separate from a dataset. So the organization would have to be mapping from dataset to tags.

vcschapp:
We discussed this PR in last week's schema TF meeting on [2024-02-28](https://github.com/OvertureMaps/schema-wg/issues/283) and the conclusion was to close it as we don't necessarily want to bring source tags into all feature types, as that tends to undermine the idea of a schema.

We can resuscitate this if that decision changes later.

PR #72, opened 2023-11-02
[Admins] Add Area type and allow only Point on Locality 

DavidKarlas:
With new relations we can now add multiple different geometries to one locality, here we are changing modeling for localities to always have point and in case it also has polygon adds it via new Area type which references Locality via `localityId` property.

Also included few minor changes into this PR:
  * Removed mentioning disputed boundaries as missing feature. This was added few weeks ago, didn't realize it was mentioned here...
  * Changed `context` property/relation name to `contextId` to reflect new guidance


jenningsanderson:
View the changes to documentation and schema here: https://dfhx9f55j8eg5.cloudfront.net/pr/72

jenningsanderson (approved):
Looks good to me — would calling it `administrativeArea` instead of just `area` be more descriptive? 

`area` is a bit of an overloaded term in geospatial, so I think it could be helpful to more clearly delineate that.

schema/admins/area.yaml
```diff
@@ -0,0 +1,26 @@
+---
+"$schema": https://json-schema.org/draft/2020-12/schema
+title: Area Schema
```
ibnt1, line 3: I think it would help clarify that this is _Locality_ Area Schema at least here in the title. 

Not sure if it's worth considering or not because it's a minor thing, so i'll leave that up to you but if we care to be consistent with when we qualify the _kind_ of feature, then we should call this new type `localityArea`, similar to how we have `administrativeBoundary`, or conversely rename that to `boundary`.

schema/admins/area.yaml
```diff
@@ -0,0 +1,26 @@
...
+  properties:
+    unevaluatedProperties: false
+    required: [localityId]
+    allOf:
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+    properties:
+      localityId:
+        description: References specific feature of locality type
+        type: string
+      maritime:
```
ibnt1, line 24: `isMaritime` to reflect that it is a boolean property?

ibnt1 (approved):
looks good, left couple minor comments, thanks!

schema/admins/area.yaml
jenningsanderson: +1 to localityArea or administrativeArea

schema/admins/area.yaml
jenningsanderson: +1
I'd also vote that this property be refactored out to an `admins/defs.yaml` file and reused in administrativeBoundary to be consistent.

DavidKarlas: I wanted to do that but administrativeBoundary already had `maritime`, hence used same, but I agree, will replace in both types to `isMaritime`

DavidKarlas: Done.

schema/admins/area.yaml
DavidKarlas: Done.

jwass:
> Looks good to me — would calling it `administrativeArea` instead of just `area` be more descriptive?
> 
> `area` is a bit of an overloaded term in geospatial, so I think it could be helpful to more clearly delineate that.

I kind of agree but also it's within the admins theme so I feel like adding `administrative` is redundant. As I was reviewing I was thinking we might just want `locality`, `boundary`, and `area` or something like that. I agree that area is a bit of an odd duck here so maybe there's a better word (polygon?).

jenningsanderson:
> > Looks good to me — would calling it `administrativeArea` instead of just `area` be more descriptive?
> > `area` is a bit of an overloaded term in geospatial, so I think it could be helpful to more clearly delineate that.
> 
> I kind of agree but also it's within the admins theme so I feel like adding `administrative` is redundant. As I was reviewing I was thinking we might just want `locality`, `boundary`, and `area` or something like that. I agree that area is a bit of an odd duck here so maybe there's a better word (polygon?).

Yeah — the only problem is that `type` is currently unique globally, not per-theme, so `area` or `polygon` would be forever associated _only_ with this particular theme.

jwass:
> Yeah — the only problem is that `type` is currently unique globally, not per-theme, so `area` or `polygon` would be forever associated _only_ with this particular theme.
> 
> At least, that's how it's currently in the schema, but maybe it doesn't matter if `type` is reused b/n themes?

Interesting. I actually didn't think `type` was unique globally?? I think the schema is usually seeing if both a theme and type match, then dispatching to the right type. Still have to check but if it is the case, would you think `area` is okay? Or still prefer `administrativeArea`?

DavidKarlas:
I just want to add I"m super happy with current name of `localityArea` because it makes it super clear it adds areas to type before it which is `locality`...

jwass:
> I just want to add I"m super happy with current name of `localityArea` because it makes it super clear it adds areas to type before it which is `locality`...

Sounds good. I guess for consistency, should `administrativeBoundary` then be `localityBoundary` or something like that? Seems weird to have administrative only there but nowhere else.

RobSoetewey-TomTom (approved):

DavidKarlas:
FYI, I removed `isMaritime` property from `localityArea` because it needs more discussion... Since Landmass area type can be derived from data + I would like to use subType instead of property...

jwass (approved):

mojodna (approved):

schema/admins/locality.yaml
```diff
@@ -35,6 +33,6 @@ properties:
         enum: [administrativeLocality, namedLocality]
       localityType: { "$ref": "./defs.yaml#/$defs/propertyDefinitions/localityType" }
       names: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/names" }
-      context:
+      contextId:
```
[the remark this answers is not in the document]
DavidKarlas: Since we want to merge this before next week release, I would prefer if we postpone this discussion to later time and different PR, sounds OK?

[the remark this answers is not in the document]
DavidKarlas: I want to add I totally agree with renaming this property or even removing it, just needs longer discussion...

schema/admins/localityArea.yaml
```diff
@@ -0,0 +1,24 @@
...
+  properties:
+    unevaluatedProperties: false
+    required: [localityId]
+    allOf:
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+    properties:
+      localityId:
+        description: References specific feature of locality type
+        type: string
+      isMaritime: { "$ref": "./defs.yaml#/$defs/propertyDefinitions/isMaritime" }
```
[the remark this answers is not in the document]
DavidKarlas: Since we want to merge this before next week release, I would prefer if we postpone this discussion to later time, sounds OK?

DavidKarlas:
Got all needed approvals and resolved all open comments

PR #73, opened 2023-11-03
Fix type maxiumum->maximum on linearlyReferencedPosition

DavidKarlas:
Looks like test on https://github.com/OvertureMaps/schema/blob/main/examples/transportation/segment/road/lanes/lanes-on-straight-road-lr.yaml is failing now, because it has values 0-100 and not 0-1, can someone from transportation figure this one out?

### DIFF
--- a/examples/transportation/segment/road/road-with-lr-name.yaml
+++ b/examples/transportation/segment/road/road-with-lr-name.yaml
@@ -20,7 +20,7 @@ properties:
     roadNames:
       - at:
           - 0
-          - 50
+          - 0.5
         common:
           - value: Common Road Name 1
             language: local
@@ -28,8 +28,8 @@ properties:
           - value: SRN1
             language: local
       - at:
-        - 50
-        - 100
+        - 0.5
+        - 1
         common:
           - value: Common Road Name 2
             language: local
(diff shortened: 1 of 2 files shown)


brad-richardson (approved):

sanjibdutta1 (approved):
Did you run the tests? Since tests references these schema elements, some may fail. Please run the tests once

vcschapp:
> Looks like test on https://github.com/OvertureMaps/schema/blob/main/examples/transportation/segment/road/lanes/lanes-on-straight-road-lr.yaml is failing now, because it has values 0-100 and not 0-1, can someone from transportation figure this one out?

It's just an example file - can't you just edit the file? I'd suggest scaling down the maximum values so they fit in 0-1. If they're all in 0-100, maybe divide by 100?

I very much appreciate this schema fix, but I think we're going to have to demonstrate more end-to-end ownership if we want to scale as a group. If you make a one-character deletion to fix a bug, isn't it also fair to run the tests and bring any incorrect test files into sync?

DavidKarlas:
It was probably stupid from me to not just fix examples, but at time I though I'm misunderstanding something and was not sure anymore. I update examples so tests pass now.
Sorry,
David

vcschapp (approved):

AleksandraDebicka-TomTom (approved):

PR #96, opened 2024-01-09
Add wikidata property to Admins theme

DavidKarlas:
Adding wikidata property to Admins theme

### DIFF
--- a/examples/admins/locality/admin-named-locality.yaml
+++ b/examples/admins/locality/admin-named-locality.yaml
@@ -15,6 +15,7 @@ properties:
   subType: namedLocality
   theme: admins
   type: locality
+  wikidata: Q5083
   updateTime: '2023-02-22T23:55:01-08:00'
   version: 0
 type: Feature
--- a/schema/admins/locality.yaml
+++ b/schema/admins/locality.yaml
@@ -33,6 +33,7 @@ properties:
         enum: [administrativeLocality, namedLocality]
       localityType: { "$ref": "./defs.yaml#/$defs/propertyDefinitions/localityType" }
       names: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/names" }
+      wikidata: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/wikidata" }
       contextId:
         description: Context entity is the most granular entity that logically contains given entity (but doesn't have to contain it spatially due to minor discrepancies in geometries)
         type: string


sasastanojkov (approved):

jwass:
I'm okay with this but note that Places has a wikidata element inside the brand (https://github.com/OvertureMaps/schema/blob/main/schema/places/place.yaml#L76)

I wonder if there's a way to unify these? <-- @DavidKarlas  @vcschapp

jwass (approved):

vcschapp:
> I'm okay with this but note that Places has a wikidata element inside the brand (https://github.com/OvertureMaps/schema/blob/main/schema/places/place.yaml#L76)
> 
> I wonder if there's a way to unify these? <-- @DavidKarlas @vcschapp

This one seems to be done the same way as the base theme, looking at the base theme PR wikidata elements: https://github.com/OvertureMaps/schema/pull/42.

I think [having the *brand* wikidata separate in places](https://github.com/OvertureMaps/schema/blob/main/schema/places/place.yaml#L71-L76) does make sense because the brand is conceptually different than the place itself. I believe the place could also optionally have a wikidata, separate from the brand.

The question I'm having is, isn't wikidata basically a core property we might want on buildings and some road segments too? Maybe rather than piecemeal inserting it per-theme we should insert it into the core feature properties?

Thoughts?

MarkLebegge-TomTom (approved):

DavidKarlas:
I agree, I was surprised when I learned this is not on everything, well at least it is defined in global defs.yaml, do you want me to add it to every theme? I'm fine with that. I think Eiffel Tower, being in Places theme, should be able to have wikidata set without Brand. And probably some major avenues that are roads have wikidata, there are 334k highways with wikidata set in OSM: https://taginfo.openstreetmap.org/keys/highway#combinations

DavidKarlas:
After some more though... I would prefer if we could merge this as is, and not go into other Themes at this times, because for some themes it might not make sense such as Addresses so we can't just blindly apply to all, and it is really simple for each theme to add this property when that theme see data available.
@vcschapp do you agree, can we merge this before Wednesday?

jwass:
I agree that this can be merged as-is. And we can include a top-level wikidata field on all relevant themes without any change in the future. Maybe one small change would be to move the wikidata field up to top-level defs propertyDefinitions and then include and reference it in whichever themes use it. But I don't think it needs to be done now.

I still feel like this is a GERS issue though. At least in my mind GERS should be maintaining a database where records from different datasets refer to the same real-world entity. So this isn't unique to wikidata, although it might be special enough to warrant a field in the main distribution like we have now.

jenningsanderson (approved):

jenningsanderson:
> I still feel like this is a GERS issue though. At least in my mind GERS should be maintaining a database where records from different datasets refer to the same real-world entity. So this isn't unique to wikidata, although it might be special enough to warrant a field in the main distribution like we have now.

This is a great question, I will add to the GERS agenda. I see the big question here as being the separate `GERS database` to track these — or is the Overture data distribution itself the DB. I.e, theme producers ensure that wikidata IDs are accurate per feature and then the mapping of Wikidata IDs <-> GERS IDs exists at the feature level w/in the dataset? In this way, the GERS <-> Wikidata ID DB is a derivative of the Overture release, not the other way around?

vcschapp:
> I agree that this can be merged as-is. And we can include a top-level wikidata field on all relevant themes without any change in the future. Maybe one small change would be to move the wikidata field up to top-level defs propertyDefinitions and then include and reference it in whichever themes use it. But I don't think it needs to be done now.
> 
> I still feel like this is a GERS issue though. At least in my mind GERS should be maintaining a database where records from different datasets refer to the same real-world entity. So this isn't unique to wikidata, although it might be special enough to warrant a field in the main distribution like we have now.

100% agree here. It's an aliasing or concordance issue, and it's in the core of GERS. That doesn't mean that the common feature schema can't contain the concordances, but we should make sure they play together well.

vcschapp (approved):
Agreeing to merge for now, but I don't think this is the last word.

The things that need to be decided are:

1. Should these concordances or aliases be managed by GERS?
2. Should they be included in the schema?
3. Should they be a central common feature property?
4. Should they be nested under an `aliases` or `concordances` property?

PR #103, opened 2024-01-24
Adding elevation to base

jenningsanderson:
Adds elevation as a top-level attribute in `land`, `landUse`, and `water` types.

### DIFF
--- a/schema/base/defs.yaml
+++ b/schema/base/defs.yaml
@@ -7,6 +7,10 @@ description: Common schema definitions the base theme (primarily from OSM)
     sourceTags:
       description: Any attributes/tags from the original source data that should be passed through.
       type: object
+    elevation:
+      description: Elevation above sea level (in meters) of the feature.
+      type: integer
+      maximum: 9000
   propertyContainers:
     osmPropertiesContainer:
       title: "OSM Properties"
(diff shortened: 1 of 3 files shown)


BoZivk (approved):

ImreSamu:
Non-blocking comment

> "description: Elevation **above sea level (in meters)** of the feature."

Is a negative value expected here? 
Or will that be in a different field? --> "below sea level"

https://en.wikipedia.org/wiki/List_of_places_on_land_with_elevations_below_sea_level

jwass (commented):
Does elevation only apply to some types? I think it was mountain summits or things like that where it makes sense but should this be able to go on anything?

vcschapp:
> Does elevation only apply to some types? I think it was mountain summits or things like that where it makes sense but should this be able to go on anything?

Good question, I have the same one.

DavidKarlas (approved):

vcschapp:
In Schema TF [2024-01-31](https://github.com/OvertureMaps/schema-wg/issues/261), we agreed that this PR would be merged but that a bug would be entered to ensure that the `elevation` property is restricted to only those features types/sub-types for which it makes sense.

vcschapp:
Created a new issue and assigned it to @jenningsanderson : https://github.com/OvertureMaps/schema-wg/issues/263

vcschapp (approved):
Approved with the caveats discussed in Schema TF [2024-01-31](https://github.com/OvertureMaps/schema-wg/issues/261).

PR #104, opened 2024-01-25
Removing level property from connectors schema

TheRBajaj:
Removing levels property from connectors schema as it does not align well with segment feature. 
Issue: https://github.com/OvertureMaps/schema-wg/issues/138

### Testing
Tested by modifying examples and counterexamples accordingly and running test script.

### DIFF
--- a/counterexamples/transportation/connector/level-property-not-allowed.json
+++ b/counterexamples/transportation/connector/level-property-not-allowed.json
@@ -1,0 +1,18 @@
+{
+    "id": "unrecognizedLevelProperty",
+    "type": "Feature",
+    "geometry": {
+      "type": "Point",
+      "coordinates": [0, 0]
+    },
+    "properties": {
+      "theme": "transportation",
+      "type": "connector",
+      "version": 0,
+      "updateTime": "2023-04-24T09:19:00-07:00",
+      "level": 3,
+      "extExpectedErrors": [
+        "[I#/properties/level] [S#/properties/properties/unevaluatedProperties] not allowed"
+      ]
+    }
+  }
--- a/examples/transportation/connector/connector.yaml
+++ b/examples/transportation/connector/connector.yaml
@@ -12,5 +12,4 @@ properties:
   type: connector
   version: 0
   updateTime: "2023-02-23T00:02:01-08:00"
-  level: 3
 #
--- a/examples/transportation/docusaurus/connector.yaml
+++ b/examples/transportation/docusaurus/connector.yaml
@@ -9,4 +9,3 @@ properties:
   type: connector
   version: 0
   updateTime: "2023-02-23T00:02:01-08:00"
-  level: 3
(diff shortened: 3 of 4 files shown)


counterexamples/transportation/connector/bad-geometry-type.json
```diff
@@ -9,6 +9,7 @@
     "theme": "transportation",
     "type": "connector",
     "version": 2,
-    "updateTime": "2023-02-23T00:04:00-08:00"
+    "updateTime": "2023-02-23T00:04:00-08:00",
+    "level": 3
```
vcschapp, line 13: Can you please create a new counter-example for this instead of appending to this one?

---

This file is called `bad-geometry-type.json` and is meant to "test" the geometry field.

It would be preferable to create some kind of `level-property-not-allowed.json` for the level property.

---

Would you also mind using the `extExpectedErrors` feature on your new counterexample to specify the expected error output?

vcschapp (changes requested):
It all looks good, just needs a more precise counterexample. See comment inline.

jwass (approved):
I'm approving subject to addressing @vcschapp's comment about providing a new counterexample

counterexamples/transportation/connector/bad-geometry-type.json
TheRBajaj: Yup will separate the tests. How is extExpectedErrors triggered as it's not super clear since most tests don't use this feature?

vcschapp: I recommend you `grep`† for other test files that use `extExpectedErrors`, and also look at how it is used in `test.sh`. Maybe experiment by adding it to your counterexample file and seeing what happens?

@marinrelatic did some dive-deep diagnosis on this feature and fixed some problems with it in https://github.com/OvertureMaps/schema/pull/107, so he can probably help explain how it works if you get stuck. If you're still stuck after, come to me!

---

† *By `grep` I mean use your favorite find-in-files search utility!

counterexamples/transportation/connector/level-property-not-allowed.json
```diff
@@ -0,0 +1,18 @@
...
+      "theme": "transportation",
+      "type": "connector",
+      "version": 0,
+      "updateTime": "2023-04-24T09:19:00-07:00",
+      "level": 3,
+      "extExpectedErrors": [
+        "level property not allowed"
+      ]
+    }
+  }
```
vcschapp, line 18: Missing newline at EOF...

brad-richardson (approved):

counterexamples/transportation/connector/bad-geometry-type.json
TheRBajaj: resolved. made a separate test with correct `extExpectedErrors` string

vcschapp (approved):
```
                          ,
                           |\
                           |/0
                           '

      .-------.
      |_______|____________
     /====0=[]
  __/______/__\____
 jgs               `\~^~^~^~^~^~^~^~^~^


```

vcschapp:
@TheRBajaj would you kindly squash your commit history down to a single commit with a title along the lines of "Remove level property from transportation connectors"?

vcschapp:
Merging because it's a bug fix. Technically we're missing a rep from TomTom, but quorum is probably overkill for this change.

PR #112, opened 2024-01-31
🚨DO NOT MERGE UNTIL 2/15 🚨 Country and sub-division ISO code validation

TheRBajaj:
This PR adds validation to the definitions of `isoCountryCodeAlpha2` and `isoSubCountryCode` properties in the admin schema.
Issue: https://github.com/OvertureMaps/schema-wg/issues/117

### Testing
Tested by modifying examples and creating new counterexamples accordingly and running test script.

### Details
The following validation was added to `isoCountryCodeAlpha2`
    - `"minLength": 2`
    - `"maxLength": 2`
    - `"pattern": "^[A-Z]{2}$"`

The following validation was added to `isoSubCountryCode`
    - `"minLength": 4`
    - `"maxLength": 6`
    - `"pattern": "^[A-Z]{2}-[A-Z0-9]{1,3}$"`
    - This validation comes from the ISO reference guide: https://www.iso.org/obp/ui/en/#iso:std:iso:3166:-2:ed-4:v1:en
image

### DIFF
--- a/schema/admins/defs.yaml
+++ b/schema/admins/defs.yaml
@@ -73,9 +73,15 @@ description: Common schema definitions for admins theme
     isoCountryCodeAlpha2:
       description: ISO 3166-1 alpha-2 country code.
       type: string
+      minLength: 2
+      maxLength: 2  
+      pattern: ^[A-Z]{2}$
     isoSubCountryCode:
       description: ISO-3166-2 Country subdivision code.
       type: string
+      minLength: 4
+      maxLength: 6
+      pattern: ^[A-Z]{2}-[A-Z0-9]{1,3}$ 
     defaultLanguage:
       description: Most common language used within the area.
       "$ref": "../defs.yaml#/$defs/propertyDefinitions/language"
--- a/counterexamples/admins/locality/invalid-country-code-exceeded-max-length.json
+++ b/counterexamples/admins/locality/invalid-country-code-exceeded-max-length.json
@@ -1,0 +1,46 @@
+{
+    "type": "Feature",
+    "properties": {
+      "theme": "admins",
+      "type": "locality",
+      "version": 0,
+      "updateTime": "2023-02-22T23:55:01-08:00",
+      "subType": "administrativeLocality",
+      "localityType": "country",
+      "adminLevel": 1,
+      "isoCountryCodeAlpha2": "USA",
+      "defaultLanguage": "english",
+      "drivingSide": "right",
+      "contextId": "northAmericaId",
+      "extExpectedErrors": [
+        "[I#/properties/isoCountryCodeAlpha2] [S#/$defs/propertyDefinitions/isoCountryCodeAlpha2/pattern] does not match pattern '^[A-Z]{2}$'"
+      ]
+    },
+    "geometry": {
+      "type": "Polygon",
+      "coordinates": [
+        [
+          [
+            -61.8743292,
+            48.8068635
+          ],
+          [
+            -78.4016703,
+            26.273714
+          ],
+          [
+            -120.7749598,
+            32.2499745
+          ],
+          [
+            -126.928757,
+            49.0378679
+          ],
+          [
+            -61.8743292,
+            48.8068635
+          ]
+        ]
+      ]
+    }
+  }
--- a/counterexamples/admins/locality/invalid-country-code-less-than-min-length.json
+++ b/counterexamples/admins/locality/invalid-country-code-less-than-min-length.json
@@ -1,0 +1,46 @@
+{
+    "type": "Feature",
+    "properties": {
+      "theme": "admins",
+      "type": "locality",
+      "version": 0,
+      "updateTime": "2023-02-22T23:55:01-08:00",
+      "subType": "administrativeLocality",
+      "localityType": "country",
+      "adminLevel": 1,
+      "isoCountryCodeAlpha2": "U",
+      "defaultLanguage": "english",
+      "drivingSide": "right",
+      "contextId": "northAmericaId",
+      "extExpectedErrors": [
+        "[I#/properties/isoCountryCodeAlpha2] [S#/$defs/propertyDefinitions/isoCountryCodeAlpha2/pattern] does not match pattern '^[A-Z]{2}$'"
+      ]
+    },
+    "geometry": {
+      "type": "Polygon",
+      "coordinates": [
+        [
+          [
+            -61.8743292,
+            48.8068635
+          ],
+          [
+            -78.4016703,
+            26.273714
+          ],
+          [
+            -120.7749598,
+            32.2499745
+          ],
+          [
+            -126.928757,
+            49.0378679
+          ],
+          [
+            -61.8743292,
+            48.8068635
+          ]
+        ]
+      ]
+    }
+  }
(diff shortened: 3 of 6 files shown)


counterexamples/admins/locality/invalid-max-length-limit-is-alpha2.json
```diff
@@ -0,0 +1,88 @@
...
+            "value": "U.S."
+          }
+        ],
+        "alternate": [
+          {
+            "language": "local",
+            "value": "States"
+          }
+        ]
+      },
```
vcschapp, line 56: 1. [ ] This is the **old** names schema. If it validates, that would be surprising. Something is wrong, can you find out what?
2. [ ] The `names` is ultra-long but not the main point of this counterexample. Can you shorten it to just the primary name or even drop the `names` property altogether to make a simpler counterexample?

counterexamples/admins/locality/invalid-max-limit-iso-sub-country-code.json
```diff
@@ -0,0 +1,88 @@
...
+      "updateTime": "2023-02-22T23:55:01-08:00",
+      "subType": "administrativeLocality",
+      "localityType": "country",
+      "adminLevel": 1,
+      "isoSubCountryCode": "US-NYCC",
+      "defaultLanguage": "english",
+      "drivingSide": "right",
+      "contextId": "northAmericaId",
+      "names": {
+        "common": [
```
vcschapp, line 16: Same comment on `names` - rinse, lather, and repeat throughout.

vcschapp (changes requested):
Excellent stuff. There's an issue with the `names` in the counterexamples, but other than that looks good.

counterexamples/admins/locality/invalid-max-length-limit-is-alpha2.json
TheRBajaj: Removed the names property altogether since it doesn't make sense with these counterexamples.
upon looking at the `names` property schema definition, the counterexamples do use the old names schema but isn't that fine for counterexamples since it's expected to fail?

counterexamples/admins/locality/invalid-max-limit-iso-sub-country-code.json
TheRBajaj: resolved

counterexamples/admins/locality/invalid-max-length-limit-is-alpha2.json
vcschapp: Kind of yes, but mostly no.

The reason no, is that most of those counterexamples aren't meant to test the that aspect, they're supposed to test something else. So basically there's a false failure signal that could hide the true failure signal, if that makes sense.

counterexamples/admins/locality/invalid-min-limit-iso-sub-country-code.json
```diff
@@ -0,0 +1,46 @@
...
+      "subType": "administrativeLocality",
+      "localityType": "country",
+      "adminLevel": 1,
+      "isoSubCountryCode": "US",
+      "defaultLanguage": "english",
+      "drivingSide": "right",
+      "contextId": "northAmericaId",
+      "extExpectedErrors": [
+        "[I#/properties/defaultLanguage] [S#/$defs/propertyDefinitions/language/oneOf/1/pattern] does not match pattern '^[a-zA-Z]{2,3}(-[a-zA-Z]{4})?(-[a-zA-Z]{2})?$'"
+      ]
```
vcschapp, line 17: This seems like the wrong error - isn't it looking at a failure in the `defaultLanguage` property?

1. [ ] Should drop `defaultLanguage` because it contains an error that is irrelevant to the example.
2. [ ] Should fix expected error...

counterexamples/admins/locality/invalid-max-limit-iso-sub-country-code.json
```diff
@@ -0,0 +1,46 @@
...
+      "subType": "administrativeLocality",
+      "localityType": "country",
+      "adminLevel": 1,
+      "isoSubCountryCode": "US-NYCC",
+      "defaultLanguage": "english",
+      "drivingSide": "right",
+      "contextId": "northAmericaId",
+      "extExpectedErrors": [
+        "[I#/properties/defaultLanguage] [S#/$defs/propertyDefinitions/language/oneOf/1/pattern] does not match pattern '^[a-zA-Z]{2,3}(-[a-zA-Z]{4})?(-[a-zA-Z]{2})?$'"
+      ]
```
vcschapp, line 17: This seems like the wrong error - isn't it looking at a failure in the `defaultLanguage` property?

1. [ ] Should drop `defaultLanguage` because it contains an error that is irrelevant to the example.
2. [ ] Should fix expected error...

vcschapp (changes requested):
I found a second issue in a couple of counter-examples where they include a copy/pasted invalid `defaultLanguage` and the `extExpectedErrors` is testing the language issue, not the desired issue...

counterexamples/admins/locality/invalid-min-limit-iso-sub-country-code.json
TheRBajaj: I didn't notice that, changed the error message to something more appropriate

vcschapp (dismissed):
```
        _,--',   _._.--._____
 .--.--';_'-.', ";_      _.,-'
.'--'.  _.'    {`'-;_ .-.>.'
      '-:_      )  / `' '=.
        ) >     {_/,     /~)
snd     |/               `^ .'
```

vcschapp:
🚨 This is behind the schema freeze, so we'll have to wait to merge it to dev. 🚨

schema/admins/defs.yaml
```diff
@@ -73,9 +73,15 @@ description: Common schema definitions for admins theme
...
       description: ISO 3166-1 alpha-2 country code.
       type: string
+      minLength: 2
+      maxLength: 2  
+      pattern: ^[A-Z]{2}$
     isoSubCountryCode:
       description: ISO-3166-2 Country subdivision code.
       type: string
+      minLength: 4
+      maxLength: 6
```
jwass, line 83: Do you need the min/max length in both if the regex requires them?

jwass (approved):
Approved for me. But would prefer removing min/max length if they're not needed since the pattern already dictates that.

jenningsanderson (dismissed):
LGTM — Agree with @jwass, min/max length seems unnecessary if it' always 2 upper-case letters as demanded by the regex.

schema/admins/defs.yaml
TristanDiet-TomTom: As an answer to your question: no.
But I would keep the min length, cause it makes the schema easier to read (granted here it is not a complex regex).
Also if you would have a validator who doesn't have proper regex support, the length fields are a good fall back mechanism.

TristanDiet-TomTom (approved):
Approved, would leave the Length fields next to the pattern fields.

counterexamples/admins/locality/invalid-max-length-limit-is-alpha2.json
```diff
@@ -0,0 +1,46 @@
+{
```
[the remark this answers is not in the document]
TheRBajaj: will change the names before merging

schema/admins/defs.yaml
vcschapp: I tend to agree with @TristanDiet-TomTom.

vcschapp (approved):

vcschapp:
Doing a rebase and merge as this is a well-approved PR.

marinrelatic (commented):
Looks good

PR #114, opened 2024-02-02
🚨DO NOT MERGE UNTIL 2/15 🚨 Migrate namesv2 in counterexamples

jwass:
Note that this makes a schema change. It remove "local" as valid constant value for "language" tags. It was a mistake in namesv2 to have left this in.

We no longer need a "local" language as the name fields are `primary` (no language), `common` (map of valid BCP47 languages to values), and `rules` (where language is optional when it is unknown)


vcschapp (approved):
Aw, no `extExpectedErrors`?

jenningsanderson (approved):
LGTM. 

We do still need to determine the best approach to validating the languages — but that's bigger than this.

DavidKarlas (approved):

TristanDiet-TomTom (approved):

vcschapp:
Merged!

PR #115, opened 2024-02-02
updates to entity definition and data model page

danabauer:
starting a new pr for changes to feature model and gers sections
see it here: https://dfhx9f55j8eg5.cloudfront.net/pr/115

### DIFF
--- a/docusaurus/docs/gers/gers.mdx
+++ b/docusaurus/docs/gers/gers.mdx
@@ -10,11 +10,15 @@ GERS provides a mechanism to conflate datasets, matching one or more entities vi
 
 ## Components of a GERS ID
 
-Overture generates GERS IDs in a consistent format:
+Overture assigns a unique ID for an entity the first time that entity interacts with GERS. GERS generates IDs using this formula:
 
 `H3 tile address` +
 `data theme identifier` +
 `56-bit value applied uniformly across a given theme`
+
+The H3 index reflects the location of an entity when it was first assigned a GERS ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if it moves slightly. The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change.
+
+Put another way, a GERS ID should be treated as a unique identifier for an entity, and nothing more. We discourage users from attempting to extract information about an entity from its GERS ID.
 
 Here you can see a list of IDs from Overture's recent release of buildings data, pulled from the GeoParquet file with a simple [duckdb](https://duckdb.org/) query. You can find more information on accessing and querying Overture datasets in the How-to section of our documentation.
 
--- a/docusaurus/docs/gers/gers.mdx
+++ b/docusaurus/docs/gers/gers.mdx
@@ -8,19 +8,19 @@ Overture's Global Entity Reference System (GERS) is a framework for structuring,
 GERS provides a mechanism to conflate datasets, matching one or more entities via a unique GERS ID. For example, two geospatial datasets each containing a polygon that represents the Empire State Building can be easily conflated because both polygons will contain the same GERS ID, referring to the entity defined as: "a polygonal representation of the footprint of New York's Empire State Building."
 
 
-## Components of a GERS ID
+## GERS ID
 
-Overture assigns a unique ID for an entity the first time that entity interacts with GERS. GERS generates IDs using this formula:
+Overture generates a unique ID for an entity the first time it interacts with GERS, using the following formula:
 
 `H3 tile address` +
 `data theme identifier` +
 `56-bit value applied uniformly across a given theme`
 
-The H3 index reflects the location of an entity when it was first assigned a GERS ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if it moves slightly. The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change.
+The H3 index reflects the location of an entity when it is first assigned a GERS ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if it moves slightly. The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change.
 
-Put another way, a GERS ID should be treated as a unique identifier for an entity, and nothing more. We discourage users from attempting to extract information about an entity from its GERS ID.
+Put another way, a GERS ID should be treated as a unique identifier and nothing more. We discourage users from attempting to extract information about an entity from its GERS ID.
 
-Here you can see a list of IDs from Overture's recent release of buildings data, pulled from the GeoParquet file with a simple [duckdb](https://duckdb.org/) query. You can find more information on accessing and querying Overture datasets in the How-to section of our documentation.
+Here you can see a list of IDs from Overture's recent release of buildings data, selected from the GeoParquet file with a simple [duckdb](https://duckdb.org/) query. You can find more information on accessing and querying Overture datasets in the How-to section of our documentation.
 
 ```
 D SET s3_region='us-west-2';
--- a/docusaurus/docs/gers/gers.mdx
+++ b/docusaurus/docs/gers/gers.mdx
@@ -16,11 +16,11 @@ Overture generates a unique ID for an entity the first time that entity interact
 `data theme identifier` +
 `56-bit value dependent on theme`
 
-The H3 index reflects the location of an entity when Overture first assigns it an ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if its location changes slightly. The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change. Put another way, a GERS ID should be treated as a unique identifier and nothing more. Overture discourages users from attempting to extract information about an entity from its GERS ID.
+The H3 index reflects the location of an entity when Overture first assigns it an ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if its location moves within its assigned H3 layer. Put another way, a GERS ID should be treated as a unique identifier and nothing more. Overture discourages users from attempting to extract information about an entity from its GERS ID.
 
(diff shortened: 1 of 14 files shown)


docusaurus/docs/gers/gers.mdx
```diff
@@ -5,16 +5,12 @@ label: GERS
...
-
-For example, two geospatial datasets each containing a polygon that represents the Empire State Building can be easily conflated because both polygons will contain the same GERS ID, referring to the _entity_ defined as: "a polygonal representation of the footprint of New York's Empire State Building."
+GERS provides a mechanism to conflate datasets, matching one or more entities via a unique GERS ID. For example, two geospatial datasets each containing a polygon that represents the Empire State Building can be easily conflated because both polygons will contain the same GERS ID, referring to the entity defined as: "a polygonal representation of the footprint of New York's Empire State Building."
 
 
 ## Components of a GERS ID
 
-The Global Entity Reference System generates IDs in a consistent format:
+Overture generates GERS IDs in a consistent format:
```
jenningsanderson, line 13: We might want to add something to this effect: 

An entity is assigned a GERS ID the first time that Overture sees it. There is _some_ structure to the GERS ID when it is generated, using the following formula: <formula>

However, because GERS IDs are intended to be stable over time, a new ID will not be generated for an entity if it moves slightly, even though the H3 index may no longer be "correct."

Put another way, a GERS ID should be treated as a unique identifier for an entity, and nothing more. Users are discouraged from attempting to extract any information from a GERS ID, please consider it a UUID for an entity and nothing more.

... or something like that?

danabauer: Instead of "may no longer be correct" perhaps we could say: 
```
The H3 index reflects the location of the entity when it was first assigned a GERS ID. Because GERS IDs are intended to be stable over time,  a new ID will not be generated for an entity if it moves slightly. (what do you mean by slightly?)
```

Is this also true of the theme identifier? For example, a lake might initially be assigned a theme identifier related to the base theme; its ID will not change if it is later moved to a separate water theme.

jwass (approved):
Looks great!

vcschapp (approved):
Approving, as long as @jenningsanderson and @jwass have a chance to review more carefully.

DavidKarlas (dismissed):

TristanDiet-TomTom (approved):

docusaurus/docs/gers/gers.mdx
```diff
@@ -5,22 +5,20 @@ label: GERS
...
-The Global Entity Reference System generates IDs in a consistent format:
+Overture generates a unique ID for an entity the first time that entity interacts with GERS. GERS IDs have the following structure:
 
 `H3 tile address` +
 `data theme identifier` +
-`56-bit value applied uniformly across a given theme`
+`56-bit value dependent on theme`
 
-Here you can see a list of IDs from Overture's recent release of buildings data, pulled from the GeoParquet file with a simple [duckdb](https://duckdb.org/) query. You can find more information on accessing and querying Overture datasets in the How-to section of our documentation.
+The H3 index reflects the location of an entity when Overture first assigns it an ID. Because GERS IDs are intended to be stable over time, Overture will not generate a new ID for that entity if its location changes slightly. The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change. Put another way, a GERS ID should be treated as a unique identifier and nothing more. Overture discourages users from attempting to extract information about an entity from its GERS ID.
```
jenningsanderson, line 19: Let's scrap this part: ~The same is true for the data theme identifier; if Overture moves an entity from an existing data theme to a new data theme, the ID will not change~ — in that case the ID would likely change

docusaurus/docs/gers/gers.mdx
```diff
@@ -5,22 +5,20 @@ label: GERS
...
+
+Here you can see a list of unique IDs from Overture's recent release of buildings data, selected from the GeoParquet file with a simple [duckdb](https://duckdb.org/) query. You can find more information on accessing and querying Overture datasets in the How-to section of our documentation.
 
 ```
```
jenningsanderson, line 23: I think if you add "sql" here, it will add nice formatting and color: 
```
```sql
```

jenningsanderson: Maybe we also limit 5 here and reduce the size of this output

docusaurus/docs/overview/feature-model/index.mdx
```diff
@@ -47,20 +41,42 @@ Overture data is described by the [_simple feature model_](https://www.ogc.org/s
...
+## Extensible attributes
+
+Any additional top-level property prefixed with `ext` may be added to an Overture feature, usually via the conflation of an external dataset and an Overture dataset.
+
+## Measurements
 
-- [Global Entity Reference System (GERS) IDs](../../gers/gers.mdx)
-- [Names](names)
-- [Scoped and rule-based properties](scoping-rules)
+Measurements in the schema follow [The International System of Units (SI)](https://www.bipm.org/en/publications/si-brochure), e.g. meters for building heights. Regulations and restrictions encoded in the schema follow local units. Speed limits, for example, appear as miles per hour (mph) in features located in the United States. Features in Germany show kilometers per hour (kph) for speed limits.
```
vcschapp, line 77: This I would characterize a little bit differently.

I would say that:

1. Objective measurements of things in the world follow SI, meaning heights, widths, lengths, *etc.* that represent real-world properties *of features*.
    - In the schema, these values are provided as scalar numeric value without units, since the units are implied and documented in place in the schema.
    - The reason we do this is to maximize consistency and predictability.
3. Quantities specified in regulatory rules, norms, and customs follow local specifications wherever possible.
    - In the schema, these values are provided as two-element arrays where the first element is the scalar numeric value and the second value is the units, *e.g.* `[100, "km/h"]` <--- double check this is correct though.
    - The reason we use the local units is to ensure that norms are specified in the most relevant way to the people who are actually there.
    - The reason we include the units is to maximize consistency and predictability in the schema.

Those are just the rough cliff notes around how things are done and why.

docusaurus/docs/overview/feature-model/index.mdx
```diff
@@ -1,43 +1,37 @@
...
-      "primary": "Springfield Town Hall"
+      "primary": "Town Hall Seattle"
     },
     "sources": [
       {
-        "property": "",
-        "dataset": "microsoftMLBuildings,"
+        "property": "dataset",
+        "dataset": "OpenStreetMap,"
+        "recordid": "w140675988@10"
```
vcschapp, line 34: Could you double-check that this isn't specified as `recordId` with an uppercase "i" in the schema?

It should be, because the current schema uses `camelCase` for property names ... Although I hope/believe we will move the whole thing to `snake_case` by March...

vcschapp (approved):
Let's aim to merge this and do more updates in a new PR after.

TristanDiet-TomTom (approved):

docusaurus/docs/gers/gers.mdx
```diff
@@ -33,44 +31,39 @@ D SELECT id FROM read_parquet('s3://overturemaps-us-west-2/release/2024-01-17-al
...
-GERS IDs are stable (with a reasonable tolerance of error) across multiple versions of Overture data. We strive to ensure the mapping of a real-world thing to a GERS ID remains consistent. When stability is not possible, traceability will be provided. For example:
 
-- a single road segment is bisected by a new road and becomes two road segments: 1 GERS ID → 2 new GERS IDs
-- one large building footprint on the map is determined to be four smaller buildings when a higher resolution dataset becomes available: 1 GERS ID → 4 new GERS IDs
-- a building is shifted 10m west when higher resolution imagery is made available: GERS ID is preserved for that feature
 
-## Getting started
+## Stability and traceability
+
+GERS IDs are stable (within a reasonable tolerance of error) across multiple versions of Overture data. When stability is not possible, traceability will be provided. For example:
```
ibnt1, line 46: i would clarify we don't have this yet. or remove it altogether for now till we have at least the design/schema for modelling traceability.

docusaurus/docs/gers/gers.mdx
```diff
@@ -33,44 +31,39 @@ D SELECT id FROM read_parquet('s3://overturemaps-us-west-2/release/2024-01-17-al
 │ 08ba00858a164fff020008cf146d28e6 │
 │ 08ba00858a16cfff0200271c60f0a5ea │
 │ 08ba00858a161fff02008633c8f27bf7 │
-│ 08ba00858a163fff0200bf648279e290 │
-│ 08ba00858a163fff02001756776dd6ca │
-│ 08ba00858a163fff0200e04e60516f5c │
-│ 08ba00858a16efff02008127dee56493 │
-│ 08ba00858a16efff020065a1f1e26a85 │
-│ 08ba00858a144fff020057e23ab30103 │
+│ 08ba00858a163fff0200bf648279e290 │ │
```
ibnt1, line 34: | leftover?

docusaurus/docs/gers/gers.mdx
```diff
@@ -5,26 +5,24 @@ label: GERS
...
+## What exactly is an entity?
+Overture defines an entity as a physical thing or concept in the real world, represented as a feature in a geospatial dataset. It could be a segment of road, a city boundary, a grocery store, a building, a park bench or a lamp post. Overture assigns unique IDs, called GERS IDs, to these features. In most cases it's helpful to think of an entity and a feature as the same thing, but in practice it can be more complicated. An entity in the world could be represented by multiple features in a dataset, and a feature in a dataset might be a representation of multiple entities in the world. For example, a school building and its entrances and exits might be considered a single entity in the world but could be represented as multple features in a dataset, each feature with its own GERS ID.
```
ibnt1, line 11: i know this was already there and yes, could be indeed, but i wonder if this may be confusing new users that we have these in our data set, since they exist in osm for example;
i would stick with examples from types of data we already have, afaik benches and lamps are not in overture data set and might not be there anytime soon or at all.

ibnt1 (approved):
looks good to me! 
left few notes for future consideration.

vcschapp:
@danabauer is going to make some updates on this. She will finalize it in the next few hours.

danabauer:
I need to fix the GERS scenarios page before this can be merged.

jwass (approved):

PR #118, opened 2024-02-06
add schema for landCover, base theme

jonahadkins:
## Proposal

Add multiple resolutions of global land cover data to base theme in Overture (sometime in 2024). Examples below from Meta's production basemaps.

Screenshot 2024-02-06 at 9 17 39 AM

**Low Resolution, minZoom=0, maxZoom=4**

Screenshot 2024-02-06 at 9 18 12 AM

**Medium Resolution, minZoom=4, maxZoom=9**

Screenshot 2024-02-06 at 9 20 16 AM

**High Resolution, minZoom=9, maxZoom=24**

### Background
The Display Maps team at Meta recently completed a project to integrate global land cover data into it's basemaps. Landcover data provides an opportunity o display continuous, gapless representation of the Earth's natural surfaces. We derived these products from the 2020 European Space Agency (ESA) raster land cover product. Our process was generally: (1) download the image tiles, (2) process them, (3) vectorize them, and finally (4) post-process them to a usable format for vector tiling. We publicly released these with our Daylight product last year ([Blog](https://daylightmap.org/2023/10/11/landcover.html)).

### For Discussion

(1) IMO this is a huge opportunity for Overture to provide a massive dataset, critical to every global basemap used in a multitude of mapping applications.  This would be a first-of-its kind public dataset, it's hard to develop for day-to-day mappers, and is a huge technical lift at global scale.

Examples of land cover in production across several providers.
compare

(2) Providing multiple resolutions meets the needs of map developers and cartographers who would use this data in their workflows. We could provide only the highest resolution, but users would need to develop complex processes to generalize and simplify the data. Providing only medium or low resolution would fall short of expectations for this type of dataset. 

Providing tile zoom recommendations is probably a better way to define resolution than a specific reference scale like 1:1,000,000 or text like 'medium'.

(3) This is typically a "one and done" type dataset. This type of data is generally only needs refreshing every 5 to 10 years, and adding metadata such as `year` and `source` helps define that.

(4) While this type of data can be found in OpenStreetMap, this product provides continuous coverage across the globe.


vcschapp:
This sounds great!

I feel like the right place to initially have this discussion is within the `base` theme Task Force. Once they're aligned that:

1. this is a good thing to do; and
2. we can generate and provide the data once the schema is merged;

then we should push forward with it...

jonahadkins:
> I feel like the right place to initially have this discussion is within the `base` theme Task Force.

thanks @vcschapp - this PR is the result of discussion in that task force last week

BoZivk (commented):
Shouldn't this be referenced within schema/schema.yaml?

schema/base/landCover.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - 70.5310735
+          - 35.2655368
+
+
+
+
+
+
+
+
```
vcschapp, line 83: That's a whole lot of newlines! Is one enough?

schema/base/landCover.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - snow
+          - urban
+          - wetland
+      minZoom:
+        description: recommended minimum zoom according to bing tile zoom levels
+        type: number
+        minimum: 1
+        maximum: 24
+      maxZoom:
+        description: recommended maximum zoom according to bing tile zoom levels
```
vcschapp, line 45: "Bing tile" seems to be an inside-Meta term for just standard slippy map tile zooms... Is there a better name?

jonahadkins: i thought "bing tile" was a pretty standard term, but yes it corresponds to ["slippy map" tile names](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames) and the [original Google](https://www.maptiler.com/google-maps-coordinates-tile-bounds-projection/#3/-2.54/50.02) tile spec.

ill update accrodingly.

schema/base/landCover.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - forest
+          - grass
+          - mangrove
+          - moss
+          - shrub
+          - snow
+          - urban
+          - wetland
+      minZoom:
+        description: recommended minimum zoom according to bing tile zoom levels
```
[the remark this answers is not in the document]
jonahadkins: i've added [resolution](https://github.com/OvertureMaps/schema/pull/118/files#diff-93a22edf02ed0f205d46d9f18da43e699ec780037852e240283dd126d413fab1R49) as an optional column - resolution feels weird by itself since they are not round units. Zoom level, imo, feels pretty standard for anyone building maps. 

since the proposed data set is multi-resolution, recommended zooms + resolution gives the user all the necessary information

BoZivk (approved):

schema/base/landCover.yaml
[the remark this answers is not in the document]
jonahadkins: yeah i think this works well

BoZivk (approved):

jenningsanderson (dismissed):

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+          minimum: 1
+          maximum: 24
+        suggested_max_zoom:
+          description: recommended maximum tile zoom
+          type: number
+          minimum: 1
+          maximum: 24
+        suggest_hidden:
+          description: true/false if the feature is suggested to be hidden for cartographic display
+          type: boolean
```
jwass, line 50: If the cartography property will be shared by divisions (for suggesting hiding disputed) this will have to move the top-level defs and then referred here.

I think suggest_hidden won't be used at all for land_cover?

jonahadkins: agreed, will update shortly

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
+---
+"$schema": https://json-schema.org/draft/2020-12/schema
+title: Land Cover Schema
```
vcschapp, line 3: This should be just `land_cover`.

(We changed the way all the feature titles are done...)

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - "$ref": https://geojson.org/schema/MultiPolygon.json
+  properties:
+    unevaluatedProperties: false
+    allOf:
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/levelContainer
+    required:
+      - subType
+    properties:
+      subType:
```
vcschapp, line 23: Needs update to snake case: `subtype`.

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - crop
+          - forest
+          - grass
+          - mangrove
+          - moss
+          - shrub
+          - snow
+          - urban
+          - wetland
+      cartography:
```
vcschapp, line 37: Comments/ideas:

1. [ ] Isn't this invalid/ignored JSON schema? I think `cartography` requires below it a `type: object` clause, and a `properties: ...` clause under which the properties should be listed?
2. [ ] Should `cartography` be `required` for land_cover?
3. [ ] Should `cartography` have a description? Along the lines of "Contains cartographic hints for optimum use of the feature within a map"...
4. [ ] A possible reason why #1 and #2 are causing issues is because of a lack of examples/counterexamples for `land_cover`. Can you add some?

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+        suggested_min_zoom:
```
vcschapp, line 38: I think `min_zoom`/`max_zoom` probably suffices if we're clear in the description of `cartography` that it is cartographic hints/suggestions.

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+          description: recommended minimum tile zoom
```
vcschapp, line 39: From my limited research it's probably safe to say that these zoom levels are compatible with the Slippy Maps and Bing Maps tile standards.

schema/base/land_cover.yaml
```diff
@@ -0,0 +1,50 @@
...
+          type: number
```
vcschapp, line 40: Should the type be `integer` to just clamp to discrete integer zooms?

schema/base/land_cover.yaml
vcschapp: From the [2024-04-17](https://github.com/OvertureMaps/schema-wg/issues/293) schema TF meeting, seems like we need to drop `suggest_hidden`, define the `cartography` property as a "container" in the main `defs.yaml` and then include it into this schema by reference in the `allOf` clause up around line 17...

----

I realized I was wrong in proposing mix-and-match sub-properties in `cartography` per feature type. That won't work because it will probably break the Parquet schema generator.

vcschapp (commented):
Can we add a couple of examples and counter-examples specific to the `land_cover` feature type?

jenningsanderson:
@jonahadkins Wasn't there a version of this where min-max zoom level was expressed instead as a scale?

jonahadkins:
> @jonahadkins Wasn't there a version of this where min-max zoom level was expressed instead as a scale?

originally,  but we thought it was more confusing

schema/defs.yaml
```diff
@@ -313,3 +314,21 @@ description: Common schema definitions shared by all themes
         moving along the line from beginning to end
       properties:
         side: { "$ref": "#/$defs/propertyDefinitions/side" }
+    cartographyContainer:
+      title: cartography
+      description: defines cartograhic hints for optimal use of Overture features in map making
```
vcschapp, line 319: Cool.

I know it seems nitty, but would you mind punctuating the descriptions into classic sentence case (starting capital, ending full stop)?

schema/defs.yaml
```diff
@@ -313,3 +314,21 @@ description: Common schema definitions shared by all themes
...
+        cartography:
+          type: object
+          properties:
+            min_zoom:
+              description: recommended minimum tile zoom per the google tiling specification
+              type: integer
+              minimum: 0
+              maximum: 23
+            max_zoom:
+              description: recommended maximum tile zoom per the google tiling specification
```
vcschapp, line 331: Can we do sentence case, capitalize "G" for Google, and link a URL if possible?

Just curiosity: Why did you go with Google rather than Slippy Maps/Bing?

jwass (approved):
Looking forward to getting this out there!

schema/defs.yaml
jonahadkins: Bing is off by one (starts at 1 instead of 0), the vector tile spec starts at 0 as defined by Google

vcschapp (dismissed):

jwass:
PR for data license approval: https://github.com/OvertureMaps/tf-data-platform/pull/253

vcschapp:
In schema meeting [2024-05-01](https://github.com/OvertureMaps/schema-wg/issues/302), @jwass indicated including ESA WorldCover for use as landcover is OK by legal. So this is not a blocker for merging.

schema/defs.yaml
```diff
@@ -313,3 +314,21 @@ description: Common schema definitions shared by all themes
...
+    cartographyContainer:
+      title: cartography
+      description: Defines cartograhic hints for optimal use of Overture features in map making.
+      type: object
+      properties:
+        cartography:
+          type: object
+          properties:
+            min_zoom:
+              description: Recommended minimum tile zoom per the [Google Maps](https://www.maptiler.com/google-maps-coordinates-tile-bounds-projection/#3/15.00/50.00) tiling specification.
```
DavidKarlas, line 326: Since markdown links are not working in documentation, could you delete [Google Maps] part and just leave blank URL?

vcschapp (approved):

DavidKarlas (approved):

PR #128, opened 2024-02-22
Schema: Remove mixed types: road.surface (Parquet compatibility)

TheRBajaj:
# Description

This PR removes the mixed schema types for `roads.surface` property

### Before
`roads.surface` could accept multiple types of values that could be put into the same property like array and string values.

### After

`roads.surface` only accepts array property instead of array and string

# Reference

https://github.com/OvertureMaps/tf-transportation/issues/57

# Testing

Tested using test script


schema/transportation/segment.yaml
```diff
@@ -147,18 +147,16 @@ properties:
           description: Physical surface of the road. May either be
             specified as a single global value for the segment, or as
             an array of surface rules.
-          oneOf:
```
sanjibdutta1, line 150: don't we require similar fix for flags?

TheRBajaj: yup I will have another PR out for the flags fix too. We decided to break this issue up into multiple PR's

brad-richardson (dismissed):
Nice!

schema/transportation/segment.yaml
```diff
@@ -147,18 +147,16 @@ properties:
...
+          items:
+            type: object
+            allOf:
+              - { "$ref": "../defs.yaml#/$defs/propertyContainers/atRangeContainer" }
+            unevaluatedProperties: false
+            properties:
+              value: { "$ref": "#/$defs/propertyDefinitions/roadSurface" }
+          minItems: 1
+          uniqueItems: true
           default:
```
vcschapp, line 160: This `default` value makes no sense now - can you delete it?

vcschapp (changes requested):
Thanks Rahul. Couple small things:

1. [x] The `default` value should be deleted.
2. [x] The counterexample [`bad-road-surface-invalid-flat-value.json`](https://github.com/OvertureMaps/schema/blob/18023ca39cbee55df604897d8d22689dce54f574/counterexamples/transportation/segment/road/bad-road-surface-invalid-flat-value.json) is no longer applicable and should be deleted.

marinrelatic (approved):

brad-richardson (approved):

vcschapp (approved):

sanjibdutta1 (approved):

PR #130, opened 2024-02-23
Divisions

vcschapp:
# Description

This change implements the [Feature-Oriented Divisions](https://wiki.overturemaps.org/x/nwEjAQ) proposal as a new divisions theme, *a.k.a.* "admins 2.0", within the Overture JSON Schema model.

There are some minor departures from the design proposal. These are described in the **Notes** section at the bottom of this PR along with their rationale. 👇

# Reference

1. https://github.com/OvertureMaps/schema-wg/issues/282 (Issue under which this change is made)
2. [Feature-Oriented Divisions](https://wiki.overturemaps.org/x/nwEjAQ) proposal (Overture Confluence wiki)

# Testing

1. Added relevant examples.
2. Added relevant counterexamples. (Complete as of 2024-02-26.)

# Checklist

1. [x] Add relevant examples.
3. [x] Add relevant counterexamples.
5. [ ] Update Docusaurus documentation, if an update is required. *Since we likely won't have data in for the March release, there is no additional Docusaurus documentation at this time.*
6. [ ] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation. *I am working with @danabauer on this. We will meet on 2024-02-27 to start brainstorming the docs layout.*

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/130)

# Notes

As mentioned up there somewhere, this PR introduces minor deviations from the [Feature-Oriented Divisions]() proposal design document. These deviations and their justifications are listed below.

1. `division`:
   - Rename `subDivision` (`sub_division` after the recent snake_case transition) to `region` as proposed by Jake, because it is a nicer name, and because it aligns with what's in the place address field now.
    - Drop the `demographics` property as I agree with @mojodna that it's a Trojan Horse for non-map data and even population is suspect.
   - Promote `population` to a top-level property since its home in `demographics` is now gone.
   - Localize `local_type` as both @jwass and @mojodna  pointed out the localization issues there, i.e. how the locals refer to "canton" will depend on what language they speak.
   - Add `capital_division_id` at @jwass' suggestion.
   - Drop `default_language` from `norms`. This may need to go back in, but I am hopelessly confused as to its real-world utility and just want to sync with MSFT first.
3. `area`:
   - Rename the feature type to `division_area`. This is to front-run a likely decision in the Schema TF to make feature type names globally unique, in which case `area` and `part` (buildings) are going to be found to be too generic. (See minutes of [2024-02-21](https://github.com/OvertureMaps/schema-wg/issues/278) TF meeting for recent discussion.)
   - Drop the `flags` property and replace it with a `class` property. This feels more usable to me (very subjective) but is also responsive to feedback from @jwass and @sasastanojkov, who made flag-critical feedback comments in the design doc.
4. `boundary`:
   - Repeat the `subtype` property from the other feature types, because it facilitates certain types of filtering, making some display use cases easier without a join.
   - Repeat the `class` change from `division_area` for similar reasons and because it creates more alignment between the feature types.
   - Drop the `isDisputed` and `isAmbiguous` flag ideas and instead allow one or both sides of the boundary to be marked disputed. We had some follow-up discussion about this during the Admins Task Force meeting on [2024-02-26](https://github.com/OvertureMaps/tf-admin/issues/50).


schema/defs.yaml
```diff
@@ -189,33 +200,58 @@ description: Common schema definitions shared by all themes
         language: { "$ref": "#/$defs/propertyDefinitions/language" }
         value:
           type: string
-    names:
+          minLength: 1
+          pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
+    allNames:
```
sasastanojkov, line 205: Why extract common names as a separate property? As far as I see, there is no clear benefit from it. Or am I missing something?

schema/defs.yaml
```diff
@@ -189,33 +200,58 @@ description: Common schema definitions shared by all themes
...
+          description: >-
+            Rules for names that cannot be specified in the simple
+            common names property. These rules can cover other name
+            variants such as official, alternate, and short; and they
+            can optionally include geometric scoping (linear
+            referencing) and side-of-road scoping for complex cases.
+          type: array
+          items: { "$ref": "#/$defs/propertyDefinitions/nameRule" }
+          minItems: 1
+    commonNames:
```
jwass, line 222: This name is a little confusing as I thought it might just be the common map but I see it's the actual full names schema - just not in a `names` container. I'm fine with it.

schema/divisions/division.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+          This property is localized using a standard Overture names
+          structure. So for example, in Switzerland the top-level
+          administrative subdivision corresponding to subtype 'region'
+          is the canton, which is may be translated in each of
+          Switzerland's official languages as, 'canton' in French, 
+          'kanton' in German, 'cantone' in Italian, and 'chantun' in
+          Romansh.
+        allOf:
+          - "$ref": "../defs.yaml#/$defs/propertyDefinitions/commonNames"
```
jwass, line 53: It's nice to have this map but I wonder how necessary it is... and how well we'll be able to populate it but I like the consistency with the names structure

jwass (approved):

schema/defs.yaml
danabauer: @jwass what do you mean by `names` container?

schema/divisions/defs.yaml
```diff
@@ -0,0 +1,142 @@
+---
+"$schema": https://json-schema.org/draft/2020-12/schema
+description: Common schema definitions for divisions theme
+"$defs":
+  typeDefinitions:
```
sasastanojkov, line 5: So far we had propertyDefinitions. Any reason why this is now type definitions?

examples/divisions/division/country.yaml
```diff
@@ -0,0 +1,22 @@
...
+properties:
+  theme: divisions
+  type: division
+  update_time: "2024-02-23T15:08:01Z"
+  version: 0
+  subtype: country
+  names:
+    primary: United States
+  country: US
+  hierarchies:
```
sasastanojkov, line 16: Should countries have themselves in hierarchies or should they just have empty hierarchy?

schema/defs.yaml
sasastanojkov: I found my answer in division.locality_type

schema/divisions/division.yaml
```diff
@@ -0,0 +1,178 @@
...
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
+      - "$ref": ../defs.yaml#/$defs/propertyContainers/namesContainer
+      - if:
+          properties:
+            subtype: { enum: [country] }
+        then:
+          required: [names, subtype, country, hierarchies]
+          not: { required: [parent_division_id] }
+        else:
+          required: [names, subtype, country, hierarchies, parent_division_id]
```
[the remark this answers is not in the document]
vcschapp: This is a legitimate question. I'm not sure where the line should be drawn. 

The one thing I think we can fall back on is that it's is easier to *remove* a constraint than it is to *add* one, in the sense that if we relax that constraint, it won't break our data pipelines or anyone who is consuming the data. Constraints are roughly the opposite of properties: it's easy to add properties later but not remove them.

So one legitimate approach is just to leave it, and relax later if there's need. WDYT?

schema/divisions/division.yaml
vcschapp: Yes, I don't really know if we will be able to. I'll take that up with the admins group. If they think we can never get data for it, it seems like legitimate grounds for deletion.

schema/divisions/defs.yaml
vcschapp: Yeah, I think `propertyDefinitions` is wrong based on how we're using it. (And I think I'm probably the one who created `propertyDefinitions`, so my mistake.)

Basically in the top-level `defs.yaml`, mostly what we're defining is reusable "types" that are then used as the schemas for properties. But we're not defining the properties themselves, just the types/schemas. So I think it would be more correct if we migrate to the `typeDefinitions` terminology.

examples/divisions/division/country.yaml
vcschapp: Good question. The [design document](https://wiki.overturemaps.org/x/nwEjAQ) suggested including the feature itself in the hierarchy, so this is being faithful to the design doc.

IMO the benefit of including it is its a nice consistent way to present the data. If someone wants to look at the full hierarchy for a feature, they don't have to construct it by inserting the feature into the hierarchy, they can just immediately see the whole thing at a glance.

schema/defs.yaml
vcschapp: @danabauer, by "`names` container", Jake means a predefined reusable schema called `namesContainer` that you can reference from feature type schemas to instantly add the standard `names` into that feature type.

vcschapp: @jwass it's not the full names schema, it's the `primary` name plus the `common` names, but minus `rules`. There's another thingy called `allNames` directly above it that includes both `commonNames` and the `rules`.

Let me know if you can think of a better name for `commonNames` - I'm not married to it by any means.

schema/divisions/division.yaml
vcschapp: Added to the agenda for Monday's meeting ([2024-03-04](https://github.com/OvertureMaps/tf-admin/issues/51)).

schema/base/land.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 "$schema": https://json-schema.org/draft/2020-12/schema
-title: Land Schema
+title: land_use
```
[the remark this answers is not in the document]
vcschapp: Fixed and updated. Thanks!

examples/divisions/division/country.yaml
sasastanojkov: Ok, makes sense. 
I would be on the team "do not put feature itself into hierarchy", but I am ok either way.

schema/divisions/defs.yaml
sasastanojkov: I like the `propertyDefinitions` name since it is just there to define possible properties that can be used by different features. In any case, I would like to have single value everywhere, whether it is `type` or `property` definitions. Having two different names doing the same thing seems messy.

sasastanojkov (approved):

DavidKarlas:
I'm not seeing division in https://dfhx9f55j8eg5.cloudfront.net/pr/130/reference?

schema/defs.yaml
jwass: I think I'm fine with it especially because `commonNames` isn't exposed anywhere as an attribute name so it's just for people looking over the schema.

jwass (approved):

vcschapp:
> I'm not seeing division in https://dfhx9f55j8eg5.cloudfront.net/pr/130/reference?

I deliberately left it out. It'll be a follow-up. I'm meeting with Dana to chat about the docs.

schema/divisions/defs.yaml
vcschapp: My point is that they aren't *properties*, they are types/schemas that may be used *for* properties, which is different.

Agreed it seems messy. I intend to refactor the other parts of the schema to align, but didn't want to change it all at once.

MarkLebegge-TomTom (approved):

schema/divisions/division.yaml
vcschapp: The outcome of the meeting was that we will merge the PR as-is and I will follow fast with a smaller that removes the `primary` property, so there is just the names map.

This will resolve the difficult conundrum of what to set as "primary" in a multi-lingual country, by simply removing it.

PR #138, opened 2024-03-01
Remove mixed schema types for minOccupancy property

TheRBajaj:
# Description

This PR removes the mixed schema types for the `min_occupancy` property in `segments.yaml`

### Before
`min_occupancy` property could accept multiple types of values that could be put into the same property like array and string values.

### After
`min_occupancy` only accepts array property instead of array and string

`min_occupancy` currently does not have any combobulation changes so a combobulation PR will not be made.

# Reference

https://github.com/OvertureMaps/tf-transportation/issues/57

# Testing

Tested using test script

# Checklist

1. [x] Add relevant examples.
3. [x] Update in-schema documentation using plain English written in complete sentences, if an update is required.
4. [x] Update Docusaurus documentation, if an update is required.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/<138>)


ibnt1 (dismissed):

counterexamples/transportation/segment/road/bad-road-invalid-min-occupancy.json
```diff
@@ -0,0 +1,33 @@
+{
+    "id": "invalidMinOccupancy",
```
marinrelatic, line 2: nit: It's useful to specify in the name why is the `minOccupancy` invalid. 
e.g. 
- `bad-road-invalid-min-occupancy-type.json`
- `"id": "invalidMinOccupancyType"`

Not a blocker.

marinrelatic (approved):

schema/transportation/segment.yaml
```diff
@@ -76,15 +76,13 @@ properties:
...
-                    allOf:
-                      - "$ref": "#/$defs/propertyContainers/ruleContainer"
-                      - "$ref": "#/$defs/propertyDefinitions/relationalExpression"
+              type: array
+              items:
+                type: object
+                unevaluatedProperties: false
+                allOf:
+                  - "$ref": "#/$defs/propertyContainers/ruleContainer"
+                  - "$ref": "#/$defs/propertyDefinitions/relationalExpression"
```
vcschapp, line 85: I think this one is more badly broken than I previously thought. (My apologies @TheRBajaj, this isn't your issue, this is just a conceptual problem with the schema, but we should fix it.)

The whole `relationalExpression` thing makes no sense here.

*i.e.* It makes no sense to have rules like:

```yaml
min_occupancy:
 - is_at_least: 3
```

That's because a `min` or minimum already implies that something "is at least". Also, if we allow relational, we can get absurdity:

```yaml
min_occupancy:
 - is_at_most: 3
```

That doesn't make any sense: it's putting a maximum limit on a minimum!

I think what we need, and again apologies @TheRBajaj I didn't know this aspect of the schema was in such rough shape, is to dump the whole `relationalExpression` thing and just have a `value` property, which must be an integer, of value at least `1`.

So the rules should look like:

```yaml
min_occupancy:
 - value: 3
 - when: { mode: [truck] }
   value: 4
```

Could you make this change? It basically requires:

1. [ ] Drop current line 85.
2. [ ] Add a `properties` section at line `85` that has one field, `value`, that has integer type and a `minimum` of `1`.
3. [ ] Repair the examples/counter-examples accordingly.

vcschapp (changes requested):
My apologies, I have to request a change simply because the existing schema was broken a bit more fundamentally than I thought. Could you see my comment above?

schema/transportation/segment.yaml
TheRBajaj: got it, not a problem at all! will make the necessary changes.

examples/transportation/segment/road/lanes/restrictions/lanes-hov-occupancy-scoped.yaml
```diff
@@ -38,19 +38,11 @@ properties:
...
-                mode:
-                  - hov
-                recognized:
-                  - as_permitted
-            - is_at_least: 1
-              when:
-                mode:
-                - bicycle
+            - value: 3
+              when: { mode: [hov] }
```
marinrelatic, line 42: Why mixing the format instead of having?
```
- value: 3
    when:
        mode:
            -hov
```

counterexamples/transportation/segment/road/bad-road-invalid-min-occupancy-value.json
```diff
@@ -0,0 +1,37 @@
...
+          }
+        ],
+        "lanes": [
+            {
+                "restrictions": {
+                    "min_occupancy": [
+                        {
+                            "value": 0
+                        }
+                    ]
```
vcschapp, line 27: Sorry for one further change, but I can't think of a good reason why `min_occupancy` shouldn't be a flat value. IMO we should just make it a flat integer with a minimum value of 1.

vcschapp (approved):

vcschapp (changes requested):
My apologies. I think we need to flatten `min_occupancy`. There's no good reason for it to be a list of rules. Can be just a flat integer. Apologies for not nailing this down correctly ahead of time.

marinrelatic (dismissed):

vcschapp (dismissed):
Ship it!

ibnt1 (approved):

brad-richardson (approved):

PR #142, opened 2024-03-11
Adding additional locality type names to admins/defs.yaml.

batinic-m:
Adding three additional locality type names to admin schema.

A bit of statistics to back the change:
'department' - appears as a type of locality in 4% of the world's admins.
'parish' and 'governorate' - each appear as a type of locality in 3% of the world's admins.
For comparison, 'state' appears as a locality type in 3.3% of admins, while 'district' is by far the most popular with 22%.

### DIFF
--- a/schema/admins/defs.yaml
+++ b/schema/admins/defs.yaml
@@ -7,7 +7,7 @@ description: Common schema definitions for admins theme
     localityType:
       description: Describes the entity's type in the categorical nomenclature used locally.
       type: string
-      enum: [country, county, state, region, province, district, city, town, village, hamlet, borough, suburb, neighborhood, municipality]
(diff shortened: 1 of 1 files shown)


vcschapp:
> Adding three additional locality type names to admin schema.
> 
> A bit of statistics to back the change: 'department' - appears as a type of locality in 4% of the world's admins. 'parish' and 'governorate' - each appear as a type of locality in 3% of the world's admins. For comparison, 'state' appears as a locality type in 3.3% of admins, while 'district' is by far the most popular with 22%.

@batinic-m @DavidKarlas this change would only hit in April data release, when we're expecting Divisions data to be populated. Why do we still want to keep iterating on the admins schema?

batinic-m:
@vcschapp David made me aware of this change, so I'll close the PR and create a relevant one. Thanks for the heads up.

PR #144, opened 2024-03-12
Min floor

jwass:
# Description

This parallels the "min_level" property in OpenStreetMap. https://wiki.openstreetmap.org/wiki/Key:building:min_level

Ideally there are two things that would make this property unnecessary:
* We would want to know the height instead of the start "level" which feels less useful
* This should only be on building parts and not on footprints. However there are cases where OSM buildings are not tagged as a part and it's difficult to just switch them. So we compromise and just include the min_floor field.

# Reference

# Testing

Added field to an example.

# Checklist

1. [x] Add relevant examples.
3. [ ] Update in-schema documentation using plain English written in complete sentences, if an update is required.
4. [ ] Update Docusaurus documentation, if an update is required.

# Documentation Website


schema/buildings/defs.yaml
```diff
@@ -20,6 +20,11 @@ shapeContainer:
       description: >-
         The height of the bottom part of building in meters. Used if a building or part of building starts above the ground level.
       type: number
+    min_floor:
+      description: >-
+        The "start" floor of this building or part. This is the analog of "min_height" but is sometimes populated when the
```
LukaszMichalski-TomTom, line 25: Is this split to multiple lines a problem ? It is different from other descriptions, that's why I am asking

LukaszMichalski-TomTom (approved):

schema/buildings/defs.yaml
vcschapp: Not a problem - "usually" we wrap the lines around 72 or 80 though, although that's pretty inconsistent.

schema/buildings/defs.yaml
```diff
@@ -20,6 +20,11 @@ shapeContainer:
...
+        start height is unknown.
+      type: integer
```
vcschapp, line 27: Any validation desired? *e.g.* `minimum` of zero, or `minimum` of one?

How are basements typically handled? Or floors that are not labeled with a number, P2, P3, B, B2, sub-basement, L, mezzanine and so forth?

jwass: I took a look at OSM. There are ~137000 ways that have a building:min_level. Less than 1000 are less than zero. Given that - I think we should probably make this greater than 0 and omit the value when it's less than zero. This includes our understanding that this number should be positive anyway.

Basements and other underground levels (like parking levels) are a separate field in OSM that's not currently handled in Overture. But it does not affect the value of this field.

jwass: Added exclusiveMinimum.

schema/buildings/defs.yaml
```diff
@@ -20,6 +20,12 @@ shapeContainer:
       description: >-
         The height of the bottom part of building in meters. Used if a building or part of building starts above the ground level.
       type: number
+    min_floor:
+      description: >-
+        The "start" floor of this building or part. This is the analog of "min_height" but is sometimes populated when the
+        start height is unknown.
```
vcschapp, line 26: Should we update this description a bit based on what we learned looking at the docs page in schema TF meeting the other day?

It seems to serve a different purpose than `min_height`, although admittedly in the rendering use case you can presumably use `min_floor` as a proxy for `min_height`...

Maybe along the lines of:

> The "start" floor of this building or part. Indicates that the building or part is "floating" and its bottom-most floor is above ground level, usually because it is part of a larger building in which some parts do reach down to ground level. An example is a building that has an entry road or driveway at ground level into an interior courtyard, where part of the building bridges above the entry road. This property may sometimes be populated when `min_height` is missing and in these cases can be used as a proxy for `min_height`.

jenningsanderson (dismissed):

LukaszMichalski-TomTom (dismissed):

jenningsanderson (approved):

jonahadkins (approved):

DavidKarlas (approved):

vcschapp (approved):

PR #162, opened 2024-04-11
Extend perspectives property to indicate inclusion

sasastanojkov:
# Description

Current perspectives representation allows for same entity to be simultaneously allowed by set of countries and disputed by another set. This would require us to do additional checks and would be overall inconsistent. Entity could have holder X allowing it, holder Y disputing it, and what about any other holder Z? To resolve this, single entity can be allowed by set 
of countries and disputed by the rest of the world by default or disputed by set of countries and allowed by the rest of the world.
This will allow us to say perspective from country X covers all entities having that country as allowed by or not having it as disputed by perspective.
In that sense it will make data more consistent by default.
We currently only support countries to dispute entities but are open to extending list of entity types which could dispute any entity.

Example (also in test):
Previous implementation:
```
  perspectives:
    - type: country
      holder: TR
```
Current implementation:
```
  perspectives:
    mode: accepted_by
    countries: [TR]
```

# Checklist

1. [x] Add relevant examples.
3. [x] Update in-schema documentation using plain English written in complete sentences, if an update is required.
4. [ ] Update Docusaurus documentation, if an update is required.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/162)

### DIFF
--- a/schema/divisions/defs.yaml
+++ b/schema/divisions/defs.yaml
@@ -89,6 +89,26 @@ description: Common schema definitions for divisions theme
           type: string
           minLength: 1
           pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
+    perspectives:
+      description: Collection of perspectives of single feature.
+      type: object
+      unevaluatedProperties: false
+      required: [holders, includes]
+      properties:
+        holders:
+          description: Array of perspective holders.
+          type: array
+          items: { "$ref": "#/$defs/typeDefinitions/perspective" }
+          minItems: 1
+          uniqueItems: true
+        includes:
+          description:
+            Indicates if all counteries from the collection
+            or all except those in collection acknowledge the feature as is.
+
+            If value is true, only holders within collection see entity as it is represented.
+            If value is false, everyone except those in collection see entity as it is represented.
+          type: boolean
     perspective:
       description:
         A political perspective from which the division is viewed
--- a/schema/divisions/defs.yaml
+++ b/schema/divisions/defs.yaml
@@ -90,17 +90,17 @@ description: Common schema definitions for divisions theme
           minLength: 1
           pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
     perspectives:
-      description: Collection of perspectives of single feature.
+      description: Political perspectives from which division is viewed.
       type: object
       unevaluatedProperties: false
       required: [mode, countries]
       properties:
         mode:
-          description: Mode in which perspective should be used.
+          description: Whether perspective holder accept or dispute this division.
           type: string
           enum: [accepted_by, disputed_by]
         countries:
-          description: Array of counteries with given perspective.
+          description: Countries holding the given mode of perspective.
           type: array
           items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/iso3166_1Alpha2CountryCode" }
           minItems: 1
--- a/schema/divisions/defs.yaml
+++ b/schema/divisions/defs.yaml
@@ -93,53 +93,18 @@ description: Common schema definitions for divisions theme
       description: Collection of perspectives of single feature.
       type: object
       unevaluatedProperties: false
-      required: [holders, effect]
+      required: [mode, countries]
       properties:
-        holders:
-          description: Array of perspective holders.
+        mode:
+          description: Mode in which perspective should be used.
+          type: string
+          enum: [accepted_by, disputed_by]
+        countries:
+          description: Array of counteries with given perspective.
           type: array
-          items: { "$ref": "#/$defs/typeDefinitions/perspective" }
+          items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/iso3166_1Alpha2CountryCode" }
           minItems: 1
           uniqueItems: true
-        effect:
-          description:
-            Indicates if all perspective holders from the collection
-            or all except those in collection acknowledge the feature as is.
-
-            If value is "include", only holders within collection see entity as it is represented.
-            If value is "exclude", everyone except those in collection see entity as it is represented.
-          type: string
-          enum: [include, exclude]
-    perspective:
-      description:
-        A political perspective from which the division is viewed
-      type: object
-      unevaluatedProperties: false
-      required: [type, holder]
-      oneOf:
-        - if:
-            properties:
-              type: { enum: [country] }
-          then:
-            properties:
-              holder: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/iso3166_1Alpha2CountryCode" }
-          else:
-            false
-      properties:
-        type:
-          description:
-            Type of perspective
-
-            If the perspective is of type 'country', then the 'holder'
-            property must be an ISO 3166-1 alpha-2 country code
-            identifying the country that holds this perspective.
-          type: string
-          enum:
-            - country   # The perspective is held by the sovereign state
-                        # identified by the ISO-3166-1 alpha-2 country
-                        # code in the 'value' property.
-        holder:
-          type: string
     boundarySideDivision:
       description:
         Represents a division on one side or the other of a boundary
--- a/schema/divisions/defs.yaml
+++ b/schema/divisions/defs.yaml
@@ -93,7 +93,7 @@ description: Common schema definitions for divisions theme
       description: Collection of perspectives of single feature.
       type: object
       unevaluatedProperties: false
-      required: [holders, includes]
+      required: [holders, effect]
       properties:
         holders:
           description: Array of perspective holders.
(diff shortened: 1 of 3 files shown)


vcschapp (commented):
I [summarized what I think the current approach is](https://github.com/OvertureMaps/schema/discussions/117#discussioncomment-9146307) is in the Ongoing divisions discussion.

## Level-setting

Before going further, I want to make sure we're on the same page as to the before-and-after state if this PR were to be merged.

Let's say the country of Bar is not recognized by the country of Foo, but it is generally recognized and is considered to be in the default perspective.

- In the current state of our schema, we would need two instances of Bar. One would have no `perspectives` and would be in the default perspective. The other one would have a `perspectives` property that enumerates every country *except* Foo.
- If we merged this change, then we would still need two instances of Bar to solve the problem. Again, one would be in the default perspective. The difference is that the other one would have a single "exclusionary" perspective referencing Foo rather than an entry for every other country but Foo.

## Purpose of this PR

I am inferring that the purpose of this PR, then, is to reduce the required verbosity for the non-default copy of Bar.

## Request

Can you update the PR description:

- [x] Explaining what the change accomplishes.
- [x] Giving an example both before and after the change. You can use mine if it helps.
- [ ] Elaborating on why the change is necessary.

schema/divisions/defs.yaml
```diff
@@ -89,6 +89,26 @@ description: Common schema definitions for divisions theme
           type: string
           minLength: 1
           pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
+    perspectives:
```
vcschapp, line 92: Why move this property into `defs.yaml`. It isn't expected to be used anywhere else, which is why it was modeled directly into `division`.

schema/divisions/defs.yaml
```diff
@@ -89,6 +89,26 @@ description: Common schema definitions for divisions theme
...
+      unevaluatedProperties: false
+      required: [holders, includes]
+      properties:
+        holders:
+          description: Array of perspective holders.
+          type: array
+          items: { "$ref": "#/$defs/typeDefinitions/perspective" }
+          minItems: 1
+          uniqueItems: true
+        includes:
```
vcschapp, line 104: For symmetry, I prefer `effect: "include" | "exclude"` rather than `includes: true | false`.

schema/divisions/defs.yaml
```diff
@@ -89,6 +89,26 @@ description: Common schema definitions for divisions theme
...
+          description:
+            Indicates if all counteries from the collection
+            or all except those in collection acknowledge the feature as is.
+
+            If value is true, only holders within collection see entity as it is represented.
+            If value is false, everyone except those in collection see entity as it is represented.
```
vcschapp, line 110: On reflection, I think this isn't going to work as written. If we think this change is very important to pursue, we will need to think harder about the model and/or the documentation of what it means.

A superficial issue with the description is that it refers to *countries*, whereas the model is meant to be forward-compatible with adding new entity types, for example supranational entities like the UN or other blocs of countries, or maybe entities that don't have a recognized ISO-3166-1 alpha-2 country code but view themselves as being countries.

If you replace *countries* with *perspective holders* it seems to fix the superficial issue, but then we hit the conceptual issue.

The description says:

> If value is false, everyone except those in collection see entity as it is represented.

But who is *everyone*?

If we believe that only countries will ever have perspectives, that works. *Everyone* is *all countries* and then we can reason with "all countries except the ones listed".

But if other things can have perspectives, like supranational entities, then who is *everyone*? Does *everyone* include the UN? If yes, then does just saying "exclude RU" also imply "exclude UN"?

----

My conclusion is that the positive include model is forward-compatible with different holder types, but the include/exclude model is not.

So I think we have to ask ourselves:

1. How likely it is that we will need other types of perspective holders other than countries that have country codes.
2. How important is the exclusion mechanism? Is it a nice-to-have or a must-have?
3. Is there a third way to model this that we haven't thought of yet?

schema/divisions/defs.yaml
sasastanojkov: It is right above `perspective` I wanted to keep them coupled as I don't see a point in using one without the other.

schema/divisions/defs.yaml
DavidKarlas: 1. If I understanding this question is referring to `Indicates if all countries from the collection`, I think we can change this to different wording or we can leave it as is, and update once we add new types of "holders" if ever.
2. Must-have, assume we have 2 Argentina entities, 
   i. One that does not contain part of Antartica, this one will have "Exclude AR"
   i. And another one that will contain part of Antartica that will have "Include AR"
I can't see how we could do it without exclude.
3. Not that I know of.

vcschapp: > Must-have, assume we have 2 Argentina entities,
> One that does not contain part of Antartica, this one will have "Exclude AR"
> And another one that will contain part of Antartica that will have "Include AR"
> I can't see how we could do it without exclude.

You would have one Argentina that is from Argentina's perspective ("AR"), and one Argentina that is from everyone else's perspective (all other country codes).

schema/divisions/defs.yaml
vcschapp: Seems unrelated to the main purpose of the PR.

sasastanojkov: It will be used in boundaries since they will also depend on the perspective of the given country. I will add support for that in the next PR.

schema/divisions/defs.yaml
DavidKarlas: PR as updated, can this conversation be resolved?

jwass (dismissed):

schema/divisions/defs.yaml
```diff
@@ -89,36 +89,22 @@ description: Common schema definitions for divisions theme
...
-            - country   # The perspective is held by the sovereign state
-                        # identified by the ISO-3166-1 alpha-2 country
-                        # code in the 'value' property.
-        holder:
+        mode:
+          description: Mode in which perspective should be used.
           type: string
+          enum: [accepted_by, disputed_by]
+        countries:
+          description: Array of counteries with given perspective.
```
vcschapp, line 103: Spelling: `counteries` -> `countries`

----

Could consider simplifying to: "Countries holding the given mode of perspective."

schema/divisions/defs.yaml
```diff
@@ -89,36 +89,22 @@ description: Common schema definitions for divisions theme
           type: string
           minLength: 1
           pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
-    perspective:
-      description:
-        A political perspective from which the division is viewed
+    perspectives:
+      description: Collection of perspectives of single feature.
```
vcschapp, line 93: Consider changing to *Political perspectives from which division is viewed.* ...

schema/divisions/defs.yaml
```diff
@@ -89,36 +89,22 @@ description: Common schema definitions for divisions theme
...
-            property must be an ISO 3166-1 alpha-2 country code
-            identifying the country that holds this perspective.
-          type: string
-          enum:
-            - country   # The perspective is held by the sovereign state
-                        # identified by the ISO-3166-1 alpha-2 country
-                        # code in the 'value' property.
-        holder:
+        mode:
+          description: Mode in which perspective should be used.
```
vcschapp, line 99: Consider updating description to *Whether perspective holder accept or dispute this division.*

vcschapp (commented):
Looks good to me. Minor comments on the description text, but I'm good with the substance.

DavidKarlas (dismissed):

TristanDiet-TomTom (approved):

jwass (approved):

MarkLebegge-TomTom (commented):
same thought as Vic, " not accepted by" makes more sense than a full list of countries that accept default view.

sasastanojkov:
> same thought as Vic, " not accepted by" makes more sense than a full list of countries that accept default view.

That is why we now have `enum: [accepted_by, disputed_by]`

MarkLebegge-TomTom (approved):

schema/divisions/defs.yaml
```diff
@@ -90,17 +90,17 @@ description: Common schema definitions for divisions theme
...
     perspectives:
-      description: Collection of perspectives of single feature.
+      description: Political perspectives from which division is viewed.
       type: object
       unevaluatedProperties: false
       required: [mode, countries]
       properties:
         mode:
-          description: Mode in which perspective should be used.
+          description: Whether perspective holder accept or dispute this division.
```
vcschapp, line 99: Nit: should be holder*s*, plural.

Maybe can fix that in a subsequent PR?

vcschapp (approved):

PR #163, opened 2024-04-11
Removed country and parent_division_id as required properties

sasastanojkov:
# Description

We have and will perhaps have more entities where we cannot clearly define whole hierarchy leading from country to given locality, which in term leads us not to have proper parent and country property set.
Due to this, I removed information that properties `parent_division_id` and `country` are considered required.
One of those examples is Jerusalem.
We will have this check on all other entities using some other validation mechanism, not through schema validity check.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/163)


sasastanojkov:
> You mentioned the example of Jerusalem. Why can't Jerusalem be represented with the current `perspectives` schema? Is it imperative that there be a division representing Jerusalem as a "floating city" with no parent? Why?

Jerusalem is considered a special entity governed by UN, which in terms means it doesn't have country, More on the topic:
https://www.britannica.com/topic/United-Nations-Resolution-181
Since it is "Corpus separatum" the rules that govern other divisions don't really apply to it.

vcschapp:
> > You mentioned the example of Jerusalem. Why can't Jerusalem be represented with the current `perspectives` schema? Is it imperative that there be a division representing Jerusalem as a "floating city" with no parent? Why?
> 
> Jerusalem is considered a special entity governed by UN, which in terms means it doesn't have country, More on the topic: https://www.britannica.com/topic/United-Nations-Resolution-181 Since it is "Corpus separatum" the rules that govern other divisions don't really apply to it.

Thanks - great source.

DavidKarlas:
Going to think a bit more on this, closing for now

PR #166, opened 2024-04-18
Segment routes

brad-richardson:
# Description

Adds support for segment routes as a top-level property. These are intended to capture [OSM road route relations](https://wiki.openstreetmap.org/wiki/Relation:route#Road_routes), materialized to relevant segments and optionally geometrically scoped.

# Reference

1. https://github.com/OvertureMaps/tf-transportation/issues/52

# Testing

Added new example, tested

# Checklist

1. [X] Add relevant examples.
6. [x] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/166)

### DIFF
--- a/examples/transportation/segment/road/road-with-route.yaml
+++ b/examples/transportation/segment/road/road-with-route.yaml
@@ -1,0 +1,22 @@
+---
+id: overture:transportation:segment:123
+type: Feature
+geometry:
+  type: LineString
+  coordinates: [[0, 0], [1, 1]]
+properties:
+  # Overture properties
+  theme: transportation
+  type: segment
+  update_time: "2024-04-10T00:00:00-05:00"
+  version: 1
+  subtype: road
+  class: motorway
+  connector_ids: [fooConnector, barConnector]
+  routes:
+    - name: I 95
+      network: US:I
+      ref: 95
+      symbol: https://upload.wikimedia.org/wikipedia/commons/6/61/I-95.svg
+      wikidata: Q94967
+      between: [0, 0.5]
--- a/examples/transportation/segment/road/road-with-route.yaml
+++ b/examples/transportation/segment/road/road-with-route.yaml
@@ -16,7 +16,7 @@ properties:
   routes:
     - name: I 95
       network: US:I
-      ref: 95
+      ref: "95"
       symbol: https://upload.wikimedia.org/wikipedia/commons/6/61/I-95.svg
       wikidata: Q94967
       between: [0, 0.5]
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -49,6 +49,7 @@ properties:
             Each entry in this array is the GERS ID of a transportation connector feature.
         uniqueItems: true
         default: []
+      routes: { "$ref": "#/$defs/propertyDefinitions/routes" }
 "$defs":
   propertyDefinitions:
     roadClass:
@@ -308,6 +309,30 @@ properties:
         - dirt
         - paving_stones
         - metal
+    routes:
+      description: Routes this segment belongs to
+      type: array
+      items:
+        type: object
+        unevaluatedProperties: false
+        allOf:
+          - { "$ref": "../defs.yaml#/$defs/propertyContainers/geometricRangeScopeContainer" }
+        properties:
+          name:
+            description: Full name of the route
+            type: string
+          network:
+            description: Name of the highway system
+            type: string
+          ref:
+            description: Code or number referencing the route
+            type: ["string", "integer"]
+          symbol:
+            description: URL or description of route signage
+            type: string
+          wikidata:
+            description: Wikidata ID
+            type: string
     speed:
       description: >-
         A speed value, i.e. a certain number of distance units
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -326,7 +326,7 @@ properties:
             type: string
           ref:
             description: Code or number referencing the route
-            type: ["string", "integer"]
+            type: string
           symbol:
             description: URL or description of route signage
             type: string
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -321,18 +321,24 @@ properties:
           name:
             description: Full name of the route
             type: string
+            minLength: 1
+            pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
           network:
-            description: Name of the highway system
+            description: Name of the highway system this route belongs to
             type: string
+            minLength: 1
+            pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
           ref:
-            description: Code or number referencing the route
+            description: Code or number used to reference the route
             type: string
+            minLength: 1
+            pattern: ^(\S.*)?\S$    # Leading and trailing whitespace are not allowed.
           symbol:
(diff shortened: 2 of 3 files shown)


ylinzhi (dismissed):
🛳🛳🛳

ibnt1 (dismissed):
can you add a counterexample too?

examples/transportation/segment/road/road-with-route.yaml
```diff
@@ -0,0 +1,22 @@
...
+properties:
+  # Overture properties
+  theme: transportation
+  type: segment
+  update_time: "2024-04-10T00:00:00-05:00"
+  version: 1
+  subtype: road
+  class: motorway
+  connector_ids: [fooConnector, barConnector]
+  routes:
```
vcschapp, line 16: I think this should go under `road:`. The idea with `road` is to contain all the complex properties.

If we think it will also apply to subtype `rail` or `water` I'd be tempted to repeat it under those subtypes' `rail:` or `water:` property.

schema/transportation/segment.yaml
```diff
@@ -308,6 +309,30 @@ properties:
...
         - metal
+    routes:
+      description: Routes this segment belongs to
+      type: array
+      items:
+        type: object
+        unevaluatedProperties: false
+        allOf:
+          - { "$ref": "../defs.yaml#/$defs/propertyContainers/geometricRangeScopeContainer" }
+        properties:
```
vcschapp, line 320: Is there an minimal validation we can offer on any of these sub-properties?

For *e.g.*, is anything `required`? (Hopefully something.) And is there any other validation? We can usually at least do `minLength: 1` and `pattern: ^(\S.*)?\S$` ([example](https://github.com/OvertureMaps/schema/blob/7c9c6e7c807120345812b2a76d264b890d4bfc0f/schema/defs.yaml#L204))

schema/transportation/segment.yaml
```diff
@@ -308,6 +309,30 @@ properties:
         - dirt
         - paving_stones
         - metal
+    routes:
```
bastiaanvanassche-tomtom, line 312: This schema proposal does allow multiple road numbers for a single road: the `routes` property is an array but I was wondering if there is a possibility to indicate what the most relevant route number is. 
In Europe, roads typically have multiple route numbers (international, national, regional) and in public OSM there's a capability to indicate the most relevant one (e.g. international), see this example road in Belgium: https://www.openstreetmap.org/way/582460822
image

brad-richardson: Great point, the goal with this initial schema is just to provide the "raw" data for consumption as-is from OSM. I do think this should be part of the *next* iteration where we further process + schematize, adding properties like `network_country`, `network_is_intl`, etc as discussed in the task force.

I didn't want to tackle that additional schematizing because of the wide array of tagging patterns that need to be deeply understood. US highways for example have a similar intra vs interstate network hierarchy but that doesn't always come from dedicated tags although those do exist (`is_in:state`). Typically these are just value patterns - "US:I" or "US:MA" for example.

jwass: Could we indicate a priority by saying the first entry is the most relevant?

brad-richardson: What would we use to determine priority or relevance? I could definitely add a "best effort" ordering by international, national, regional, etc using some of these known tagging patterns without too much effort.

I still think the right way is to get a deeper understanding and have future schema improvements to capture this. Any priority we choose could be wrong depending on context

examples/transportation/segment/road/road-with-route.yaml
sanjibdutta1: Exactly the same reason you said Brad. Something like name (which is also a pretty complex property with multiple different name types) which is a top level property. Another reason is that its not directly related to transportation i.e. something that impacts routing which is what most of the road properties are.

ibnt1 (approved):

ylinzhi (approved):

examples/transportation/segment/road/road-with-route.yaml
vcschapp: This is what I'm working on, at a very slow pace. It's not really new, in that it is attempting to provide a coherent schema-wide description of how things (mostly) are, and how to make decisions about them.

https://wiki.overturemaps.org/x/0YIWAg#id-[Draft]SchemaStandards-NestedProperties

sanjibdutta1 (approved):

examples/transportation/segment/road/road-with-route.yaml
brad-richardson: @vcschapp I'm not sure benefits of nested JSON outweigh the negatives as the proposal currently stands. As I've started work with the data in depth, it has been difficult to answer some simple questions like "is foot access allowed?" in pure SQL. Here's an example query fragment I implemented to solve this:
```
JSON_ARRAY_CONTAINS(
  JSON_EXTRACT(
    TRY(JSON_PARSE(road),
    '$.restrictions.access[?(@.access_type==''allowed'')].when.mode[*]'
  ),
  'foot'
)
```
Relying on JSON parsing capabilities of query engines like this isn't ideal. Whereas if we brought something like `road.restrictions` to the top level:
```
ANY_MATCH(
  restrictions.access,
  restriction -> restriction.access_type = 'allowed' AND ANY_MATCH(restriction.when.mode, mode -> mode = 'foot')
)
```
which feels much more natural (to me 🙂) and doesn't require deep JSON path understanding. The biggest drawback to this approach (as detailed in your proposal) is an overwhelming # of columns and the potential complexity of these types for creating tables in diverse environments. Hopefully we can mitigate this with existing parquet metadata tools and better documentation.

One last thought - as our JSON object gets more complicated, the penalty to use that data in complex joins or filters grows as the full object must always be parsed and stored in memory. This means we also lose at least some of the benefit of a columnar-based storage.

brad-richardson: @vcschapp given the [schema discussion](https://github.com/OvertureMaps/schema-wg/issues/302#issuecomment-2088682544) today, can I consider this resolved in favor of top-level property structuring?

brad-richardson:
@vcschapp Follow-up here, any remaining concerns now that we've settled the JSON/road property splitting issue?

vcschapp (commented):
I'm basically leaving this alone right now until the nesting discussion gets resolved in a principled way.... Ideally we can make a final decision on nesting and all related churn in transportation will stop.

jwass (approved):

vcschapp (approved):

examples/transportation/segment/road/road-with-route.yaml
vcschapp: Let's consider this resolved in light of the plan @brad-richardson, @ibnt1, and myself aligned on yesterday to introduce some flattening of properties currently contained under `road`.

vcschapp:
> @vcschapp Follow-up here, any remaining concerns now that we've settled the JSON/road property splitting issue?

None! I think what we aligned on yesterday clears up any concerns.

For the record, [this comment thread on Overture Confluence](https://wiki.overturemaps.org/display/PROJ/Segment+Top-level+Properties?focusedCommentId=44662862#comment-44662862) summarizes the direction we want to go in with nesting.

PR #173, opened 2024-04-24
Fix counterexample use of map_file and verify

brad-richardson:
# Description

Counterexamples appear to be silently passing, even when expected errors are not present. This is in part due to `verify quiet` swallowing error codes and also `map_file` running in a subshell due to piping, failing to assign the variable where it is needed.

# Reference

https://superuser.com/questions/1348948/cant-pipe-in-bashs-mapfile-but-why

# Testing

Removed `ext_expected_errors` from counterexamples/buildings/bad-class.json and validated a failure was observed when running `./test.sh -m counterexamples "buildings/bad-class.json"`.

# Checklist

1. [ ] Add relevant examples.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/173)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -162,7 +162,7 @@ function examples() {
       continue
     fi
     printf "%s..." "$instance_file"
-    if verify quiet "$instance_file"; then
+    if verify simple "$instance_file"; then
       echo OK
     else
       echo FAILED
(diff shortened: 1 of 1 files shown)


vcschapp (commented):
Is it possible your encountering the issue fixed by this commit? https://github.com/OvertureMaps/schema/commit/16ed4351327de109b025c197d676501ce5db24e9

I just merged that commit to *dev* this morning, so if it's the same one, it's not unexpected that the two ships were passing in the night, so to speak.

PR #176, opened 2024-04-26
Transportation Destinations

ibnt1:
# Description

Adds support for segment destinations as a top-level property. 

These are intended to capture OSM [destination=*](https://wiki.openstreetmap.org/wiki/Key:destination) tags on ways and [destination_sign](https://wiki.openstreetmap.org/wiki/Relation:destination_sign) relations consolidated into a common schema.

Similar to the OSM destination tags on ways, destination property describes objects that are reached, when a the segment is being followed, but also includes references to the segments from which you need to transition to reach the destination, similar to the OSM destination_sign relations.

# Reference

1. https://github.com/OvertureMaps/tf-transportation/issues/117 

# Testing

Added new examples, tested.

# Checklist

1. [x] Add relevant examples.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/176)


brad-richardson (dismissed):

counterexamples/transportation/segment/road/bad-road-flags-duplicate-flag.json
```diff
@@ -12,13 +12,12 @@
...
     "type": "segment",
     "version": 19,
-    "update_time": "2024-03-14T08:05:36-08:00",
     "subtype": "road",
     "class": "primary",
     "road": {
       "flags": [
         {
-          "values": ["is_tunnel", "is_tunnel"]
+          "values": ["is_tunnel"]
```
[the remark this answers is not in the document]
ibnt1: yes, except it also had timestamp removed, so it still fails but not with the expected reason - this was accidentally committed, leftover from testing, i just reverted it. thanks for finding this!

brad-richardson (approved):

examples/transportation/segment/road/road.yaml
```diff
@@ -59,3 +59,16 @@ properties:
...
+        type: place
+      - value: I 90
+        type: route
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
+      to_connector_id: overture:transportation:connector:123
+      final_heading: backward
```
bastiaanvanassche-tomtom, line 74: minor remark: missing EOL character at end of file

counterexamples/transportation/segment/road/destinations/bad-road-destinations-missing-final_heading.yaml
```diff
@@ -0,0 +1,28 @@
...
+      symbols:
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
+      to_connector_id: overture:transportation:connector:123
+  ext_expected_errors:
+    - "[I#/properties/destinations/0] [S#/$defs/propertyDefinitions/destinations/items/required] missing properties: 'final_heading'"
```
bastiaanvanassche-tomtom, line 28: minor remark: missing EOL character at end of file

counterexamples/transportation/segment/road/destinations/bad-road-destinations-empty-label-value.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
+      to_connector_id: overture:transportation:connector:123
+      final_heading: backward
+  ext_expected_errors:
+    - "[I#/properties/destinations/0/labels/0/value] [S#/$defs/propertyDefinitions/destinations/items/properties/labels/items/properties/value/type] expected string, but got null"
```
bastiaanvanassche-tomtom, line 29: minor remark: missing EOL character at end of file

counterexamples/transportation/segment/road/destinations/bad-road-destinations-bad-label-type.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
+      to_connector_id: overture:transportation:connector:123
+      final_heading: backward
+  ext_expected_errors:
+    - "[I#/properties/destinations/0/labels/0/type] [S#/$defs/propertyDefinitions/destinationLabelType/enum] value must be one of"
```
bastiaanvanassche-tomtom, line 29: minor remark: missing EOL character at end of file

bastiaanvanassche-tomtom (approved):

brad-richardson (approved):

vcschapp (commented):
I'm basically leaving this alone right now until the nesting discussion gets resolved in a principled way.... Ideally we can make a final decision on nesting and all related churn in transportation will stop.

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
         default: []
 "$defs":
   propertyDefinitions:
+    destinationLabelType:
+      description: >-
+        The type of object of the destination label.
+      type: string
+      enum: 
+      - place
```
vcschapp, line 65: This is an unfortunate intersection with the places theme.

What OSM tags/combinations does `place` come from, and can we think of something else to name this?

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+              pattern: ^(\S.*)?\S$
+          from:
+            description: >-
+              List of segments-connector pairs from where we need to transition 
+              to this segment to reach the destination.
+            type: array
+            items:
+              description: Pair of segment and connector IDs along the sequence
+              "$ref": "#/$defs/propertyDefinitions/subSegmentReference"
+              when:
```
vcschapp, line 112: This seems weird? A `when:` just hanging out?

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+        onto this segment to reach the destinations.
+      type: array
+      items:
+        type: object
+        unevaluatedProperties: false
+        required: [labels, from, to_connector_id, final_heading]
+        properties:
+          labels:
+            description: >-
+              Describes the destination labels on the directions sign.
```
vcschapp, line 82: nit: What if you simplified this to "Labeled destinations that can be reached by following the segment"?

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+                type: { "$ref": "#/$defs/propertyDefinitions/destinationLabelType" }
+            minItems: 1
+            uniqueItems: true
+          symbols:
+            type: array
+            items:
+              description: >-
+                A symbol or icon present on the sign.
+              type: string
+              pattern: ^(\S.*)?\S$
```
vcschapp, line 103: My first thought was, how can a string be a symbol?

My second thought was, shouldn't this be enumerated?

If I'm looking at the right TagInfo, there don't seem to be a an enormous number of alternatives: https://taginfo.openstreetmap.org/keys/destination%3Asymbol#values

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+      enum: 
+      - place
+      - street
+      - country 
+      - route
+    destinations:
+      description: >-
+        Information about what destinations can be reached if following this segment. 
+        Includes references to the segments from which it is needed to transition 
+        onto this segment to reach the destinations.
```
vcschapp, line 73: Something that's a bit "off" to me is that, depending on what part of the schema and documentation you read, the schema doesn't seem to be sure about whether `destinations` models *destinations* themselves, or whether it models the *signs* that point to destinations.

On reflection I think it's the signs, since the purpose of `destinations` is to model the signs a driver or traveller actually sees.

Can we have a go at making the documentation consistent?

I would be tempted to start with a description along the following lines:

```
Describes objects that can be reached by following a transportation segment in the same way those
objects are described on signposts or ground writing that a traveller following the segment in would
observe in the real world. This allows navigation systems to refer to signs and observable writing
that a traveller actually sees.
```

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+            type: array
+            items:
+              description: >-
+                A symbol or icon present on the sign.
+              type: string
+              pattern: ^(\S.*)?\S$
+          from:
+            description: >-
+              List of segments-connector pairs from where we need to transition 
+              to this segment to reach the destination.
```
vcschapp, line 107: Suggested tweak that feels a bit more approachable to me:

```
List of segment/connector pairs that must be followed, in order, immediately before
entering this segment, in order for these destinations to be reachable.
```

examples/transportation/segment/road/destinations/road-destinations.yaml
```diff
@@ -0,0 +1,34 @@
...
+      symbols:
+      - motorway
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:234
+        connector_id: overture:transportation:connector:234
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
```
vcschapp, line 32: Interesting. Isn't travel direction already implied by the sequence?

I think `when` is only allowed here because we mistakenly forgot to set `unevaluatedProperties: false` in the sequence item model.

schema/transportation/segment.yaml
```diff
@@ -455,23 +526,23 @@ properties:
...
       type: object
       required: [connector_id, segment_id]
       properties:
         connector_id:
           description: >-
-            Identifies the point of physical connection between the
-            previous segment in the sequence and the segment in this
-            sequence entry.
+            Identifies the point of physical connection between this segment 
+            and other segments.
```
vcschapp, line 541: I don't understand this description change. It converts from something well-defined/rigorous into something vague. Why would we want to do that?

---

If I'm understanding right, I think this is motivated by a difference between prohibited transitions and destinations, where in PT the sequence comes *after* the current segment and in D it comes *before* the current segment.

In the circumstances, it's probably better to duplicate the sequence concept into two types, `followingSequence` for PT and ``precedingSequence` for D, so that we can still have rigorous definitions for both.

counterexamples/transportation/segment/road/destinations/bad-road-destinations-missing-final_heading.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - labels:
+      - value: Seattle
+        type: place
+      symbols:
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
```
vcschapp, line 25: This should be removed.

counterexamples/transportation/segment/road/destinations/bad-road-destinations-bad-label-type.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - labels:
+      - value: Seattle
+        type: bad_type
+      symbols:
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
```
vcschapp, line 25: Should be removed.

counterexamples/transportation/segment/road/destinations/bad-road-destinations-empty-label-value.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - labels:
+      - value:
+        type: place
+      symbols:
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
```
vcschapp, line 25: Should be removed.

counterexamples/transportation/segment/road/destinations/bad-road-destinations-duplicate.yaml
```diff
@@ -0,0 +1,31 @@
...
+        type: place
+      - value: Seattle
+        type: place
+      symbols:
+      - airport
+      from:
+      - segment_id: overture:transportation:segment:567
+        connector_id: overture:transportation:connector:567
+        when:
+          heading: forward
```
vcschapp, line 27: Should be removed.

schema/transportation/segment.yaml
```diff
@@ -51,6 +57,71 @@ properties:
...
+              when:
+              allOf:
+                - "$ref": "#/$defs/propertyContainers/headingScopeContainer"
+          to_connector_id:
+            description: >-
+              Identifies the point of physical connection between the
+              previous segment in the sequence and the segment in this
+              sequence entry.
+            type: string                  
+          final_heading:
```
vcschapp, line 121: `final_heading` seems weird here. It makes more sense on prohibited transitions because they are referring to a future sequence you need to take. But here, you're literally on the segment.

We should either rename it to `heading` or replace it with a mandatory `when: heading:` block. I think the second one is more consistent with the transportation model as a whole.

vcschapp (changes requested):

examples/transportation/segment/road/destinations/road-destinations.yaml
vcschapp: New comment, during convo. w. Cozmin, not entirely related but with clearer understanding now.

Current segment, which is the "to" in the from-to model, would have:

```
when: { heading: backward }
```

On the from segments in the `from` list, would have just an `initial_heading` property.

That's the way that would make it most consistent w prohibited transitions without flipping the from-to direction.

----

If we flipped it so that destinations live on the from, then it would be:

```
when: { heading: forward }
```

And the `from` list would become `to` and that would have `final_heading: backward`.

schema/transportation/segment.yaml
```diff
@@ -455,23 +526,23 @@ properties:
           properties: { is_at_most: { allOf: [ { "$ref": "#/$defs/propertyDefinitions/weightValueWithUnit" } ] } }
         - required: [is_less_than]
           properties: { is_less_than: { allOf: [ { "$ref": "#/$defs/propertyDefinitions/weightValueWithUnit" } ] } }
-    sequenceEntry:
+    subSegmentReference:
```
vcschapp, line 529: One idea for name is `connectorOnSegment`.

schema/transportation/segment.yaml
ibnt1: renamed to `unknown`

schema/transportation/segment.yaml
ibnt1: replaced with enum

schema/transportation/segment.yaml
ibnt1: reverted, no longer needed after refactored the property to live on the 'from'

schema/transportation/segment.yaml
ibnt1: done

schema/transportation/segment.yaml
ibnt1: fixed

schema/transportation/segment.yaml
ibnt1: reverted this change

schema/transportation/segment.yaml
ibnt1: not applicable after refactoring

examples/transportation/segment/road/destinations/road-destinations.yaml
ibnt1: as discussed, when.heading and final_heading kept after moving this property on the source/from segment

schema/transportation/segment.yaml
```diff
@@ -137,6 +210,193 @@ properties:
...
+      - motorroad
+      - fuel
+      - viewpoint
+      - fuel_diesel
+      - food
+      - lodging
+      - info
+      - camp_site
+      - interchange
+      - toilets          # what's the appropriate US english value for this, restrooms?
```
brad-richardson, line 237: `bathrooms` maybe?

schema/transportation/segment.yaml
```diff
@@ -137,6 +210,193 @@ properties:
...
+    road:
```
brad-richardson, line 238: You want us to go back to JSON land??

schema/transportation/segment.yaml
```diff
@@ -62,6 +63,78 @@ properties:
...
+      - street
+      - country 
+      - route_ref
+      - toward_route_ref
+      - unknown
+    destinations:
+      description: >-
+        Describes objects that can be reached by following a transportation segment in the
+        same way those objects are described on signposts or ground writing that a traveller
+        following the segment in would observe in the real world. This allows navigation systems
```
brad-richardson, line 80:
```suggestion
        following the segment would observe in the real world. This allows navigation systems
```

schema/transportation/segment.yaml
```diff
@@ -62,6 +63,78 @@ properties:
...
+          from_connector_id:          
+            description: >-
+              Identifies the point of physical connection on this segment before which
+              the destination sign or marking is visible.
+            type: string
+          to_segment_id:
+            description: >-
+              Identifies the segment to transition to reach the destination(s) labeled 
+              on the sign or marking.
+            type: string
```
brad-richardson, line 122: Do we have an ID pattern to validate these?

brad-richardson (commented):
These are complex enough that we should probably add new examples to the road attributes page and maybe the schema examples page with diagrams. https://docs.overturemaps.org/schema/attributes/transportation/roads/

schema/transportation/segment.yaml
ibnt1: no, we don't...schema has multiple fields that represent id references, all of them currently only have `type: string` constraint

schema/transportation/segment.yaml
```diff
@@ -137,6 +210,193 @@ properties:
       enum: [ vehicle, motor_vehicle, car, truck, motorcycle, foot, bicycle, bus, hgv, hov, emergency ]
       "$comment": >-
         motor_vehicle includes car, truck and motorcycle
+    destinationSignSymbol:
+      description: >-
+        Indicates what special symbol/icon is present on a signpost, visible as road marking or similar.
+      type: string
+      enum: 
+      - motorway
```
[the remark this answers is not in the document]
ibnt1: as discussed, keeping for consistency with class value

schema/transportation/segment.yaml
```diff
@@ -137,6 +210,193 @@ properties:
...
+      - airport
+      - hospital
+      - center           # centre in raw OSM value
```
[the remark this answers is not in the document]
ibnt1: yes, added comment

schema/transportation/segment.yaml
ibnt1: ops, mistake at merge conflict resolution, reverted

schema/transportation/segment.yaml
```diff
@@ -62,6 +63,78 @@ properties:
       routes: { "$ref": "#/$defs/propertyDefinitions/routes" }
 "$defs":
   propertyDefinitions:
+    destinationLabelType:
+      description: >-
+        The type of object of the destination label.
+      type: string
+      enum: 
+      - street
```
bastiaanvanassche-tomtom, line 71: should we have `place` as well? it's at least part of your example: https://github.com/OvertureMaps/schema/pull/176/files#diff-83cf2a8a673bc7660b22d82ec08a79d2c5412a5f700b4315ca51ba9828fb2d44

ibnt1: fixed

brad-richardson (approved):

PascalClarysse-TomTom (dismissed):

mojodna (approved):

brad-richardson (dismissed):

PascalClarysse-TomTom (approved):

1
jwass, line 1: is this file a mistake?

jenningsanderson (approved):

PR #188, opened 2024-05-14
Add MultiLineString as possible geometry of boundary

sasastanojkov:
Currently we have linestring per boundary entity. This implies that if boundary between two divisions is split, by another division of the same subtype, we will have multiple entities representing boundary between two divisions. At this point GERS id is created based on divisions which boundary is separating and boundary geometry. With each change to the divisions the boundary GERS id will change, and it will be unstable.
For example, border between Spain (085e7a023fffffff012656f58e364b88) and France (0854db377fffffff01c2b9178bdd68b1) is split in 3 parts (085e7a023fffffff012ad418147bdcf8, 085e7a023fffffff010a4e09caa458ce, 085e7a023fffffff012656f58e364b88). With change of Spain or France polygon any of these boundaries can get changed GERS due to change in their geometry. If we were to create single entity to represent boundary between these two divisions, we wouldn't need geometry for generating GERS and it would be far more stable and easier to maintain.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/188)

### DIFF
--- a/schema/divisions/boundary.yaml
+++ b/schema/divisions/boundary.yaml
@@ -13,7 +13,7 @@ properties:     # JSON Schema: Top-level object properties.
       GeoJSON schema.
     unevaluatedProperties: false
     oneOf:
-      - "$ref": https://geojson.org/schema/LineString.json,
+      - "$ref": https://geojson.org/schema/LineString.json
       - "$ref": https://geojson.org/schema/MultiLineString.json
   properties:   # GeoJSON: top-level object 'properties' property.
     unevaluatedProperties: false
--- a/schema/divisions/boundary.yaml
+++ b/schema/divisions/boundary.yaml
@@ -13,7 +13,8 @@ properties:     # JSON Schema: Top-level object properties.
       GeoJSON schema.
     unevaluatedProperties: false
     oneOf:
-      - "$ref": https://geojson.org/schema/LineString.json
+      - "$ref": https://geojson.org/schema/LineString.json,
+      - "$ref": https://geojson.org/schema/MultiLineString.json
   properties:   # GeoJSON: top-level object 'properties' property.
     unevaluatedProperties: false
     required: [subtype, class, divisions]


jwass:
This makes sense to me. I guess I'm curious about @vcschapp's opinion. Was the intent of the divisions boundary to capture each segment at a boundary ?? or the entire shared border between the entities?

I agree with the post that if it's a segment, maintaining stability will be basically impossible but I wonder if we care... are people really attaching data to these boundary segments?

vcschapp:
> This makes sense to me. I guess I'm curious about @vcschapp's opinion. Was the intent of the divisions boundary to capture each segment at a boundary ?? or the entire shared border between the entities?
> 
> I agree with the post that if it's a segment, maintaining stability will be basically impossible but I wonder if we care... are people really attaching data to these boundary segments?

It also makes sense to me.

@jwass A use case we talked about for stability was trying to attach custom cartographic judgments to specific boundaries, for example whether a specific disputed boundary should be hidden or not.

@sasastanojkov

- Due to disputed portions, it will still be the case that multiple boundary features between divisions may be needed in some cases, right? Does that fit into your mental model?
- If this change gets merged in, can you work with @danabauer to ensure she understands it and can document appropriately? We should give our customers/end users as clear and definite an explanation for how boundaries work as we can.

jwass (approved):

vcschapp (approved):

vcschapp (approved):

TristanDiet-TomTom (approved):

sasastanojkov:
> > This makes sense to me. I guess I'm curious about @vcschapp's opinion. Was the intent of the divisions boundary to capture each segment at a boundary ?? or the entire shared border between the entities?
> > I agree with the post that if it's a segment, maintaining stability will be basically impossible but I wonder if we care... are people really attaching data to these boundary segments?
> 
> It also makes sense to me.
> 
> @jwass A use case we talked about for stability was trying to attach custom cartographic judgments to specific boundaries, for example whether a specific disputed boundary should be hidden or not.
> 
> @sasastanojkov
> 
> * Due to disputed portions, it will still be the case that multiple boundary features between divisions may be needed in some cases, right? Does that fit into your mental model?
> * If this change gets merged in, can you work with @danabauer to ensure she understands it and can document appropriately? We should give our customers/end users as clear and definite an explanation for how boundaries work as we can.

Yea that fits. I would say something along the lines GERS is created based on divisions divided by given boundary and its perspectives. In that sense we would have one entity "everyone agrees on" plus any number of entities grouped by their respective perspectives (hope this makes sense).

If we take, for instance border between France and Spain it is single entity nobody disputes having 3 line strings.
If we take boundary between Chile and Argentina we have one boundary which isn't disputed having two line strings and two entities for Argentinas and Chileans perspective of the disputed boundary. That way we have 3 GERS-ified entities with stable ids (agreed upon border, rest of the border from AR perspective, rest of the border from CL). In case they start disputing over some other part of the border we take out that part from mutually agreed upon border and put implementations by each side in their respective entities as another linestring.

After changes in [PR](https://github.com/OvertureMaps/schema/pull/162) are completed, I will open a new one adding perspectives to boundaries in the same manner it is done for divisions, and users will be able to use the same mechanism to include/exclude boundary entities as they do for divisions.

PR #198, opened 2024-05-22
Adding all of the OSM variants for names

jenningsanderson:
# Description

Allows more `variants` of names to come through during name conversion from name tags in OSM. 

# Reference

1. https://github.com/OvertureMaps/tf-base/issues/62
2. https://github.com/OvertureMaps/tf-data-platform/pull/288

# Testing

Ran successfully against `base` features

# Checklist

1. [N/A] Add relevant examples.
3. [N/A] Add relevant counterexamples.
4. [N/A] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
5. [N/A] Update in-schema documentation using plain English written in complete sentences, if an update is required.
6. [N/A] Update Docusaurus documentation, if an update is required.
7. [N/A] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/198)

### DIFF
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -194,9 +194,16 @@ description: Common schema definitions shared by all themes
         variant:
           type: string
           enum:
-            - common
-            - official
             - alternate
+            - common
+            - historical
+            - international
+            - local
+            - national
+            - nickname
+            - official
+            - reference
+            - regional
             - short
         language: { "$ref": "#/$defs/propertyDefinitions/language" }
         value:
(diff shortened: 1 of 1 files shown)


vcschapp (commented):
How important is this change?

One thing that's nice about the simple common/official/alternate/short schema is that it is simple.

---

If we merge this change, is there follow-up work to ensure that all the themes consistently provide all these names where available?

vcschapp (commented):
Decision in [2024-06-05](https://github.com/OvertureMaps/schema-wg/issues/309) schema TF meeting: map all these extra tags to "alternate".

This is a two-way door, since we can add new name categories later.

So the schema will stay the same, but the name mapping rules will need to change.

We do need to capture the tag mapping logic for names. @jwass will try to capture these rules.

jenningsanderson:
Closing for now — in favor of remapping everything to _alternate_

PR #203, opened 2024-05-29
Segment restructure

brad-richardson:
# Description

Storing road data as JSON carries a number of issues and was always considered a "stop gap" during rapid feature iteration for transportation. As we approach GA, we need to properly schematize this data. It is proposed that this also includes some structure simplification ("flattening") for ease of querying, whereas before a priority was making the JSON nicely human-readable. I've left in `road` as-is to allow us to ship both deliveries for one release to ease migration.

Here is the high level schema change:
Old JSON path | New column
-- | --
road.restrictions.speed_limits | speed_limits
road.restrictions.access | access_restrictions
road.restrictions.prohibited_transitions | prohibited_transitions
road.surface | road_surface
road.flags | road_flags
road.width | width_rules
road.lanes | lanes
road.level | level_rules

Additionally, for the non-geometry vehicle scoping block (`when: vehicle:`), we will simplify that frequently used structure as follows (note that this is a **breaking change** for consumers of this data):
```
speed_limits:
  when:
    ...
    vehicle:
      - dimension: <axle_count, height, length, weight, width>
        comparison: <greater_than, greater_than_equal, equal, less_than, less_than_equal>
        value: str
        unit: str
```
vs previously:
```
speed_limits:
  when:
    ...
    vehicle:
      axle_count:
        is_more_than:
          value:
          unit:
        is_less_than:
          ...
        ...
        ...
      height:
        ...
      length:
        ...
      weight:
        ...
      width:
        ...
```

See additional context in [this discussion](https://github.com/OvertureMaps/schema/discussions/184)

# Reference

1. https://github.com/OvertureMaps/schema/discussions/184
2. https://github.com/OvertureMaps/tf-transportation/issues/229

# Testing 

Updated all examples and counterexamples to reflect changes, including copying subproperties out of `road` to separate properties.

# Checklist

1. [X] Add relevant examples.
3. [X] Add relevant counterexamples.
4. [X] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
5. [X] Update in-schema documentation using plain English written in complete sentences, if an update is required.
6. [X] Update Docusaurus documentation, if an update is required.
7. [ ] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/203)

### DIFF
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -25,7 +25,6 @@ properties:
           required: [class]
           properties:
             class: { "$ref": "#/$defs/propertyDefinitions/roadClass" }
-            road: { "$ref": "#/$defs/propertyDefinitions/road" }
             speed_limits: { "$ref": "#/$defs/propertyContainers/speedLimitsContainer" }
             access_restrictions: { "$ref": "#/$defs/propertyContainers/accessContainer" }
             prohibited_transitions: { "$ref": "#/$defs/propertyContainers/prohibitedTransitionsContainer" }
@@ -35,6 +34,8 @@ properties:
             lanes: { "$ref": "#/$defs/propertyContainers/lanesContainer" }
             level_rules: { "$ref": "#/$defs/propertyContainers/levelRulesContainer" }
             level: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/level" }
+            # The `road` property is deprecated and will be removed in the 2024-07 release
+            road: { "$ref": "#/$defs/propertyDefinitions/road" }
       - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
       - "$ref": ../defs.yaml#/$defs/propertyContainers/namesContainer
     properties:
@@ -137,6 +138,8 @@ properties:
         motor_vehicle includes car, truck and motorcycle
     road:
       description: >-
+        ** Note: this field is deprecated in favor of promoting its sub-properties into top-level properties and will be removed in the 2024-07 release **
+
         Properties for segments whose segment subtype is road. The road
         subtype includes any variety of road, street, or path, including
         dedicated paths for walking and cycling.
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -241,7 +241,7 @@ properties:
         - required: [is_less_than]
           properties: { is_less_than: { type: integer } }
     vehicleScopeDimension:
-      description: TODO
+      description: Enumerates possible vehicle dimensions for use in restrictions
       type: string
       enum:
         - axle_count
@@ -250,7 +250,7 @@ properties:
         - weight
         - width
     vehicleScopeComparison:
-      description: TODO
+      description: Enumerates possible comparison operators for use in scoping
       type: string
       enum:
         - greater_than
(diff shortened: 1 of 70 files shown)


schema/transportation/segment.yaml
```diff
@@ -26,6 +26,15 @@ properties:
           properties:
             class: { "$ref": "#/$defs/propertyDefinitions/roadClass" }
             road: { "$ref": "#/$defs/propertyDefinitions/road" }
```
ibnt1, line 28: i would include a comment in `road` property definition's description that it is a deprecated property in favor of top-level properties corresponding to its sub-properties

ibnt1 (approved):

schema/transportation/segment.yaml
```diff
@@ -19,12 +19,24 @@ properties:
     unevaluatedProperties: false
     required: [subtype]
     allOf:
+      - title: "Segment Properties"
+        required: [class]
+        properties:
+          class: { "$ref": "#/$defs/propertyDefinitions/roadClass" }
+          access_restrictions: { "$ref": "#/$defs/propertyContainers/accessContainer" }
```
vcschapp, line 26: I have a very soft preference for using `access` instead of `access_restrictions` for no good reason other than that it's shorter. But I can live with both.

schema/transportation/segment.yaml
```diff
@@ -128,166 +140,28 @@ properties:
         motor_vehicle includes car, truck and motorcycle
     road:
       description: >-
+        ** Note: this field is deprecated in favor of promoting its sub-properties into top-level properties and will be removed in the 2024-07 release **
+
         Properties for segments whose segment subtype is road. The road
         subtype includes any variety of road, street, or path, including
         dedicated paths for walking and cycling.
       type: object
       unevaluatedProperties: false
```
vcschapp, line 149: Since the actual properties are deleted, does it make sense to explicitly set `unevaluatedProperties: true` until the `road` property is deleted? Or alternatively, just to not delete the sub-properties?

vcschapp (approved):

schema/transportation/segment.yaml
brad-richardson: I didn't delete the sub-properties, just shuffled them into containers for reuse. Did I miss something?

schema/transportation/segment.yaml
brad-richardson: No strong preference for me either, I'll take a poll on this in the tf-transportation meeting tomorrow

bastiaanvanassche-tomtom (approved):

schema/transportation/segment.yaml
vcschapp: Nope, it seems like I missed something!

vcschapp: Yeah I definitely read the red and not the green. That's a pretty rookie mistake. Sorry for the churn, but I'll resolve this one.

schema/transportation/segment.yaml
vcschapp: Resolving this for now because it's not pressing and we can change it later. I think there's more value in merging the change first.

BruceHarold:
Any problem using float to model speed limit vehicle dimension values?

PR #208, opened 2024-06-03
Building is underground

jwass:
# Description

Adds an `is_underground` flag to allow buildings or parts to be tagged as completely underground. While Overture has the level column, it is currently insufficient for entirely determining whether a feature is below ground. As it's mostly taken from OSM's "layer" tag, that value can be negative for above ground features.

Note - Transportation has the notion of segments being below ground, but that is captured as tunnel and covered flags which do not really apply well to buildings.

# Reference

1. TODO - Add documentation once we know there's alignment

1. [x ] Add relevant examples.

# Documentation Website

### DIFF
--- a/schema/buildings/defs.yaml
+++ b/schema/buildings/defs.yaml
@@ -13,8 +13,9 @@ shapeContainer:
       exclusiveMinimum: 0
     is_underground:
       description: >-
-        Whether the entire building or part is below ground-level. This is different than the level column which is used
-        to indicate z-ordering of elements and negative values may be above ground.
+        Whether the entire building or part is completely below ground. This is useful for rendering which typically omits these
+        buildings or styles them differently because they are not visible above ground. This is different than the level column
+        which is used to indicate z-ordering of elements and negative values may be above ground.
       type: boolean
     num_floors:
       description: >-
--- a/schema/buildings/defs.yaml
+++ b/schema/buildings/defs.yaml
@@ -13,7 +13,7 @@ shapeContainer:
       exclusiveMinimum: 0
     is_underground:
       description: >-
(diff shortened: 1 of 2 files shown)


schema/buildings/defs.yaml
```diff
@@ -11,6 +11,11 @@ shapeContainer:
         Height of the building or part in meters. The height is the distance from the lowest point to the highest point.
       type: number
       exclusiveMinimum: 0
+    is_underground:
+      description: >-
+        Whether the entire building or part below ground-level. This is different than the level column which is used
```
jenningsanderson, line 16: part *is* below ground

jenningsanderson (dismissed):

skmoore:
What about a case where a building from OSM has no parts, but has floors below and above ground? This wouldn't cover that case, correct?

jwass:
> What about a case where a building from OSM has no parts, but has floors below and above ground? This wouldn't cover that case, correct?

@skmoore Correct. There are other ways to capture that but is not part of this. This is for when the entire thing is underground - i.e., has different render characteristics because it won't be seen from above

schema/buildings/defs.yaml
jwass: Fixed. Thanks!

jwass:
@skmoore I added a bit more language requiring that the building/part be *completely* below ground.

vcschapp (approved):
I like the short punchy name :)

TristanDiet-TomTom (approved):
Simple and clear attribute.

LukaszMichalski-TomTom (approved):

DavidKarlas (approved):

PR #215, opened 2024-06-12
Fix new issues with boundary relationship modeling

vcschapp:
# Description

## Primary change

The primary change in this PR is to align the foreign key references to division features within boundary features with the schema relationships guidance.

In a recent change, PR #200, the schema for the divisions theme's boundary feature fell out of alignment with the *Relationships in the Schema* guidance.
    
The guidance is that a property or nested sub-property that contains a foreign key reference to another feature by Overture ID must end in the suffix `_id`.
    
Before PR #200, things were fine because the Overture ID was contained in a nested sub-property, `divisions[*].division_id`. But since PR #200 removed the nested structures from the `divisions` property, it should have renamed it `division_ids`. This change fixes that.

## Secondary change

In another recent change, PR #188, the boundary feature's geometry property description got decoupled from the actual allowed content of the geometry property. This change fixes the description.

# Reference

- Foreign key relationships:
    - https://github.com/OvertureMaps/schema/pull/200
    - https://github.com/OvertureMaps/schema-wg/blob/admin/decisions/adr-relationship-modeling.md
    - https://wiki.overturemaps.org/x/oAEjAQ
- Description change: https://github.com/OvertureMaps/schema/pull/188

# Testing

Updated examples and counterexamples and ran the test script.

# Checklist

1. [X] ~~Add~~ Update relevant examples.
3. [X] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
4. [X] Update in-schema documentation using plain English written in complete sentences, if an update is required.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/215)

### DIFF
--- a/counterexamples/divisions/boundary/bad-division_ids-1.yaml
+++ b/counterexamples/divisions/boundary/bad-division_ids-1.yaml
@@ -7,11 +7,11 @@ geometry:
 properties:
   theme: divisions
   type: boundary
-  update_time: "2024-02-23T22:47:06Z"
-  version: 0
+  update_time: "2024-06-12T17:52:42Z"
+  version: 1
   subtype: region
   class: land
-  divisions: ["example:division:country:left", "example:division:country:right", "example:division:country:right", 123]
+  division_ids: ["example:division:country:left", "example:division:country:right", "example:division:country:right", 123]
   is_disputed: "true"
   ext_expected_errors:
     - maximum 2 items required, but found 4 items
--- a/schema/divisions/boundary.yaml
+++ b/schema/divisions/boundary.yaml
@@ -17,7 +17,7 @@ properties:     # JSON Schema: Top-level object properties.
       - "$ref": https://geojson.org/schema/MultiLineString.json
   properties:   # GeoJSON: top-level object 'properties' property.
     unevaluatedProperties: false
-    required: [subtype, class, divisions]
+    required: [subtype, class, division_ids]
     allOf:
       - "$ref": ../defs.yaml#/$defs/propertyContainers/overtureFeaturePropertiesContainer
     properties: # JSON Schema: properties within GeoJSON top-level object 'properties' property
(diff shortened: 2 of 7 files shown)


schema/divisions/boundary.yaml
```diff
@@ -30,16 +30,16 @@ properties:     # JSON Schema: Top-level object properties.
...
+          Identifies the two divisions to the left and right, respectively, of
+          the boundary line. The left- and right-hand sides of the boundary are
+          considered from the perspective of a person standing on the line
+          facing in the direction in which the geometry is oriented, i.e. facing
+          toward the end of the line.
           
-          The first array element represents the left division. The second
-          element represents the right division.
+          The first array element is the Overture ID of the left division. The
+          second element is the Overture ID of the right division.
```
sasastanojkov, line 42: Overture ID or GERS id? Since technically division GERS ids will go there.

vcschapp: My understanding is we're aiming to rename "GERS ID" to "Overture ID" to reduce cognitive load. Same concept, same stable ID, different name.

@jenningsanderson @danabauer Do I have this right?

jwass (approved):

schema/divisions/boundary.yaml
danabauer: We're in limbo until the Steering Committee weighs in. One possible (and likely) outcome is that we'll refer to the `id` as "Overture ID" and the reference system as GERS.

danabauer: I agree that you should use "Overture ID"

counterexamples/divisions/boundary/bad-division_ids-1.yaml
```diff
@@ -7,11 +7,11 @@ geometry:
 properties:
   theme: divisions
   type: boundary
-  update_time: "2024-02-23T22:47:06Z"
-  version: 0
+  update_time: "2024-06-12T17:52:42Z"
+  version: 1
```
danabauer, line 11: this is great to see the version updated here. Do you know if we're doing this consistently across themes?

danabauer (approved):

TristanDiet-TomTom (approved):

DavidKarlas (approved):

counterexamples/divisions/boundary/bad-division_ids-1.yaml
vcschapp: It's really just me who does this. I can be meticulous to the point of silliness sometimes.

PR #217, opened 2024-06-12
Add num_floors_underground

skmoore:
# Description

Adds a `num_floors_underground` flag for buildings that have one or more floors below ground level. This will allow for more accurate 3d representations of buildings that are partially or fully below ground. This should map to the `building:levels:underground` tag from OSM.

# Reference

1. TODO.

1. [x] Add relevant examples.

# Documentation Website

### DIFF
--- a/counterexamples/buildings/bad-floors-underground.yaml
+++ b/counterexamples/buildings/bad-floors-underground.yaml
@@ -12,4 +12,4 @@ properties:
   num_floors_underground: -1
   # Expected errors:
   ext_expected_errors:
-    - "[I#/properties/num_floors_underground] exclusiveMinimum: got -1, want 0"
+    - "exclusiveMinimum: got -1, want 0"
--- a/counterexamples/buildings/bad-floors-underground.yaml
+++ b/counterexamples/buildings/bad-floors-underground.yaml
@@ -1,0 +1,13 @@
+---
+id: overture:buildings:footprint:1234
+type: Feature
+geometry:
+  type: Polygon
+  coordinates: [[[0, 0], [0, 1], [1, 1], [1, 0], [0, 0]]]
+properties:
+  theme: buildings
+  type: building
+  version: 1
+  num_floors_underground: -1
+  ext_expected_errors:
+    - "[I#/properties/num_floors_underground] [S#/$defs/propertyDefinitions/featureUpdateTime/format] 'abc2023-02-22T23:55:01-08:00' is not valid 'date-time'"
--- a/counterexamples/buildings/bad-floors-underground.yaml
+++ b/counterexamples/buildings/bad-floors-underground.yaml
@@ -7,5 +7,9 @@ geometry:
 properties:
   theme: buildings
   type: building
+  update_time: "2023-02-22T23:55:01-08:00"
   version: 1
   num_floors_underground: -1
+  # Expected errors:
+  ext_expected_errors:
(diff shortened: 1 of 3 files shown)


jwass (dismissed):

vcschapp (approved):

vcschapp (changes requested):
@skmoore Would you mind adding one counterexample that has an invalid value for `num_floors_underground` and uses the `ext_expected_exception` property to verify that the right validation exception is generated?

jenningsanderson (dismissed):

jwass (approved):
Thanks!

TristanDiet-TomTom (approved):

DavidKarlas (approved):

jenningsanderson:
> @skmoore Would you mind adding one counterexample that has an invalid value for `num_floors_underground` and uses the `ext_expected_exception` property to verify that the right validation exception is generated?

I added a counter example, but I cannot generate the proper expected exception?

counterexamples/buildings/bad-floors-underground.yaml
```diff
@@ -0,0 +1,11 @@
+---
+id: overture:buildings:footprint:1234
+type: Feature
+geometry:
+  type: Polygon
+  coordinates: [[[0, 0], [0, 1], [1, 1], [1, 0], [0, 0]]]
+properties:
+  theme: buildings
+  type: building
```
vcschapp, line 9: - Missing `update_time` which means it's schema-invalid even without the `num_floors_underground`.
- Would you mind adding the `ext_expected_errors: []` property?
    - That property acts like an assertion to verify that the schema validator fails with the error you are expecting.
    - It's insurance against counterexamples that are invalid, but for the wrong reason. There should be lots of examples of use in other nearby counterexample files.

vcschapp (changes requested):

counterexamples/buildings/bad-floors-underground.yaml
vcschapp: Still seems to be missing `update_time` and `ext_expected_errors`...

vcschapp (changes requested):

vcschapp:
> > @skmoore Would you mind adding one counterexample that has an invalid value for `num_floors_underground` and uses the `ext_expected_exception` property to verify that the right validation exception is generated?
> 
> I added a counter example, but I cannot generate the proper expected exception?

Can you help me understand what actions you did, and how the actual result different from your expectation?

The way I usually do this is:

1. Add `ext_expected_exceptions: ["some nonsense"]`
2. Run `./test.sh -m counterexamples "path/to/your/counterexample.yaml`.
    - The test run should fail and print detailed output.
    - The reason it fails is because none of the failure outputs match the text "some nonsense".
4. Look at the output.
    - Make sure there aren't any *unexpected* failures... If there are, the counterexample has another issue that needs fixing.
    - Find the *expected* failure line in the output.
5. Replace "some nonsense" with a useful substring from the expected failure line.
6. Repeat step #2, which should pass now.

counterexamples/buildings/bad-floors-underground.yaml
jenningsanderson: I added `update_time` — However, where can I find what the expected errors should look like? Because mine look different...

vcschapp (approved):

DavidKarlas (approved):

jenningsanderson (approved):

PR #225, opened 2024-06-27
Adding capital_of_divisions to Divisions

DavidKarlas:
# Description

This will allow easier rendering of Division data since reverse lookup via capital_division_ids is not needed anymore.

# Reference

1. This was discussed briefly in https://github.com/OvertureMaps/tf-admin/issues/98

# Testing

# Checklist

1. [x] Add relevant examples.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/225)

### DIFF
--- a/schema/divisions/defs.yaml
+++ b/schema/divisions/defs.yaml
@@ -105,3 +105,14 @@ description: Common schema definitions for divisions theme
           items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/iso3166_1Alpha2CountryCode" }
           minItems: 1
           uniqueItems: true
+    capitalOfDivisionItem:
+      description: One division that has capital
+      type: object
+      unevaluatedProperties: false
+      required: [division_id, subtype]
+      properties:
+        division_id:
+          description: ID of the division
+          allOf:
+            - "$ref": "../defs.yaml#/$defs/propertyDefinitions/id"
+        subtype: { "$ref": "#/$defs/typeDefinitions/placetype" }
--- a/schema/divisions/division.yaml
+++ b/schema/divisions/division.yaml
@@ -197,8 +197,8 @@ properties:     # JSON Schema: Top-level object properties.
         items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/id" }
       capital_of_divisions:
         description:
-          Division IDs and subtypes of division's this division is capital of. If present,
-          this property will refer to the division IDs of the countries, regions, counties, etc.
+          Division IDs and subtypes of divisions this division is a
+          capital of.
         type: array
         minItems: 1
         uniqueItems: true
--- a/schema/divisions/division.yaml
+++ b/schema/divisions/division.yaml
@@ -195,4 +195,12 @@ properties:     # JSON Schema: Top-level object properties.
         minItems: 1
         uniqueItems: true
         items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/id" }
+      capital_of_divisions:
+        description:
+          Division IDs and subtypes of division's this division is capital of. If present,
+          this property will refer to the division IDs of the countries, regions, counties, etc.
+        type: array
+        minItems: 1
+        uniqueItems: true
+        items: { "$ref": "defs.yaml#/$defs/typeDefinitions/capitalOfDivisionItem" }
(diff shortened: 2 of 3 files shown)


danabauer:
tagging @jonahadkins for visibility

schema/divisions/division.yaml
```diff
@@ -195,4 +195,12 @@ properties:     # JSON Schema: Top-level object properties.
         minItems: 1
         uniqueItems: true
         items: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/id" }
+      capital_of_divisions:
+        description:
+          Division IDs and subtypes of division's this division is capital of. If present,
+          this property will refer to the division IDs of the countries, regions, counties, etc.
```
jwass, line 201: Maybe make a note here the subtype is a direct copy of the division_id's subtype.

jonahadkins:
cant wait to CAPITALize on this

jonahadkins (approved):

jwass:
Just to spitball a little bit, what about `capital_of` which has `division_id` instead of `capital_of_divisions` which then also has `division_id`? 

Trying to think of other places in the schema where we try to capture this reverse relationship but I don't think there are any

@vcschapp / @DavidKarlas

schema/divisions/defs.yaml
```diff
@@ -105,3 +105,14 @@ description: Common schema definitions for divisions theme
...
+      description: One division that has capital
+      type: object
+      unevaluatedProperties: false
+      required: [division_id, subtype]
+      properties:
+        division_id:
+          description: ID of the division
+          allOf:
+            - "$ref": "../defs.yaml#/$defs/propertyDefinitions/id"
+        subtype: { "$ref": "#/$defs/typeDefinitions/placetype" }
```
jwass, line 118: Maybe it should be `division_id` and `division_subtype`?

schema/divisions/division.yaml
DavidKarlas: Could you help me phrase that to make it as clear as possible?

jwass: "This field indicates when this division is the capital of one or more other divisions. Each item contains the division ID and subtype of the division(s) for which it is the capital".

It is indeed difficult to not end that sentence with "is the capital of".

DavidKarlas: Updated text.

DavidKarlas:
> Just to spitball a little bit, what about `capital_of` which has `division_id` instead of `capital_of_divisions` which then also has `division_id`?
> 
> Trying to think of other places in the schema where we try to capture this reverse relationship but I don't think there are any
> 
> @vcschapp / @DavidKarlas

Only concern I have is that we currently have `capital_division_ids`, which makes me wonder if we should omit `division` part there as well, I also understand here we have object that inside has `division_id`, so it not exactly same case, but might look weird to have `capital_of` and `capital_division_ids` as top level properties setting next to each other.

jwass:
> Only concern I have is that we currently have capital_division_ids, which makes me wonder if we should omit division part there as well, I also understand here we have object that inside has division_id, so it not exactly same case, but might look weird to have capital_of and capital_division_ids as top level properties setting next to each other.

Good point. Okay - yeah let's just keep capital_of_divisions as-is.

jwass (approved):

TristanDiet-TomTom (dismissed):

schema/divisions/division.yaml
vcschapp, line 201: 1. [  ] Minor: Can you drop the apostrophe in *division's* on the first line? The apostrophe indicates possession, whereas you want plural.
2. [ ] Minor: Can we insert an *a* between "capital" and "of" on the first line?
3. [ ] Minor: I would suggest dropping the second sentence. It doesn't add much and seems to trail off.

Can we change the description to just:

> Division IDs and subtypes of divisions this division is a capital of.

vcschapp (changes requested):
This looks awesome. I have a minor comment on the description. I'll try to update the PR myself to avoid blocking this while @DavidKarlas is on vacation.

vcschapp (approved):
Approved.

schema/divisions/division.yaml
vcschapp: I posted an update with this change.

jenningsanderson (approved):
This is critical information, nice!

jwass (approved):

PR #237, opened 2024-07-10
Remove update_time as top-level property. Add update_time as optional sources property.

jwass:
# Description

Removes update_time and adds it to sources

# Reference

# Testing

Passes

# Checklist

1. [X] Add relevant examples.
2. [X] Add relevant counterexamples.
3. [X] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
4. [X] Update in-schema documentation using plain English written in complete sentences, if an update is required.
5. [N/A] Update Docusaurus documentation, if an update is required.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/237/schema)


TristanDiet-TomTom (dismissed):

TristanDiet-TomTom (dismissed):

ibnt1 (dismissed):

jenningsanderson:
After further conversation with @ibnt1 — is it a better proposal to _drop_ `update_time` at the top level this time and introduce it as optional within `sources`?

mojodna:
Why not both?

I think it's valuable to know when an Overture entity changed (if not the current release) without referring to the changelog. It's also valuable to know when the source data was fetched/updated, especially for OSM (in which case, we probably want the feature's `timestamp` vs. when it was snapped). In many cases, they will be the same, but if/when we're conflating information from multiple sources, they'll start to differ.

jwass:
I'm okay removing it from the top-level and moving into sources. Then we can bring it back to the top feature-level as @mojodna suggested but when the pieces are in place to support it correctly.

ibnt1 (approved):

mojodna (approved):

jenningsanderson (approved):

TristanDiet-TomTom (approved):

PR #244, opened 2024-07-16
Use element_at() instead of subscript for tags map key lookup

jwass:
When retrieving a key from a map (like tags in OSM), our queries typically use subscript syntax like tags['highway']. The default behavior of subscript syntax is to fail and error if 'highway' is not in the tags map. However, there is a session property in trino and presto that overrides this behavior allowing it to return null.  It seems that Athena and other systems are deployed with this behavior which makes it somewhat non-standard usage. Instead, we should now use element_at(tags, 'highway') which explicitly returns null when 'highway' is not in the tags map. The behavior should be unchanged when Trino/Presto is configured with the legacy behavior, but will have identical results when those systems are deployed with the more modern defaults.

# Description

TODO.

# Reference

1. TODO.

# Testing

TODO.

# Checklist

1. [ ] Add relevant examples.

# Documentation Website

### DIFF
--- a/task-force-docs/base/infrastructure.sql
+++ b/task-force-docs/base/infrastructure.sql
@@ -8,20 +8,20 @@ WITH classified_osm AS (
     SELECT CAST(
         CASE
             -- Transit
-            WHEN tags['railway'] IN ('station','halt') THEN ROW('transit', 'railway_' || tags['railway'])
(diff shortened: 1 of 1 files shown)


atiannicelli (approved):
Looks good.

jenningsanderson (approved):

PR #257, opened 2024-07-24
Connectors LR

brad-richardson:
# Description

Adds connector ids with pre-computed linear reference values as it isn't possible to accurately compute with the current segment geometry + connector coordinate data. This value will be computed during the transportation prep process when we still have all shape points.

An added benefit here is that in many cases, connector data will no longer need to be scanned until we have additional properties on those features.

Plan is to deprecate connector_ids for removal in October which will be a prereq for declaring transportation as GA-ready.

# Reference

1. https://github.com/OvertureMaps/schema/discussions/219
2. https://github.com/OvertureMaps/schema/discussions/226
3. https://github.com/OvertureMaps/tf-transportation/pull/295

# Testing

Testing done during [implementation](https://github.com/OvertureMaps/tf-transportation/pull/295)

# Checklist

1. [X] Add relevant examples.
4. [X] Add relevant counterexamples.
5. [X] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
6. [X] Update in-schema documentation using plain English written in complete sentences, if an update is required.
7. [X] Update Docusaurus documentation, if an update is required.
8. [X] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/257)

### DIFF
--- a/schema/transportation/segment.yaml
+++ b/schema/transportation/segment.yaml
@@ -61,6 +61,7 @@ properties:
           "$comment": >-
             Each entry in this array is the GERS ID of a transportation connector feature.
         uniqueItems: true
+        minItems: 2
         default: []
       connectors:
         description: >-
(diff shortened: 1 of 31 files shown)


schema/transportation/segment.yaml
```diff
@@ -61,6 +61,26 @@ properties:
           "$comment": >-
             Each entry in this array is the GERS ID of a transportation connector feature.
         uniqueItems: true
+        minItems: 2
+        default: []
+      connectors:
```
ibnt1, line 66: would be good to also update the description for `connector_ids` - that ids should be consumed from the `connectors` and a comment on deprecation

ibnt1 (approved):

mojodna (approved):

jwass (approved):

PascalClarysse-TomTom (approved):

PR #258, opened 2024-07-24
[Non-schema Change] Add sql query files for mappings

atiannicelli:
This adds the queries that map to the mapping that exist today. We have decided that SQL statements are better than the config yaml files that are there now so this converts those yaml files to the SQL files. These will be used by the building pipeline in the tf-data-platform repo once they are merged. 

I also plan on removing the old config files once the tf-data-platform code has been switched over to use these new files.

I also fixed the subtype for bridge structures to be set to "transportation" if it is not filled in by other means.


task-force-docs/buildings/facade_material_query.sql
jenningsanderson, line 1: If it's the same logic for both material and facade, can this be reduced to:
```
WHEN COALESCE( 
        lower(trim(element_at(tags, 'building:material'))), 
        lower(trim(element_at(tags, 'building:facade:material')))
) IN ( ... ) THEN '...'
```

?

atiannicelli: Oh, well, I figured we wanted to go through all of the building:materials first... I figured there was an order... For instance if there is a feature with 
building:material=cement AND building:facade:material=slate
Then, should we mark building_material=brick or building_material=concrete ??

If you use the current logic then we prioritize building_material so it would be building_material=concrete
With your logic it would switch to building=brick

I was trying to keep the same logic as before.

jenningsanderson (approved):

jwass:
I'm all for the element_at() change, but is the lower(trim(...)) really necessary? This feels like correcting OSM data that should instead be fixed at the source.

atiannicelli:
We do currently "correct/transform" the height values. We convert values with "m" "meters" "feet" "inches" and even ' and '' as units into a standardized height in meters.
I didn't think that the `lower(trim(` code would be considered changing the values as "Brick" is just as acceptable in OSM as "brick". It is not a big number that we will convert, but I just thought we don't want to go through OSM and start fixing capitalization issues.

By adding this code we will convert: 
3,384 Building Materials (out os a total of 1,876,586 non null building:material or building:facade:material features)
6,514 Roof Materials (out of a total of 1,611,772 non null roof:material features)
5,551 Roof Shapes (out of a total of 6,874,732 non null roof:shape features)

Also, we currently use [trim](https://github.com/OvertureMaps/tf-data-platform/blob/dev/datapipelines/airflow/dags/overture/osm_tags_to_overture.py#L307) and [lower](https://github.com/OvertureMaps/tf-data-platform/blob/dev/datapipelines/airflow/dags/overture/osm_tags_to_overture.py#L351) in the names conversion. I thought it made sense to use it any place that we are comparing strings from OSM to a list of string values.

jwass:
Ah, right. I was actually thinking you were trimming or lowering that actual tags themselves, not the values. Yeah this makes total sense.

PR #330, opened 2025-03-17
Remove bbox from schema

jenningsanderson:
# Description

So here's the deal:

Since `bbox` is a top-level GeoJSON property, validators are upset with the notion of including it inside of `properties`, even though it's different.

Ultimately, our JSON schema declares how Overture data _should be_ represented as GeoJSON. A complete and valid conversion of our data from parquet to GeoJSON is as follows:

```
{
  "id": <id as string>,
  "bbox": <Build from `bbox` struct as [bbox.xmin, bbox.ymin, bbox.xmax, bbox.ymax],
  "geometry": <Convert from WKB to GeoJSON>,
  "properties": {
     ...everything else (Excluding id, bbox, geometry)
  }
}
```

Rather than include the bbox struct in the JSON schema, we should be very clearly documenting the above transformation in our schema docs and describing the `bbox` struct and how to use it in our docs. 

(We could probably do more to describe just the _difference_ between a GeoJSON representation and our parquet data structure). 

# Reference

# Testing

# Checklist

1. [ ] Add relevant examples.

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/330)

### DIFF
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -44,29 +44,6 @@ description: Common schema definitions shared by all themes
           type: array
           items: { "$ref": "#/$defs/propertyDefinitions/nameRule" }
           minItems: 1
-    bbox:
-      title: Bounding box that fully contains the feature.
-      type: object
-      description: >-
-        A bounding box object explicitly defining the 2d bounds of the feature.
-        Can be converted to an official GeoJSON bbox array as `[bbox.xmin, bbox.ymin, bbox.xmax, bbox.ymax]`
-      properties:
-        xmin:
-          type: number
-          minimum: -180
-          maximum: 180
-        xmax:
-          type: number
-          minimum: -180
-          maximum: 180
-        ymin:
-          type: number
-          minimum: -90
-          maximum: 90
-        ymax:
-          type: number
-          minimum: -90
-          maximum: 90
     commonNames:
       description: The common translations of the name.
       type: object
@@ -404,4 +381,3 @@ description: Common schema definitions shared by all themes
         type: { "$ref": "#/$defs/propertyDefinitions/featureType" }
         version: { "$ref": "#/$defs/propertyDefinitions/featureVersion" }
         sources: { "$ref": "#/$defs/propertyDefinitions/sources" }
-        bbox: { "$ref" : "#/$defs/propertyDefinitions/bbox"}


vcschapp:
I thought `bbox` was separate from `properties` in GeoJSON....

> ### 5/ [Bounding Box](https://datatracker.ietf.org/doc/html/rfc7946#section-5)
>
>   A GeoJSON object MAY have a member named "bbox" to include
   information on the coordinate range for its Geometries, Features, or
   FeatureCollections.  The value of the bbox member MUST be an array of
   length 2*n where n is the number of dimensions represented in the
   contained geometries, with all axes of the most southwesterly point
   followed by all axes of the more northeasterly point.  The axes order
   of a bbox follows the axes order of geometries.

How are validators struggling with that, and *which* validators are struggling?

vcschapp:
[Change classification](https://lf-overturemaps.atlassian.net/wiki/spaces/SCHEM/pages/14286874/Schema+versioning+and+stability#Change-Classification): Isn't this a **major** change?

image

jenningsanderson:
> I thought `bbox` was separate from `properties` in GeoJSON....
> 
> > ### 5/ [Bounding Box](https://datatracker.ietf.org/doc/html/rfc7946#section-5)
> > A GeoJSON object MAY have a member named "bbox" to include
> > information on the coordinate range for its Geometries, Features, or
> > FeatureCollections.  The value of the bbox member MUST be an array of
> > length 2*n where n is the number of dimensions represented in the
> > contained geometries, with all axes of the most southwesterly point
> > followed by all axes of the more northeasterly point.  The axes order
> > of a bbox follows the axes order of geometries.
> 
> How are validators struggling with that, and _which_ validators are struggling?

Our validators in central data pipeline :/

vcschapp (approved):

TristanDiet-TomTom (approved):

DavidKarlas (approved):

vcschapp:
We had quite a lengthy discussion about this in today's [Schema TF meeting](https://github.com/OvertureMaps/schema-wg/issues/353).

Here is my summary of the main points.

1. @jenningsanderson mentioned that the proximate cause of the decision to remove is that the presence of `bbox` broke our custom schema conversion logic in the data pipeline.
    - And that this highlights the need to generate the GeoParquet schema as part of the schema build process and include the artifact in our releases.
3. @TristanDiet-TomTom raised the issue that we need to have a principled framework around changes like this, *i.e.* Tristan was trying to focus on the root cause.
    - There's a bit of a gray zone as to whether the JSON Schema is the schema for the canonical format (GeoJSON) or whether it also shades into schematizing things that are properly considered implementation details of a different format, *i.e.* GeoParquet.
    - Group consensus was it should be the former.
    - Hence the group was aligned with removing `bbox`.
4. We noted that our funky Parquet `bbox` is a bit like the `id` and `geometry` properties in that some kind of special translation is required to go between "flat" GeoParquet and the GeoJSON structure.
5. Overall, the consensus was that we should ideally be able to solve problems like GeoJSON *versus* GeoParquet `bbox` with tooling.
    - @an-rose re-affirmed that if we send her an RFP for tooling, she'll look to hire on it.
    - @danabauer suggested working with Matthias from Fiboa who may be available and who has directly relevant expertise and experience.

PR #354, opened 2025-07-02
[Places] Initial commit of places operating status

jwass:
# Major change release plan

Minor release as its a new optional column

Add operating_status field

TODO.

# Description

Allow Overture to include places that are known to be closed

# Reference

[*List of relevant links to GitHub issues, PRs, and other documentation.*
](https://github.com/OvertureMaps/rfd/discussions/37)

# Testing

Added example and counterexample

# Checklist

1. [x ] Add relevant examples.
2. [ x] Add relevant counterexamples.
3. [ x] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/34)

### DIFF
--- a/counterexamples/places/bad-operating-status.yaml
+++ b/counterexamples/places/bad-operating-status.yaml
@@ -1,0 +1,48 @@
+---
+id: overture:places:place:1
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  categories:
+    primary: some_category
+  confidence: 0.9
+  websites:
+    - https://www.example.com
+  emails:
+    - info@example.com
+  socials:
+    - https://www.twitter.com/example
+  phones:
+    - +32 1207
+  brand:
+    names:
+      primary: My Sweet POI Brand
+    wikidata: Q1000
+  addresses:
+    - freeform: "770 Broadway, Floor 8"
+      locality: "New York"
+    - freeform: "770 Brodway, #802"
+      locality: "New York"
+      region: "US-NY"
+      country: "US"
+  # Overture properties
+  theme: places
+  type: place
+  version: 1
+  sources:
+    - property: ""
+      dataset: metaPlaces
+      record_id: "10101"
+    - property: "/properties/brand"
+      dataset: msftPlaces
+      record_id: "10df72b8"
+  names:
+    primary: My Sweet POI
+    common:
+      es: Something in Spanish
+    rules:
+      - variant: short
+        value: MSPOI
+  operating_status: blah
--- a/counterexamples/places/bad-operating-status.yaml
+++ b/counterexamples/places/bad-operating-status.yaml
@@ -8,8 +8,7 @@ properties:
   categories:
     primary: some_category
   confidence: 0.9
-  websites:
-    - https://www.example.com
+    - https://www.example.com/
   emails:
     - info@example.com
   socials:
--- a/counterexamples/places/bad-operating-status.yaml
+++ b/counterexamples/places/bad-operating-status.yaml
@@ -23,7 +23,7 @@ properties:
   addresses:
     - freeform: "770 Broadway, Floor 8"
       locality: "New York"
-    - freeform: "770 Brodway, #802"
+    - freeform: "770 Broadway, #802"
       locality: "New York"
       region: "US-NY"
       country: "US"
--- a/examples/places/place-with-operating-status.yaml
+++ b/examples/places/place-with-operating-status.yaml
@@ -9,7 +9,7 @@ properties:
     primary: some_category
   confidence: 0.9
   websites:
-    - https://www.example.com
+    - https://www.example.com/
   emails:
     - info@example.com
   socials:
--- a/examples/places/place-with-operating-status.yaml
+++ b/examples/places/place-with-operating-status.yaml
@@ -1,0 +1,48 @@
+---
+id: overture:places:place:1
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  categories:
+    primary: some_category
+  confidence: 0.9
+  websites:
+    - https://www.example.com
+  emails:
+    - info@example.com
+  socials:
+    - https://www.twitter.com/example
+  phones:
+    - +32 1207
+  brand:
+    names:
+      primary: My Sweet POI Brand
+    wikidata: Q1000
+  addresses:
+    - freeform: "770 Broadway, Floor 8"
+      locality: "New York"
+    - freeform: "770 Brodway, #802"
+      locality: "New York"
+      region: "US-NY"
+      country: "US"
+  # Overture properties
+  theme: places
+  type: place
+  version: 1
+  sources:
+    - property: ""
+      dataset: metaPlaces
+      record_id: "10101"
+    - property: "/properties/brand"
+      dataset: msftPlaces
+      record_id: "10df72b8"
+  names:
+    primary: My Sweet POI
+    common:
+      es: Something in Spanish
+    rules:
+      - variant: short
+        value: MSPOI
+  operating_status: closed
--- a/examples/places/place-with-operating-status.yaml
+++ b/examples/places/place-with-operating-status.yaml
@@ -45,4 +45,4 @@ properties:
     rules:
       - variant: short
         value: MSPOI
-  operating_status: closed
+  operating_status: permanently_closed
--- a/examples/places/place-with-operating-status.yaml
+++ b/examples/places/place-with-operating-status.yaml
@@ -23,7 +23,7 @@ properties:
   addresses:
     - freeform: "770 Broadway, Floor 8"
       locality: "New York"
-    - freeform: "770 Brodway, #802"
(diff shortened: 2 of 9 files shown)


schema/places/place.yaml
```diff
@@ -93,3 +93,9 @@ properties:
           "$ref": "../defs.yaml#/$defs/propertyDefinitions/address"
         uniqueItems: true
         minItems: 1
+      operating_status:
+        description: >-
+          Indicates if the place is no longer open. Intended to expand to "temporary closed" or others.
+          When operating_status is missing, the place is assumed to be open.
```
vcschapp: I assume there's a reason why we don't want to provide an explicit `"open"` enumeration member. Can you explain why?

jensgoossens-tomtom: I've thought about this as well. 

not including 'open' and infer that it's open when its missing makes it that we can roll out this schema change without doing any code changes in the entire pipeline, so we can decouple the release timings. (it's more of a convenience). I do agree that giving the naming of the field, you would expect an 'open' status there as well. 

Most data suppliers don't explicitly mention it's open - it's inferred. However I think there is an alternative to this, by only modelling for closure status.
f.e. add a property:
out_of_operation with values ['permanent', 'temporary', 'seasonally'] 

But that closes the door if we want to adapt of more labeling/bucketing style like f.e. foursquare does in their 'closed_bucket' field with values: LikelyOpen/LikelyClosed/Unsure/..

Thoughts @vcschapp  / @jwass / @rahulpratapm ?

jensgoossens-tomtom:
> > How does this fit in with [confidence=0](https://github.com/OvertureMaps/schema/blob/6bb9176217e61eaa717d8e2b840551d912205528/schema/places/place.yaml#L45)?
> 
> Good question. This is something we should document carefully also.
> 
> (Related: Are we still thinking about switching `confidence` to `existence`?)

We're still debating on this one - for now we settled on setting confidence to 0 when it's also marked as closed.
From discussions happening - I think the confidence score will be redefined to indeed more like 'existence' to indicate whether the place ever existed or not, and used more to weed out junk vs real places (even places that are now closed, but used to exist)

schema/places/place.yaml
ericgodwin: This was discussed in the 8/21/2025 Places TF meeting.  The decision was that we should roll out with three initial enumeration values:  open, closed, and temporarily closed.  This allows us to support future states without having to make another schema change.  Although it may require a code change, the desire is that we default to open if we do not explicitly know that a place is closed.  @jensgoossens-tomtom and @jwass does this make sense?

ericgodwin: Clarification, the list would be:  open, permanently closed, and temporarily closed for clarity.

jensgoossens-tomtom: Updated the enum values accordingly in the PR

jensgoossens-tomtom: Additionally added a line of documentation to the confidence field that it gets set to 0 if a place is marked as 'closed' in operating_status

vcschapp: Separate from this, but are we going to rename `confidence` to `existence`?

vcschapp: I would read "closed with confidence 0" as "you are not confident it is closed".

schema/places/place.yaml
```diff
@@ -93,3 +93,9 @@ properties:
           "$ref": "../defs.yaml#/$defs/propertyDefinitions/address"
         uniqueItems: true
         minItems: 1
+      operating_status:
+        description: >-
+          Indicates the operating status of a place, can be one of ["open","permanently_closed", "temporarily_closed"]
+        type: string
+        default: "open"
+        enum: ["open","permanently_closed", "temporarily_closed"]
```
vcschapp: nit: add space after first comma

vcschapp:
@mojodna @jenningsanderson @jensgoossens-tomtom This change will have to be ported to the Pydantic model.

vcschapp (approved):

counterexamples/places/bad-operating-status.yaml
```diff
@@ -0,0 +1,48 @@
...
+  phones:
+    - +32 1207
+  brand:
+    names:
+      primary: My Sweet POI Brand
+    wikidata: Q1000
+  addresses:
+    - freeform: "770 Broadway, Floor 8"
+      locality: "New York"
+    - freeform: "770 Brodway, #802"
```
sethfitz:
```suggestion
    - freeform: "770 Broadway, #802"
```

examples/places/place-with-operating-status.yaml
```diff
@@ -0,0 +1,48 @@
...
+  phones:
+    - +32 1207
+  brand:
+    names:
+      primary: My Sweet POI Brand
+    wikidata: Q1000
+  addresses:
+    - freeform: "770 Broadway, Floor 8"
+      locality: "New York"
+    - freeform: "770 Brodway, #802"
```
sethfitz:
```suggestion
    - freeform: "770 Broadway, #802"
```

examples/places/place-with-operating-status.yaml
```diff
@@ -0,0 +1,48 @@
...
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  categories:
+    primary: some_category
+  confidence: 0.9
+  websites:
+    - https://www.example.com
```
sethfitz: As above

```suggestion
    - https://www.example.com/
```

counterexamples/places/bad-operating-status.yaml
```diff
@@ -0,0 +1,48 @@
...
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  categories:
+    primary: some_category
+  confidence: 0.9
+  websites:
+    - https://www.example.com
```
sethfitz: This is dumb, but Pydantic's `HttpUrl` type automatically appends trailing slashes to make them "valid" URLs and can't be configured otherwise.

```suggestion
    - https://www.example.com/
```

schema/places/place.yaml
```diff
@@ -93,3 +93,9 @@ properties:
           "$ref": "../defs.yaml#/$defs/propertyDefinitions/address"
         uniqueItems: true
         minItems: 1
+      operating_status:
+        description: >-
+          Indicates the operating status of a place, can be one of ["open", "permanently_closed", "temporarily_closed"]
+        type: string
+        default: "open"
```
sethfitz: `default` is weird. Semantically, it means that if absent, it should be assumed to be "open". However, depending how the JSON Schema (or Pydantic model) is used, it will transform data being parsed such that the output will differ from the input by including the default value. One interpretation of publishing a "validated" dataset is that it will contain values that have been transformed. If the JSON Schema (or artifacts derived from Pydantic models) is used as a reference, you'll still have `""` combined with the knowledge that it "means" "open".

As such, my preference would be to update the description with the default value. We could drop `default` to eliminate the ambiguity, but derivatives (documentation, primarily) can use this as a cue to show the semantically default value.

mojodna:
Pydantic equivalent here: https://github.com/OvertureMaps/schema/pull/389

schema/places/place.yaml
larryfreil: I'd also recommend expanding the definition with some guidance his this compares to the idea of 'opening hours'. Namely that saying to the effect of "This is not meant to be an indication of 'opening hours' or a business is open at a the current time-of-day or day-of-week, but that the business is open at least once during the course of a few days or weeks."

jensgoossens-tomtom: @mojodna - How about removing the default, but also make the operating_status required - if that makes sense?

sethfitz: That works for me, although it means that the Places pipeline would need to *always* populate that field. It's certainly better from a data consumer perspective, as what you see in the data (absent reading any documentation) is enough to interpret the meaning.

larryfreil (dismissed):
Thank you for adding the clarity to the description.

schema/places/place.yaml
jensgoossens-tomtom: We've already setup the PR in the Places pipeline to set it fixed to open until we tackle the individual feeds and map the value through where it exists:

https://github.com/OvertureMaps/tf-data-platform/pull/2047/files

mojodna (approved):

jenningsanderson (approved):

schema/places/place.yaml
jwass: Resolved by removing default field

schema/places/place.yaml
jwass: The combination of both existence and operating_status needs to be thought through a lot more. In the meantime, we have decided that confidence of 0 means "place is not open", not "we have no confidence in the operating status field". Places group will be addressing this near-term but is okay to merge for now.

PR #355, opened 2025-07-03
Improve example validity

mojodna:
# Description

This fixes a number of (minor) validity problems with the examples discovered while experimenting with Zod and Pydantic.

# Reference

n/a

# Testing

`./test.sh` passes successfully.

These were also tested as part of (partial) Zod and Pydantic ports of the JSON Schema.

PR #356, opened 2025-07-03
Improve example validity

mojodna:
# Description

This fixes a number of (minor) validity problems with the examples discovered while experimenting with Zod and Pydantic.

Replaces #355.

# Reference

n/a

# Testing

`./test.sh` passes successfully.

These were also tested as part of (partial) Zod and Pydantic ports of the JSON Schema.


vcschapp (approved):
```
        d8b                 
        Y8P                 
                            
88888b. 888 .d8888b .d88b.  
888 "88b888d88P"   d8P  Y8b 
888  888888888     88888888 
888  888888Y88b.   Y8b.     
888  888888 "Y8888P "Y8888  
```

PR #360, opened 2025-07-14
[Divisions] Add support for Geographical Regions

DavidKarlas:
# Description

There are many regions in the world that people often search and some apps might want to highlight on map when searched for. Hence proposing adding new subtype called "geographical_region", with "classes", "disputed"(https://github.com/OvertureMaps/rfd/discussions/52) and "region" for start.

# Reference

https://github.com/OvertureMaps/tf-admin/issues/100
https://github.com/OvertureMaps/rfd/discussions/52
https://github.com/OvertureMaps/tf-admin/issues/102

# Testing

Added example cases.

# Checklist

1. [x] Add relevant examples.

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/360)


sasastanojkov (dismissed):

schema/divisions/division.yaml
```diff
@@ -52,6 +59,16 @@ properties:     # JSON Schema: Top-level object properties.
...
           - hamlet        #A small, isolated human settlement in a rural area
                           #Example: Tjarnabyggð, Iceland.
+
+          - disputed      #A geographical region that is claimed by multiple countries or entities,
+                          #and is not universally recognized as belonging to any one of them.
+                          #Example: Bir Tawil.
+
+          - region        #A general geographical region.
+                          #Example: Middle east.
+          - former_administrative_division # A division that no longer exists, but is still relevant as people
```
stepps00: Should these three additions to `class` be included in this PR? If so, can you explain the use case for each?

Asking mostly because the `former_administrative_division` feels different than the other values - it is just indicating "currency." I'm thinking with the Sogn og Fjordane example, the class could be set to `county`, but a boolean could indicate whether or not that place is "current."

DavidKarlas: I gave some more thought to this, and agree, former_administrative_division need some more thought, removed it for now, probably best to be part of separate PR.

docs/schema/reference/divisions/division.mdx
```diff
@@ -190,6 +191,22 @@ A microhood is a mini-neighborhood that is contained within a neighborhood. Exam
   </TabItem>  
 </Tabs>
```
sasastanojkov, line 192:
```suggestion
```

vcschapp:
1. Should we be adding the linked RFD into (RFD 52) into the schema meeting so schema TF can be involved?
2. This PR seems to go beyond RFD 52 in some respects. Why, and what are the use cases? Especially, what is the use case for `subtype=geographical_region,class=region` versions just `subtype=disputed` as @stepps00 suggested in the RFD?

vcschapp:
Closing this PR, as it is over 1Y old, @DavidKarlas isn't actively working on Overture any more, and to some degree this has been superseded by the current work on colloquial areas.

PR #365, opened 2025-07-30
[Pydantic][3/12] overture-schema-validation

mojodna:
Constraint-based validations:

- Base constraint classes (BaseConstraint, StringConstraint, CollectionConstraint)
- String pattern validation (language tags, country codes, ISO dates, JSON pointers)
- Collection validation (uniqueness, size constraints, composite uniqueness)
- Numeric constraints (confidence scores, zoom levels, non-negative values)
- Conditional validation (required-if, mutually exclusive, at-least-one-of)
- Mixin-based validation system with decorators for model-level
  validation
- JSON Schema generation using constraint metadata


packages/overture-schema-validation/README.md
```diff
@@ -0,0 +1,340 @@
...
+from overture.schema.validation import (
+    LanguageTagConstraint,
+    CountryCodeConstraint,
+    UniqueItemsConstraint,
+)
+
+class PlaceProperties(StrictBaseModel):
+    # String pattern validation
+    language: Annotated[str, LanguageTagConstraint()] = Field(
+        ..., description="IETF BCP-47 language tag"
```
brad-richardson, line 41: Since this is the first example in the list, it is probably worth showing how you actually build the full field. I initially thought `...` was some kind of fancy syntax

packages/overture-schema-validation/README.md
```diff
@@ -0,0 +1,340 @@
...
+##### ⚠️ CRITICAL: Inheritance Order Matters
+
+When using `ConstraintValidatedModel`, it **MUST** come first in the inheritance list:
+
+```python
+# ✅ CORRECT - ConstraintValidatedModel first
+class MyModel(ConstraintValidatedModel, BaseModel):
+    pass
+
+# ❌ WRONG - Will not generate JSON Schema metadata
```
brad-richardson, line 261: Any tests we could add to ensure this? Or more of a gotcha when adding new things?

packages/overture-schema-validation/README.md
```diff
@@ -0,0 +1,340 @@
...
+```python
+# Invalid country code
+ValidationError: 1 validation error for MyModel
+country
+  Invalid ISO 3166-1 alpha-2 country code: USA [type=value_error]
+
+# Mutual exclusion violation
+ValidationError: 1 validation error for BoundaryModel
+is_land, is_territorial
+  Fields is_land, is_territorial are mutually exclusive and cannot all be true [type=value_error]
```
brad-richardson, line 283: Is this custom? Could we have an option to produce more machine-friendly output?

packages/overture-schema-validation/README.md
sethfitz: The correct (equivalent, but clearer) syntax, where you only need to provide the type, is:

```python
class PlaceProperties(StrictBaseModel):
    # String pattern validation
    language: Annotated[str, LanguageTagConstraint(), Field(
        description="IETF BCP-47 language tag"
    )]
```

`LanguageTag` is a type alias for that whole thing, so in practice, it's `language: LanguageTag`, and if it's optional, `language: LanguageTag | None = None`.

`...` was also new syntax to me (it's valid!). It means "not implemented"; in Pydantic, as a value, it means "this field has no default, so it's required".

(READMEs all need an update pass.)

packages/overture-schema-validation/README.md
sethfitz: Yes. It's changed a bit since "mutually exclusive" was the wrong way to model it (and even the current approach is super awkward), but we own the error messages: https://github.com/OvertureMaps/schema/pull/365/files#diff-d05a1ec15406f4dd8f1889d2eb7427b5bd72aaf63c5a28c5d594b4bbca15032eR260-R268

packages/overture-schema-validation/README.md
sethfitz: Maybe. It's more of a gotcha, but we may be able to throw at runtime if it's in the wrong order (or maybe not; subclasses may not call it as a `super()`, so we'd be out of the loop). I'll investigate more; flagging the gotcha programmatically would be great.

vcschapp:

PR #369, opened 2025-07-30
[Pydantic][7/12] overture-schema-base-theme

mojodna:
- Introduce models for all types in the base theme
- Enable base theme examples and counterexamples


packages/overture-schema-base-theme/src/overture/schema/base/land_use/models.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Core
+
+    theme: Literal["base"]
+    type: Literal["land_use"]
+    geometry: Annotated[
+        Geometry,
+        GeometryTypeConstraint("Point", "LineString", "Polygon", "MultiPolygon"),
+        Field(
+            description="Classifications of the human use of a section of land. Translates `landuse` from OpenStreetMap tag from OpenStreetMap.",
```
jenningsanderson, line 32: This description is _on_ the geometry?

sethfitz: Today, yes: https://github.com/OvertureMaps/schema/blob/dev/schema/base/land_use.yaml#L10

(Leaving cleaning up docstrings / descriptions for subsequent refactoring.)

PR #371, opened 2025-07-30
[Pydantic][9/12] overture-schema-divisions-theme

mojodna:
- Introduce models for the divisions theme
- Add ParentDivisionValidator
- Enable divisions examples and counterexamples

The divisions theme introduces complex geographical and administrative boundary
validation with conditional parent-child relationships.


packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
```diff
@@ -0,0 +1,186 @@
...
+        Side | None,
+        Field(
+            description="Side of the road on which vehicles drive in the division.",
+        ),
+    ] = None
+
+
+@parent_division_required_unless("subtype", PlaceType.COUNTRY)
+class Division(OvertureFeature, Named, CartographicallyHinted):
+    """Divisions are recognized official or non-official organizations of people as seen from a given political perspective. Examples include countries, provinces, cities, towns, neighborhoods, etc."""
```
vcschapp, line 51: This seems to be going over the 88-column limit set in the root `pyproject.toml`. I assume this is because the auto-formatter doesn't know how to break up long docstrings, but maybe we should.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Core
+
+    theme: Literal["divisions"]
+    type: Literal["division"]
+    geometry: Annotated[
+        Geometry,
+        GeometryTypeConstraint("Point"),
+        Field(
+            description="""Division geometry MUST be a Point as defined by GeoJSON schema. It represents the approximate location of a position commonly associated with the real-world entity modeled by the division feature.""",
```
vcschapp, line 63: This is obviously a direct port of the current JSON Schema. We should do a full-on update pass to rationalize all the descriptions and docstrings.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
```diff
@@ -0,0 +1,186 @@
...
+            description="""Collects information about local norms and rules within the division that are generally useful for mapping and map-related use cases.
+
+If the norms property or a desired sub-property of the norms property is missing on a division, but at least one of its ancestor divisions has the norms property and the desired sub-property, then the value from the nearest ancestor division may be assumed.""",
+        ),
+    ] = None
+    population: Annotated[
+        int | None, Field(ge=0, description="Population of the division")
+    ] = None
+    capital_division_ids: Annotated[
+        list[Id] | None,
```
vcschapp, line 171: We should be able to annotate `Id` with what it refers to.

Something like `Annotated[Id, RefersTo(Referee.FEATURE_TYPE, "division"))]`...

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
sethfitz: I added `docformatter` in a follow-up commit (that's not part of a PR yet). That takes care of this in an automated way.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
sethfitz: Agreed.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
```diff
@@ -0,0 +1,89 @@
...
+            strict=True,
+        ),
+    ] = None
+    is_territorial: Annotated[
+        bool | None,
+        Field(
+            description="""A boolean to indicate whether or not the feature geometry represents Overture's best approximation of this place's maritime boundary. For coastal places, this would tend to include the water area. The geometry can be used for data processing, reverse-geocoding, and similar purposes.""",
+            strict=True,
+        ),
+    ] = None
```
vcschapp, line 68: This is a comment on the schema today, not the Pydantic port, but I'm struggling to remember and/or rationalize how we have both `class` and `is_land`/`is_territorial`. ...

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
```diff
@@ -0,0 +1,89 @@
...
+            description="Division area geometries MUST be polygons or multi-polygons as defined by the GeoJSON schema.",
+        ),
+    ]
+
+    # Required
+
+    names: Names
+    subtype: PlaceType
+    class_: Annotated[
+        AreaBoundaryClass,
```
vcschapp, line 52: Naming: This should be `AreaClass`.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/enums.py
```diff
@@ -0,0 +1,71 @@
+from enum import Enum
+
+
+class PlaceType(str, Enum):
```
vcschapp, line 4: Naming: We should call this `DivisionSubType` or something.

While it's based on WOF, it may diverge - and anyway the whole "Place" noun and prefix can create confusion with the places theme so it'd be better not to use it here.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
sethfitz: I dug into `NewType` recently as a way to distinguish between different uses of `Annotated[str, ...]`, i.e., `RegionCode` and `CountryCode` are both strings, but they have different validation rules and represent different things. That led me to thinking about `Id` as a generic type (`Id[Literal["division"]]`) with the requisite type checks.

However, I was also thinking about what the applicability of this is, particularly when working beyond Python, and the first thing that came to mind was foreign keys, specifically outputting them (documentation and database DDL where they're supported, for example). I have a rough sketch, but it doesn't quite work like I want, and I'm not concrete enough on the value it provides.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
sethfitz: https://github.com/mojodna/schema/commit/f78767e5c1bb2b79ce1bea745d9cb7fedac12bb4

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
sethfitz: https://github.com/mojodna/schema/commit/80e54a27860cce083e7a688b76e69f59f5a0cd74

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
sethfitz: Difficulties modeling the validation rules for `is_land` / `is_territorial` are another smell.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
sethfitz: Fixed: https://github.com/mojodna/schema/commit/0f8bf2e46cf52181148f38265f4b26cdb019c69a

(but not PR'd yet)

vcschapp:

PR #373, opened 2025-07-30
[Pydantic][11/12] overture-schema-transportation-theme

mojodna:
- Introduce transportation models with supporting types and validations
- Introduce scoping system to overture-schema-core with support for
  geometric, temporal, directional, and vehicle-specific rules

The scoping system uses a mix-in architecture enabling precise conditional
application of transportation rules based on time, direction, vehicle type,
and geometric position along segments.


packages/overture-schema-core/README.md
```diff
@@ -1,54 +1,139 @@
...
-class MyFeature(OvertureFeature):
-    name: str
+#### Temporal Speed Limit
+
+```yaml
+speed_limits:
+  - between: [0, 1]
+    max_speed: {value: 30, unit: km/h}
+    when:
+      during: "Mo-Fr 07:00-09:00,17:00-19:00"  # Rush hours only
```
brad-richardson, line 95: Leaving a note so I don't forget - do we have any validation for this pattern?

packages/overture-schema-transportation-theme/src/overture/schema/transportation/enums.py
```diff
@@ -0,0 +1,252 @@
...
+    CAMP_SITE = "camp_site"
+    INTERCHANGE = "interchange"
+    RESTROOMS = "restrooms"  # 'toilets' in OSM
+
+
+class RoadFlag(str, Enum):
+    """Simple flags that can be on or off for a road segment. Specifies physical characteristics and can overlap."""
+
+    IS_BRIDGE = "is_bridge"
+    IS_LINK = "is_link"  # Note: `is_link` is deprecated and will be removed in a future release in favor of the link subclass
```
brad-richardson, line 117: This is probably a good time to action on this. cc: @danabauer @Rachmanin0xFF

packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py
```diff
@@ -0,0 +1,561 @@
...
+    """Properties defining travel headings that match a rule."""
+
+    model_config = ConfigDict(frozen=True)
+
+    # Optional
+
+    heading: Heading | None = None
+
+
+# TODO generate minProperties: 1 in JSON Schema (and validate)
```
brad-richardson, line 70: Should this be a `FIXME` before merge? Does this exist in the current schema definition?

brad-richardson: (maybe this isn't terribly critical though)

uv.lock
```diff
@@ -600,6 +603,22 @@ dev = [
     { name = "ruff" },
 ]
 
+[[package]]
+name = "overture-schema-transportation-theme"
```
brad-richardson, line 607: @thehoneymad an FYI - these models or some artifacts might be useful for some of our routing needs

packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py
sethfitz: This is implemented in a cleanup branch (and I think it's reflected in the unified JSON Schema linked from the big PR).

packages/overture-schema-core/README.md
sethfitz: Not yet. I was thinking of using an OSM opening hours parser (now that we're in Python, but it would only apply for Python), but I'm not sure how that relates to the goal of formalizing opening hours as an extension (they may be entirely different things).

vcschapp: We should have a goal to support validating and also parsing it into some kind of usable data structure.

vcschapp:

PR #378, opened 2025-08-08
pydantic data types

mojodna:
- **Replace type aliases with NewType to improve type checks**
- **Introduce abstract data types with target types**

### DIFF
--- a/packages/overture-schema-base-theme/src/overture/schema/base/types.py
+++ b/packages/overture-schema-base-theme/src/overture/schema/base/types.py
@@ -1,28 +1,40 @@
-from typing import Annotated, Any
+from typing import Annotated, Any, NewType
 
 from pydantic import Field
 
-Elevation = Annotated[
-    int,
-    Field(
-        le=9000,
-        description="Elevation above sea level (in meters) of the feature.",
-    ),
-]
+Elevation = NewType(
+    "Elevation",
+    Annotated[
+        int,
+        Field(
+            le=9000,
+            description="Elevation above sea level (in meters) of the feature.",
+        ),
+    ],
+)
 
-Depth = Annotated[
-    int,
-    Field(
-        ge=0,
-        description="Depth below surface level (in meters) of the feature.",
-    ),
-]
+Depth = NewType(
+    "Depth",
+    Annotated[
+        int,
+        Field(
+            ge=0,
+            description="Depth below surface level (in meters) of the feature.",
+        ),
+    ],
+)
 
-Height = Annotated[float, Field(gt=0, description="Height of the feature in meters.")]
+Height = NewType(
+    "Height",
+    Annotated[float, Field(gt=0, description="Height of the feature in meters.")],
+)
 
-SourceTags = Annotated[
-    dict[str, Any],
-    Field(
-        description="Any attributes/tags from the original source data that should be passed through."
-    ),
-]
+SourceTags = NewType(
+    "SourceTags",
+    Annotated[
+        dict[str, Any],
+        Field(
+            description="Any attributes/tags from the original source data that should be passed through."
+        ),
+    ],
+)
--- a/packages/overture-schema-base-theme/src/overture/schema/base/types.py
+++ b/packages/overture-schema-base-theme/src/overture/schema/base/types.py
@@ -2,10 +2,12 @@ from typing import Annotated, Any, NewType
 
 from pydantic import Field
 
+from overture.schema.core.types.abstract.types import Float64, Int32
+
 Elevation = NewType(
     "Elevation",
     Annotated[
-        int,
+        Int32,
         Field(
             le=9000,
             description="Elevation above sea level (in meters) of the feature.",
@@ -16,7 +18,7 @@ Elevation = NewType(
 Depth = NewType(
     "Depth",
     Annotated[
-        int,
+        Int32,
         Field(
             ge=0,
             description="Depth below surface level (in meters) of the feature.",
@@ -26,7 +28,7 @@ Depth = NewType(
 
 Height = NewType(
     "Height",
-    Annotated[float, Field(gt=0, description="Height of the feature in meters.")],
+    Annotated[Float64, Field(gt=0, description="Height of the feature in meters.")],
 )
 
 SourceTags = NewType(
--- a/packages/overture-schema-core/src/overture/schema/core/json_schema.py
+++ b/packages/overture-schema-core/src/overture/schema/core/json_schema.py
@@ -6,11 +6,35 @@ from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
 from pydantic_core import core_schema
 
 from ._cache import get_type_adapter
+from .types.abstract import AbstractTypeDefinition
 
 
-class OptionalFieldGenerator(GenerateJsonSchema):
-    """Simplify the generated JSON Schema for nullable fields by removing null from
-    anyOf and removing null defaults."""
+class EnhancedJsonSchemaGenerator(GenerateJsonSchema):
+    """Enhanced JSON Schema generator with optional field and abstract type support.
+
+    This generator provides two key enhancements over the default Pydantic generator:
+    1. Optional field handling: Simplifies nullable fields by removing null from anyOf
+       and removing null defaults to make fields truly optional.
+    2. Abstract type support: Detects AbstractType annotations and applies explicit
+       JSON schema overrides when specified.
+    """
+
+    def generate_inner(self, schema: core_schema.CoreSchema) -> JsonSchemaValue:
+        """Override to detect abstract types and generate appropriate JSON schema."""
+
+        # Check for metadata that might contain AbstractType information
+        metadata = schema.get("metadata", {})
+
+        # Look for AbstractTypeDefinition in the metadata (from Annotated types)
+        for item in metadata:
+            if isinstance(item, AbstractTypeDefinition):
+                # We found an AbstractTypeDefinition in the metadata
+                explicit_schema = item.json_schema_override
+                if explicit_schema:
+                    return explicit_schema  # type: ignore[no-any-return]
+
+        # Generate standard schema (includes Annotated constraints automatically)
+        return super().generate_inner(schema)
 
     def nullable_schema(self, schema: core_schema.NullableSchema) -> JsonSchemaValue:
         """Generates a JSON schema that matches a nullable schema.
@@ -64,10 +88,10 @@ def json_schema(models: type[BaseModel] | UnionType | type) -> dict[str, Any]:
         TypeError: If models is not a BaseModel or union type.
     """
     if isinstance(models, type) and issubclass(models, BaseModel):
-        return models.model_json_schema(schema_generator=OptionalFieldGenerator)
+        return models.model_json_schema(schema_generator=EnhancedJsonSchemaGenerator)
 
     if get_origin(models) is not None:
         adapter = get_type_adapter(models)
-        return adapter.json_schema(schema_generator=OptionalFieldGenerator)
+        return adapter.json_schema(schema_generator=EnhancedJsonSchemaGenerator)
 
     raise TypeError(f"Expected BaseModel or union type, got {type(models)}")
--- a/packages/overture-schema-core/src/overture/schema/core/types/abstract/__init__.py
+++ b/packages/overture-schema-core/src/overture/schema/core/types/abstract/__init__.py
@@ -1,0 +1,38 @@
+"""Abstract data types for Overture schemas.
+
+This module provides abstract data types with automatic constraint validation
+and multi-target serialization support (Scala, Spark, Parquet, JSON Schema).
+"""
+
+from .abstract_type import AbstractType, AbstractTypeDefinition, AbstractTypeRegistry
+from .types import (
+    Float32,
+    Float64,
+    Int8,
+    Int32,
+    Int64,
+    UInt8,
+    UInt16,
+    UInt32,
+    get_abstract_type,
+    get_target_type,
+)
+
+__all__ = [
+    # Core types
+    "AbstractType",
+    "AbstractTypeDefinition",
+    "AbstractTypeRegistry",
+    # Concrete types
+    "UInt8",
+    "UInt16",
+    "UInt32",
+    "Int8",
+    "Int32",
+    "Int64",
+    "Float32",
+    "Float64",
+    # Utilities
+    "get_target_type",
+    "get_abstract_type",
+]
--- a/packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
+++ b/packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
@@ -1,0 +1,124 @@
+from dataclasses import dataclass, field
+from typing import Annotated, Any
+
+from pydantic import Field
+from pydantic.fields import FieldInfo
+
+
+@dataclass(frozen=True)
+class AbstractTypeDefinition:
+    """Type-safe definition for an abstract data type."""
+
+    base: type[Any]
+    constraints: FieldInfo | None = None
+    target_mappings: dict[str, str] = field(default_factory=dict)
+    json_schema_override: dict[str, Any] | None = None
+
+
+class AbstractTypeRegistry:
+    """Internal registry of abstract type definitions."""
+
+    TYPES = {
+        "UINT8": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=0, le=255),
+            target_mappings={
+                "scala": "Byte",  # Note: Scala Byte is signed, but closest match
+                "spark": "ByteType",
+                "parquet": "INT32",  # Parquet promotes small ints
+            },
+        ),
+        "UINT16": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=0, le=65535),
+            target_mappings={
+                "scala": "Short",  # Note: Scala Short is signed
+                "spark": "ShortType",
+                "parquet": "INT32",
+            },
+        ),
+        "UINT32": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=0, le=4294967295),
+            target_mappings={
+                "scala": "Long",  # Use Long for safety since Int is signed
+                "spark": "LongType",
+                "parquet": "INT64",  # Promote to avoid overflow
+            },
+        ),
+        "INT8": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=-128, le=127),
+            target_mappings={
+                "scala": "Byte",
+                "spark": "ByteType",
+                "parquet": "INT32",
+            },
+        ),
+        "INT32": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=-(2**31), le=2**31 - 1),
+            target_mappings={
+                "scala": "Int",
+                "spark": "IntegerType",
+                "parquet": "INT32",
+            },
+        ),
+        "INT64": AbstractTypeDefinition(
+            base=int,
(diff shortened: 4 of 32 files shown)


packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
+from dataclasses import dataclass, field
+from typing import Annotated, Any
+
+from pydantic import Field
+from pydantic.fields import FieldInfo
+
+
+@dataclass(frozen=True)
```
vcschapp: Any reason why not `slots=True`? There's a performance benefit and the downsides don't really seem to apply to this case.

packages/overture-schema-base-theme/src/overture/schema/base/types.py
```diff
@@ -1,28 +1,42 @@
...
 from pydantic import Field
 
-Elevation = Annotated[
-    int,
-    Field(
-        le=9000,
-        description="Elevation above sea level (in meters) of the feature.",
-    ),
-]
+from overture.schema.core.types.abstract.types import Float64, Int32
```
vcschapp, line 6: Something feels wrong about the module `abstract.types`. They aren't all that abstract if they are being used in a concrete type.

Something also feels wrong about `types.abstract.types`.

---

IMO for export these names should come out of `overture.schema.core.primitive` - a clean-ish module path that doesn't repeat itself, doesn't have unnecessary confusing words (`abstract`) and doesn't lean on a relatively vague word (`type`).

packages/overture-schema-core/src/overture/schema/core/types/abstract/__init__.py
```diff
@@ -0,0 +1,38 @@
+"""Abstract data types for Overture schemas.
+
+This module provides abstract data types with automatic constraint validation
+and multi-target serialization support (Scala, Spark, Parquet, JSON Schema).
```
vcschapp: I feel like Scala is getting baked further and further in as a target assumption.

I really want to push back that the way to support Scala is via Java, and that we shouldn't bother doing both.

I understand that you can get *slightly* more idiomatic usability by directly emitting Scala code in the narrow use case of Scala-on-Spark, but emitting appropriate Java code is just about as usable and has the benefit of being runnable everywhere that Java is supported, including Spark, any Java project, any Scala project, any real Kotlin project, the Android OS, *etc.*

Ultimately a lot of things are nice to have, and having every fancy Scala bell and whistle is definitely one of those nice things, but in a world of constrained resources where support is really hard and in reality people move on and funding dies, having something that's 5% less nice but that works in 50X the use cases is preferable.

----

I don't want to become a zombie project like some I can point to (LinkML among them) where there are 20 generation targets but only 5 of them work and the rest are deprecated, obsolete, broken, unsupported, or all of the above.

I also don't want to start with ambitious dreams ("we'll do it all"), get halfway there, and then lose momentum and have the rest die.

Lastly, I don't want to over-index on "the current Overture internal use case". One of the theories behind [RFD 22](https://github.com/OvertureMaps/rfd/discussions/22), earlier work from [March 2025](https://github.com/OvertureMaps/schema-wg/issues/355) and even earlier work elsewhere is that schema-related efforts can broaden the adoption of Overture by broadening its usability.

Given that Java can do everything of substance that Scala can do, the ROI on Java is much higher than Scala, and I really don't want to have to support both.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
...
+class AbstractTypeDefinition:
+    """Type-safe definition for an abstract data type."""
+
+    base: type[Any]
+    constraints: FieldInfo | None = None
+    target_mappings: dict[str, str] = field(default_factory=dict)
```
vcschapp: - Should target mappings be enumerated rather than `str`?
- Should we have a mechanism to enforce coverage of all the target mappings? Otherwise you find out much later in the process when you try to generate something.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
+from dataclasses import dataclass, field
+from typing import Annotated, Any
+
+from pydantic import Field
+from pydantic.fields import FieldInfo
+
+
+@dataclass(frozen=True)
+class AbstractTypeDefinition:
```
vcschapp: The prefix `Abstract` is throwing me off, for some of the reasons mentioned in an above comment.

Can we name this `typing.TypeMetadata`?

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
...
+    base: type[Any]
+    constraints: FieldInfo | None = None
+    target_mappings: dict[str, str] = field(default_factory=dict)
+    json_schema_override: dict[str, Any] | None = None
+
+
+class AbstractTypeRegistry:
+    """Internal registry of abstract type definitions."""
+
+    TYPES = {
```
vcschapp: This structure feels a bit quirky/non-idiomatic to me.

At minimum I'd say it should be made a private implementation details and the `TYPES` dict should be frozen.

The  `dict`-within-a-`dict` structure also feels heavy without adding value: the expression `AbstractTypeRegistry.TYPES["UINT8"]` needs one lookup of `TYPES` in the `TYPES` `dict` and then another lookup of the `"UINT8"`.

I'd suggest putting this at file scope:

```python
_TYPE_DEFINITIONS = MappingProxyType({
    "UINT8": ...,
})
```

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
...
+        "UINT8": AbstractTypeDefinition(
+            base=int,
+            constraints=Field(ge=0, le=255),
+            target_mappings={
+                "scala": "Byte",  # Note: Scala Byte is signed, but closest match
+                "spark": "ByteType",
```
vcschapp: So is a Spark `ByteType`, and so is anything that's fundamentally JVM-based. It's Java turtles all the way down.

This makes me think, though, that supporting unsigned integer types as first-class citizens may fundamentally be a fool's errand, and we should probably just provide the signed types.

- JSON doesn't care about signs.
- Spark lives purely in the signed world, whether your flavor is Python, or JVM-based Spark
- Python itself lives purely in the signed world

Given the number of core Overture target platforms that don't have unsigned integer support, I think we should drop unsigned integers as a core Overture type, make everything signed, handle unsigned-ness, like everything else, via range constraints.

(Which is what I think you're already implicitly doing in some places like `num_floors_underground` above.)

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
...
+        "FLOAT64": AbstractTypeDefinition(
+            base=float,
+            target_mappings={
+                "scala": "Double",
+                "spark": "DoubleType",
+                "parquet": "DOUBLE",
+            },
+            json_schema_override={"type": "number"},
+        ),
+        "GEOMETRY": AbstractTypeDefinition(
```
vcschapp: `BBox`?

packages/overture-schema-core/src/overture/schema/core/types/abstract/types.py
```diff
@@ -0,0 +1,48 @@
...
+Float64 = NewType("Float64", AbstractType["FLOAT64"])  # type: ignore[misc,name-defined,type-arg] # noqa: F821
+
+# Mapping for class-based types that don't use Annotated metadata
+_CLASS_TYPE_MAPPING: dict[type[Any], AbstractTypeDefinition] = {
+    Geometry: AbstractTypeRegistry.TYPES["GEOMETRY"],
+}
+
+
+# Utility functions for easy access
+def get_target_type(concrete_type: type[Any], language: str) -> str | None:
```
vcschapp: If the language decorations are going in the core package, it really feels like they should be enumerated and that all enumerated targets should be non-optional.

packages/overture-schema-core/src/overture/schema/core/json_schema.py
```diff
@@ -6,11 +6,35 @@
...
 
-class OptionalFieldGenerator(GenerateJsonSchema):
-    """Simplify the generated JSON Schema for nullable fields by removing null from
-    anyOf and removing null defaults."""
+class EnhancedJsonSchemaGenerator(GenerateJsonSchema):
+    """Enhanced JSON Schema generator with optional field and abstract type support.
+
+    This generator provides two key enhancements over the default Pydantic generator:
+    1. Optional field handling: Simplifies nullable fields by removing null from anyOf
+       and removing null defaults to make fields truly optional.
```
vcschapp: This is nice.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
```diff
@@ -0,0 +1,124 @@
...
+            },
+        ),
+        "FLOAT32": AbstractTypeDefinition(
+            base=float,
+            target_mappings={
+                "scala": "Float",
+                "spark": "FloatType",
+                "parquet": "FLOAT",
+            },
+            json_schema_override={"type": "number"},
```
vcschapp: Won't this happen anyway?

```python
>>> class Foo(BaseModel):
...     bar: float
... 
>>> Foo.model_json_schema()
{'properties': {'bar': {'title': 'Bar', 'type': 'number'}}, 'required': ['bar'], 'title': 'Foo', 'type': 'object'}
```

packages/overture-schema-core/src/overture/schema/core/json_schema.py
```diff
@@ -6,11 +6,35 @@
...
+    2. Abstract type support: Detects AbstractType annotations and applies explicit
+       JSON schema overrides when specified.
```
vcschapp: It feels like this might be a solution in search of a problem.

The number types should all get their JSON type from Pydantic automatically, shouldn't they? And the only other two primitive types we need to support are `Geometry` and `BBox` which have `__get_pydantic_json_schema__` implementations that provide the JSON Schema.

packages/overture-schema-validation/src/overture/schema/validation/types.py
```diff
@@ -20,35 +22,53 @@
...
+        str,
+        CountryCodeConstraint(),
+        Field(description="ISO 3166-1 alpha-2 country code"),
+    ],
+)
+HexColor = NewType("HexColor", Annotated[str, HexColorConstraint()])
+ISO8601DateTime = NewType(
+    "ISO8601DateTime", Annotated[str, ISO8601DateTimeConstraint()]
+)
+JSONPointer = NewType("JSONPointer", Annotated[str, JSONPointerConstraint()])
```
vcschapp: nit: `JsonPointer`

packages/overture-schema-validation/src/overture/schema/validation/types.py
```diff
@@ -20,35 +22,53 @@
...
+    Annotated[
+        str,
+        CountryCodeConstraint(),
+        Field(description="ISO 3166-1 alpha-2 country code"),
+    ],
+)
+HexColor = NewType("HexColor", Annotated[str, HexColorConstraint()])
+ISO8601DateTime = NewType(
+    "ISO8601DateTime", Annotated[str, ISO8601DateTimeConstraint()]
+)
```
vcschapp: nit `Iso8601DateTime`, but also can we just use `datetime.datetime`? Maybe this is a case for a new type definition since  the ideal in Java would presumably be `java.time.Instant`.

```
>>> from datetime import datetime
>>> from pydantic import BaseModel, Field
>>> from typing import Annotated, NewType
>>> Iso8601Timestamp = NewType("Iso8601Timestamp", Annotated[datetime, Field(..., description="ISO-8601 timestamp", json_schema_extra={"format": "date-time", "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(?:Z|[+-]\\d{2}:\\d{2})$"})])
>>> class Foo(BaseModel):
...   update_time: Iso8601Timestamp
... 
>>> Foo.model_json_schema()
{'properties': {'update_time': {'description': 'ISO-8601 timestamp', 'format': 'date-time', 'pattern': '^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(?:Z|[+-]\\d{2}:\\d{2})$', 'title': 'Update Time', 'type': 'string'}}, 'required': ['update_time'], 'title': 'Foo', 'type': 'object'}
```

vcschapp (changes requested):

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: Nope. 'Twas my first time using data classes. This went away entirely in my refactor.

packages/overture-schema-base-theme/src/overture/schema/base/types.py
sethfitz: I like `overture.schema.core.primitives` (with appropriate wording updated). Moved.

packages/overture-schema-core/src/overture/schema/core/types/abstract/__init__.py
sethfitz: Noted (to other readers: we started talking about this offline earlier in the week). Mentions of Scala have been removed from this PR in favor of only specifying Spark and Parquet mappings in a language-agnostic way.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: Makes sense. This went away entirely in the refactor.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: I cleaned this up substantially in the refactor (while keeping it private). The core intuition was to key things by type rather than a string representation.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: The concrete benefit here is the Pydantic constraints that get tacked on. Data from Shapefiles can have some weird types attached (and data intended to roundtrip with restrictive database schemas), so I'm inclined to keep the list of types as-is. Otherwise I agree that there's little to no benefit in including them.

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: Rebased (now that that's merged) and mapped. Figuring out how to map it led me to switching to the Spark JSON serialization and the Parquet Schema Definition Language for representation, which are all programming language agnostic. It'll be less direct when generating code, but it'll also be less opinionated.

packages/overture-schema-core/src/overture/schema/core/types/abstract/types.py
sethfitz: This makes sense. I'll think about how I want to do that in a follow-up commit. I suspect that it might make things like structs (`BBox`) easier to understand when looking at mappings.

packages/overture-schema-core/src/overture/schema/core/json_schema.py
sethfitz: Yeah, this was an annoying default Pydantic behavior. What is `None` (baby don't hurt me).

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: It will. I was noodling around more complex representations (like `BBox`) and translations between JSON strings and things like `datetime` (where I need to just try it, since it's a JSON string, but probably with a Pydantic-provided pattern).

packages/overture-schema-core/src/overture/schema/core/json_schema.py
sethfitz: See https://github.com/OvertureMaps/schema/pull/378#discussion_r2296496380

sethfitz: It seemed like it might be useful at the time, even though I didn't have a concrete example.

packages/overture-schema-validation/src/overture/schema/validation/types.py
sethfitz: This isn’t new to this PR. We can (and should) fix this separately.

packages/overture-schema-validation/src/overture/schema/validation/types.py
sethfitz: I’ll tackle this in a follow-up commit when I have some free time.

mojodna:
I massively simplified this. There was no reason to attach the metadata (other than field constraints) to the type directly. That, combined with using the type itself as the lookup key, allowed me to simplify the lookups and generally improve it. 

When I have more time, I plan to look into the following:

- using a data class to encapsulate the mappings (or find a way for the mappings to be grouped by target rather than type, but I think that’ll probably be less discoverable when updating)
- using `datetime`
- checking whether `resolve` needs to be part of the public API (since the only purpose for the mappings is to look up target types). There’s value in a function that recursively calls `get_origin` to peel the annotations off—that was something I kept doing in different ways as part of the documentation generation, and the updated approach to target type lookups should make things a lot simpler.

packages/overture-schema-validation/src/overture/schema/validation/types.py
sethfitz: I remember what the problem is. It’s not in the Pydantic modeling, but in translation. IIRC, our date time columns are strings as a side effect of the switch from GeoJSON, so they won’t cleanly map to Instants (using automatic deserialization). Longer term, we might want to consider changing the data pipeline to fix the column types so they’re no longer strings. 

/cc @jenningsanderson

sethfitz: (Not that it matters, but this might buy us some space savings in the Parquet output)

packages/overture-schema-core/src/overture/schema/core/types/abstract/abstract_type.py
sethfitz: Target mappings are going to move into code generation since the mappings aren't useful otherwise. I'm envisioning something that validates that all types are mapped to something (or `None` if it can't be supported for whatever reason) on a per-target basis.

packages/overture-schema-validation/src/overture/schema/validation/types.py
sethfitz: Fixed anyway.

packages/overture-schema-validation/src/overture/schema/validation/types.py
sethfitz: We also lose the ability to specify our own pattern (since `date-time` doesn't come with one), but validators _should_ be smart enough to follow the rules.

mojodna:
Since my last message, I...

- removed the target mappings and related functions (like `resolve`; those belong to the code/schema/documentation generation process and aren't generally useful). this obviated *any* encapsulation of mappings.
- removed our own datetime type (and constraint) in favor of `datetime.datetime`
- renamed `JsonPointer`
- removed additional unnecessary tests

PR #380, opened 2025-08-08
Use formal enum for geometry types (better type safety)

vcschapp:
# Description

This is a minor change to use a formal Python `Enum` class to model the geometry types instead of using string values. 

As well as being generally more consistent, this should allow for better IDE introspection and compatibility with other systems that use Python type annotations.

# Testing

```shell
$ make check
```

### DIFF
--- a/packages/overture-schema-addresses-theme/src/overture/schema/addresses/address/models.py
+++ b/packages/overture-schema-addresses-theme/src/overture/schema/addresses/address/models.py
@@ -10,6 +10,7 @@ from overture.schema.core import (
 )
 from overture.schema.core.geometry import (
     Geometry,
+    GeometryType,
     GeometryTypeConstraint,
 )
 from overture.schema.core.types import CountryCode, TrimmedString
@@ -47,7 +48,9 @@ class Address(Feature[Literal["addresses"], Literal["address"]]):
 
     # Core
     geometry: Annotated[
-        Geometry, GeometryTypeConstraint("Point"), Field(description="Geometry (Point)")
+        Geometry,
+        GeometryTypeConstraint(GeometryType.POINT),
+        Field(description="Geometry (Point)"),
     ]
 
     # Optional
--- a/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
@@ -111,23 +111,19 @@
               "items": {
                 "items": {
                   "items": {
-                    "items": {
-                      "type": "number"
-                    },
-                    "minItems": 2,
-                    "type": "array"
+                    "type": "number"
                   },
-                  "minItems": 4,
+                  "minItems": 2,
                   "type": "array"
                 },
-                "minItems": 1,
+                "minItems": 4,
                 "type": "array"
               },
               "minItems": 1,
               "type": "array"
             },
             "type": {
-              "const": "MultiPolygon",
+              "const": "Polygon",
               "type": "string"
             }
           },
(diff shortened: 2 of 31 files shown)


packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
```diff
@@ -111,23 +111,19 @@
               "items": {
```
vcschapp, line 111: I think we should drop these baseline files - they noise up the diffs a lot and the lack of order-preserving-ness means that a small change in the Python can be a big change in the baseline.

We should just trust our tests and process.

If we want to, we can (and probably should) generate JSON Schema as build artifacts (which can be attached to releases).

packages/overture-schema-addresses-theme/src/overture/schema/addresses/address/models.py
```diff
@@ -47,7 +48,9 @@ class Address(Feature[Literal["addresses"], Literal["address"]]):
 
     # Core
     geometry: Annotated[
-        Geometry, GeometryTypeConstraint("Point"), Field(description="Geometry (Point)")
+        Geometry,
+        GeometryTypeConstraint(GeometryType.POINT),
```
mojodna, line 52: Looking at this makes me think that `Geometry` should be generic over `GeometryType` instead of using an annotation.

packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
sethfitz: Order _is_ preserved (generally; I'm not sure why `Geometry` flipped the order here):  https://github.com/OvertureMaps/schema/blob/d36627a671692a393c10c754e746d68f94419de5/packages/overture-schema-transportation-theme/tests/test_segment_json_schema_baseline.py#L22

I'm not ready to throw them away yet. Doing a hash compare might be ok, but I've found these useful for understanding how a change to a description or docstring somewhere manifests. Seeing unexpected changes here has been the only indication that I did something weird (which, yes, points to a need for broader test coverage).

packages/overture-schema-core/src/overture/schema/core/geometry.py
```diff
@@ -11,31 +12,60 @@
...
-)
+
+class GeometryType(Enum):
+    GEOMETRY_COLLECTION = 0, "GeometryCollection"
+    LINE_STRING = 1, "LineString"
+    POINT = 2, "Point"
+    POLYGON = 3, "Polygon"
+    MULTI_LINE_STRING = 4, "MultiLineString"
+    MULTI_POINT = 5, "MultiPoint"
+    MULTI_POLYGON = 6, "MultiPolygon"
```
sethfitz, line 23: What's the purpose of the numbers? (For me and for posterity.)

packages/overture-schema-core/src/overture/schema/core/geometry.py
```diff
@@ -11,31 +12,60 @@
...
+    @staticmethod
+    def _generate_next_value(
+        _name: str, _start: int, count: int, _last_values: list[Any]
+    ) -> int:
+        return count + 1
+
+    def __lt__(self, other: "GeometryType") -> bool:
+        if not isinstance(other, GeometryType):
+            return NotImplemented
+        return self._value < other._value
```
mojodna, line 38: What is this used for?

sethfitz: What are these used for?

mojodna (approved):

packages/overture-schema-core/src/overture/schema/core/geometry.py
vcschapp: Good questions. `_generate_next_value` can go now, I believe, that's an artifact of an earlier aborted attempt to make `enum.auto()` work with tuples, which isn't possible in Python 3.10.

`__lt__` is to make the enum sortable, which is used indirectly in at least one place (look for `sorted(...)`).

packages/overture-schema-core/src/overture/schema/core/geometry.py
vcschapp: To make it a standard enum that happens to have a string property, rather than a string enum.

PR #313, opened 2025-01-15
[WIP] Add spatial index container

jenningsanderson:
# Category

What kind of change is this?

Please select *one* of the following five options.

Consult [Pull request merging criteria](https://github.com/OvertureMaps/schema-wg#Pull-request-merging-criteria) for a description of each category.

1. [ ] MAJOR schema change as defined in [Schema versioning and stability](https://lf-overturemaps.atlassian.net/wiki/x/GgDa).
2. [X] MINOR schema change as defined in [Schema versioning and stability](https://lf-overturemaps.atlassian.net/wiki/x/GgDa).
3. [ ] Cosmetic change.
4. [ ] Documentation change by member.
5. [ ] Documentation change by Overture tech writer.

# Description

The `bbox` attribute should really be described in the schema, this adds it.

Instead of a spatial indexing container, it adds `bbox` to the overture properties container. 

# Reference

# Testing

# Checklist

*Checklist of tasks commonly-associated with schema pull requests. Please review the relevant checklists and ensure you do all the tasks that are required for the change you made.*

1. [ ] Add relevant examples.
2. [ ] Add relevant counterexamples.
3. [ ] Update any counterexamples that became obsolete. For example, if a counterexample uses property `A` but is not intended to test property `A`'s validity, and you made a schema change that invalidates property `A` in that counterexample, fix the counterexample to align it with your schema change.  
4. [ ] Update in-schema documentation using plain English written in complete sentences, if an update is required.
5. [ ] Update Docusaurus documentation, if an update is required.
6. [ ] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation Website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/313)

PR #381, opened 2025-08-12
Add bounding box to model [Updated 8/19]

vcschapp:
# History

1. Initial revision created week of 8/11.
2. Discussion with @mojodna 8/18.
3. Updates made on 8/19:
    - Flattened bounding box to 2D and removed `Dims`.
    - Restored files accidentally deleted ..... For now!
    - Updated Pydantic validation method to support `dict`.

# Description

This change adds a primitive type called `BBox` representing bounding box whose JSON Schema representation is GeoJSON-compatible, and adds a `bbox` field to the Feature model.

## Key Design Decisions

1. **Primitive Type**. Like `Geometry`, `BBox` is modeled as a "primitive" with Pydantic accessories rather than being derived from a Pydantic `BaseModel` itself. It is structured so it is a "basic value type" that will fit seamlessly within a Pydantic model with appropriate GeoJSON-compatible JSON Schema, serde, and Pydantic validation, without itself being a `BaseModel`.
2. **Code Generation Corollary**. A corollary to being designed as a primitive type is that `BBox` will need special case handling for all code generation targets (when we get there) rather than working out of the box with generic code that can handle any `BaseModel`.
5. **Dimensions/Usability**. The `BBox` type is designed to maximize the usability of the 2D case while supporting higher dimensionalities (3D, 4D, *etc.*) if necessary.
    - This is aligned with the GeoJSON spec where 2D is the minimum but an arbitrary number of extra dimensions may be added (see [section 5, *Bounding Box*](https://datatracker.ietf.org/doc/html/rfc7946#section-5)).
    - This is also aligned with how `Geometry` is currently implemented as a Shapely geometry (which is hard-limited to 2.5D, since the underlying Shapely geometry carries a Z-coordinate but doesn't do spatial operations on it) and at the JSON Schema level (since [the current JSON Schema for a GeoJSON position](https://github.com/OvertureMaps/schema/blob/pydantic/packages/overture-schema-core/src/overture/schema/core/geometry.py#L274-L280) just has `"minItems": 2`.
    - It should be noted that the current Overture `bbox` column at the Parquet level only supports the 2D case with `xmin, `ymin`, `xmax`, and `ymax`. We will have to decide whether we want to support additional dimension there or not.
6. **Validation Considerations**. The main validation rules currently enforced are that a `BBox` has to have at least 2 evenly-matched number (`float`/`int`) pairs, so it is made up of an even number of primitive numbers, at least 4.
    - There is currently no validation of the number values of the coordinate position pairs to ensure they remain within bounds.
    - This is moistly aligned with the current implementation of `Geometry` as a wrapper around a Shapely geometry, which actually has no attached CRS and doesn't care what values the numbers have.
    - As we start to build up a code generation system that can generate code to directly validate a Spark dataframe, we can and probably should add validation in at the `Feature` level that verifies that any `BBox`  is in fact the bounding envelope of the feature's geometry. We would add this both in Pydantic itself and in the code generated targets, for example the envisioned Spark in-dataframe validation.
7. **Evolution of Feature Type**. The currently "active" Overture JSON Schema does not model `bbox`, although there was at one time an abortive [attempt to do so](https://github.com/OvertureMaps/schema/commit/5d0ea0dd506b2b3f1e564ea7cd49572eb0ddb067) that was [later reverted](https://github.com/OvertureMaps/schema/commit/163060d056fa479d06221a67c8a87646cd547908), so addition of `bbox` directly into the feature model is an evolution and a step toward bringing all representations of the schema into a cohesive whole.

# Reference

1. https://github.com/OvertureMaps/schema/pull/313
2. https://github.com/OvertureMaps/schema/pull/330

# Testing

Added new unit tests.

### DIFF
--- a/packages/overture-schema-core/src/overture/schema/core/bbox.py
+++ b/packages/overture-schema-core/src/overture/schema/core/bbox.py
@@ -1,0 +1,200 @@
+from dataclasses import dataclass
+from typing import Any
+
+from pydantic import (
+    GetCoreSchemaHandler,
+    GetJsonSchemaHandler,
+    ValidationError,
+    ValidationInfo,
+)
+from pydantic_core import InitErrorDetails, core_schema
+
+
+@dataclass(order=True, frozen=True, slots=True)
+class BBox:
+    """
+    Two-dimensional bounding box.
+
+    Parameters
+    ----------
+    xmin : float | int
+        Minimum X-coordinate
+    ymin : float | int
+        Minimum Y-coordiate
+    xmax : float | int
+        Maximum X-coordinate
+    ymax : float | int
+        Maximum Y-coordinate
+
+
+    Attributes
+    ----------
+    xmin : float | int
+        Minimum X-coordinate
+    ymin : float | int
+        Minimum Y-coordiate
+    xmax : float | int
+        Maximum X-coordinate
+    ymax : float | int
+        Maximum Y-coordinate
+
+
+    Examples
+    --------
+    Create a bounding box:
+
+    >>> bbox = BBox(xmin=-123.5, ymin=37.7, xmax=-122.4, ymax=38.0)
+
+    Serialize a bounding box to its GeoJSON representation:
+
+    >>> BBox(1, 2, 3, 4).to_geo_json()
+    (1, 2, 3, 4)
+
+    Read a bounding box from its GeoJSON representation:
+
+    >>> BBox.from_geo_json([1, 2, 3, 4])
+    BBox(xmin=1, ymin=2, xmax=3, ymax=4)
+
+    Read a bounding box from its GeoJSON representation (simplified form for easier interactive use):
+
+    >>> BBox.from_geo_json(1, 2, 3, 4)
+    BBox(xmin=1, ymin=2, xmax=3, ymax=4)
+    """
+
+    xmin: float | int
+    ymin: float | int
+    xmax: float | int
+    ymax: float | int
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.xmin, float | int):
+            raise TypeError(f"`xmin` must be a `float` or `int`; but {repr(self.xmin)} is a `{type(self.xmin).__name__}`")
+        elif not isinstance(self.ymin, float | int):
+            raise TypeError(f"`ymin` must be a `float` or `int`; but {repr(self.ymin)} is a `{type(self.ymin).__name__}`")
+        elif not isinstance(self.xmax, float | int):
+            raise TypeError(f"`xmax` must be a `float` or `int`; but {repr(self.xmax)} is a `{type(self.xmax).__name__}`")
+        elif not isinstance(self.ymax, float | int):
+            raise TypeError(f"`ymax` must be a `float` or `int`; but {repr(self.ymax)} is a `{type(self.ymax).__name__}`")
+
+    def to_geo_json(self) -> tuple[float | int, ...]:
+        """
+        Return a GeoJSON-compliant `bbox` array.
+
+        Returns
+        -------
+        tuple of float
+            GeoJSON-compliant bounding box coordinates array.
+        """
+        return (self.xmin, self.ymin, self.xmax, self.ymax)
+
+    @classmethod
+    def from_geo_json(cls, *bbox: float | int | tuple[float | int, ...] | list[float | int]) -> "BBox":
+        """
+        Convert a GeoJSON-compliant `bbox` array into a `Bbox` value.
+
+        Parameters
+        ----------
+        bbox : tuple[float | int, ...] | list[float | int]
+            GeoJSON-compliant `bbox` array containing 4 numbers
+
+        Returns
+        -------
+        BBox
+            `BBox` instance corresponding to the input `bbox` array
+
+        Raises
+        ------
+        TypeError
+            If `bbox` is not a `tuple` or `list` or contains non-numeric values
+        ValueError
+            If `bbox` is a `tuple` or `list` and does not have length 4
+        """
+        if len(bbox) == 1 and isinstance(bbox[0], tuple | list):
+            return cls._from_geo_json(bbox[0])
+        else:
+            return cls._from_geo_json(bbox)
+
+    @classmethod
+    def _from_geo_json(cls, bbox: tuple | list) -> "BBox":
+        n: int = len(bbox)
+
+        if not isinstance(bbox, tuple | list):
+            raise TypeError(
+                f"`bbox` must be a `tuple` or `list`; but {repr(bbox)} is a `{type(bbox).__name__}`"
+            )
+        elif n != 4:
+            raise ValueError(
+                f"`bbox` must have length 4; but {repr(bbox)} has length {n}"
+            )
+
+        incompatible_items = [
+            (i, v, type(v).__name__)
+            for i, v in enumerate(bbox)
+            if not isinstance(v, float | int)
+        ]
+        if len(incompatible_items) > 0:
+            incompatible_str = ",".join(
+                f"{repr(v)} (type `{t}` at index {i})" for (i, v, t) in incompatible_items
+            )
+            raise TypeError(
+                f"`bbox` must contain only `float` or `int`; but {repr(bbox)} contains the following incompatible values: {incompatible_str} "
+            )
+
+        return BBox(bbox[0], bbox[1], bbox[2], bbox[3])
+
+
+    @classmethod
+    def __get_pydantic_core_schema__(
+        cls, _source_type: type[Any], _handler: GetCoreSchemaHandler
+    ) -> core_schema.CoreSchema:
+        def validator(value: BBox | tuple[float | int, ...] | list[float | int], info: ValidationInfo) -> "BBox":
+            try:
+                if isinstance(value, BBox):
+                    return value
+                elif isinstance(value, tuple | list):
+                    return cls.from_geo_json(value)
+                elif isinstance(value, dict):
+                    BBox(**value)
+                else:
+                    raise TypeError(
+                        f"expected `BBox` or `tuple` or `list`; got `{type(value).__name__}` with value {repr(value)}"
+                    )
+            except Exception as e:
+                context = info.context or {}
+                loc = context.get("loc_prefix", ()) + ("value",)
+                raise ValidationError.from_exception_data(
+                    title=cls.__name__,
+                    line_errors=[
+                        InitErrorDetails(
+                            type="value_error",
+                            loc=loc,
+                            input=value,
+                            ctx={"error": f"invalid bounding box value: {str(e)}"}
+                        )
+                    ]
+                ) from e
+
+        def serialize_bbox(v: "BBox", info: ValidationInfo | None) -> "tuple[float | int, ...] | 'BBox'":
+            if info and info.mode == "json":
+                return v.to_geo_json()
+            return v
+
+        return core_schema.with_info_plain_validator_function(
+            validator,
+            serialization=core_schema.plain_serializer_function_ser_schema(
+                serialize_bbox, info_arg=True
+            ),
+        )
+
+    @classmethod
+    def __get_pydantic_json_schema__(
+        cls, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> dict[str, Any]:
+        return {
+            'type': 'array',
+            'minItems': 4,
+            'maxItems': 4,  # Expressly limit to subset of GeoJSON bboxes that are 2D.
+            'items': {
+                'type': 'number',
+            }
+        }
--- a/packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
+++ b/packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
@@ -76,6 +76,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Geometry (Point)",
       "properties": {
--- a/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
@@ -95,6 +95,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Geometry (Polygon or MultiPolygon)",
       "oneOf": [
--- a/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
@@ -433,6 +433,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Geometry (Point, LineString, Polygon, or MultiPolygon)",
       "oneOf": [
--- a/packages/overture-schema-base-theme/tests/land_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_baseline_schema.json
@@ -307,6 +307,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Geometry (Point, LineString, Polygon, or MultiPolygon)",
       "oneOf": [
--- a/packages/overture-schema-base-theme/tests/land_cover_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_cover_baseline_schema.json
@@ -112,6 +112,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Geometry (Polygon or MultiPolygon)",
       "oneOf": [
--- a/packages/overture-schema-base-theme/tests/land_use_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_use_baseline_schema.json
@@ -385,6 +385,15 @@
     }
   },
   "properties": {
+    "bbox": {
+      "items": {
+        "type": "number"
+      },
+      "maxItems": 4,
+      "minItems": 4,
+      "title": "Bbox",
+      "type": "array"
+    },
     "geometry": {
       "description": "Classifications of the human use of a section of land. Translates `landuse` from OpenStreetMap tag from OpenStreetMap.",
       "oneOf": [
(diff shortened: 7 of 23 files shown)


packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
sethfitz, line 1: Please restore these. I'm not ready to drop them yet (especially with some changes from Dana that we want to make sure manifest in the right places).

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.min, float | int):
+            raise TypeError(f"`min` must be a `float` or `int`; but {repr(self.min)} is a `{type(self.min).__name__}`")
+        elif not isinstance(self.max, float | int):
+            raise TypeError(f"`max` must be a `float` or `int`; but {repr(self.max)} is a `{type(self.max).__name__}`")
+
+
+
+@dataclass(order=True, frozen=True, slots=True)
```
sethfitz: Nothing else uses data classes here (and the upcoming code generation relies on `BaseModel.model_fields` for detailed introspection, which Pydantic supports more thoroughly than on data classes; as noted in the design decision, BBox would require special handling).

Geometry makes sense not inheriting from `BaseModel` because it's a primitive type that maps to different types in different targets. BBox is structural, even if its representations are either field-based or array-based.

Given that the primary / initial uses are JSON Schema generation (where we need to align with GeoJSON) and code/doc generation (where we align with GeoParquet), I don't see the benefit of the added complexity over a simple `BaseModel` subclass.

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Create a 4D bounding box:
+
+    >>>  bbox = BBox(xmin=0, ymin=0, xmax=1, ymax=1, more=(Dim(0, 1), Dim(-1, 0))
+    """
+
+    xmin: float | int
+    ymin: float | int
+    xmax: float | int
+    ymax: float | int
```
sethfitz, line 67: Why `| int`?

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+    more: tuple[Dim, ...] = ()
```
sethfitz: What's the short/medium-term utility of this? It seems unnecessary.

If we want to support 3D geometries, we should follow [GeoParquet's Bounding Box Columns](https://geoparquet.org/releases/v1.1.0/#Bounding%20Box%20Columns:~:text=by%20this%20method.-,Bounding%20Box%20Columns,-A%20bounding%20box), which include `zmin` / `zmax` (and would have follow-on implications for the data pipeline).

4D geometries are rare (and the `M` dimension can follow different rules); if users have a need for them, they can use 4D WKB geometries (acknowledging Shapely's limitations) with 3D bboxes. (I've never seen a 4D bbox; not to say that they don't exist in the wild, but they seem quite uncommon.)

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+            for v in bbox
+            if not isinstance(v, float | int)
+        ]
+        if len(incompatible_items) > 0:
+            incompatible_str = ",".join(
+                f"{repr(v)} (type {t})" for (v, t) in incompatible_items
+            )
+            raise TypeError(
+                f"`bbox` must contain only `float` or `int`; but {repr(bbox)} contains the following incompatible values:{incompatible_str} "
+            )
```
sethfitz: Pydantic will handle this in `BBox(bbox[0], bbox[1], bbox[mid], bbox[mid+1], more)`, so this is redundant.

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+    @classmethod
+    def __get_pydantic_core_schema__(
+        cls, _source_type: type[Any], _handler: GetCoreSchemaHandler
+    ) -> core_schema.CoreSchema:
+        def validator(value: BBox | tuple[float | int, ...] | list[float | int], info: ValidationInfo) -> "BBox":
+            try:
+                if isinstance(value, BBox):
+                    return value
+                elif isinstance(value, tuple | list):
+                    return cls.from_geo_json(value)
```
sethfitz, line 155: If we're pulling from something derived from Parquet, it'll be a `dict` with 4 keys, and should be parseable directly by `BBox`.

sethfitz (changes requested):

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: These are `float`s (and specifically not `double`s) in our Parquet output.

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
vcschapp: Whoops, they are presumably deleted because once I made my change I got a zillion pages of diff noise from those files, and in an effort to delete them to re-create them maybe more got deleted than wanted.

I can restore them for this PR, but the diff noise is very bad and I want to get rid of them fairly soon. Small feature adds shouldn't produces big diffs in many files.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Because `0` in Python is an `int` and `0.0` is a `float`. (And the same applies to every representable integer value.)

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: I believe the bounding box is a primitive type.

In my view what makes it a primitive type is:

- It is a foundational geospatial primitive where you need "all or nothing" to get something useful.
- It has a fundamentally different representation across data formats.

I don't think the added complexity is particularly great, and it's basically a one-time thing as only geometry and bounding box would get this special treatment.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: > What's the short/medium-term utility of this? It seems unnecessary.

I'm trying to maintain consistency with `Geometry`, including the pre-existing JSON Schema geometry, which explicitly follows the GeoJSON schema and allows for multi-dimensional geometry.

Unfortunately, the GeoJSON schema doesn't assign meaning to any dimensions beyond `x` and `y`, which causes difficulty.

We could, if we wanted, just interpret the third dimension as `z` and drop additional dimensions, or throw on additional dimensions.

> and would have follow-on implications for the data pipeline

Yes, but not dire ones. I mean for now we can just have the code generation omit additional dimensions.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Sure, we can splat the `dict` across the `BBox` constructor.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: It won't unless we replace `@dataclass` with inheriting from `BaseModel`, which is a point of earlier disagreement.

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,235 @@
...
+        if len(incompatible_items) > 0:
+            incompatible_str = ",".join(
+                f"{repr(v)} (type {t})" for (v, t) in incompatible_items
+            )
+            raise TypeError(
+                f"`bbox` must contain only `float` or `int`; but {repr(bbox)} contains the following incompatible values:{incompatible_str} "
+            )
+
+        mid: int = int(n/2)
+        more = tuple(Dim(min, max) for (min, max) in zip(bbox[2:mid], bbox[mid+2:], strict=False))
```
vcschapp: `strict=False` might be an LLM copy/paste artifact, I imagine we want the opposite.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: To expand on this, one key point of commonality between `Geometry` and `BoundingBox` that sets them apart from other types (and why they need support as primitives) is that not only does the physical representation change by data format, but it also change by programming language.

- Shapely has no bounding box type I know of, so we have our own physical representation in Python. But in Java it would wrap a JTS `Envelope`, in Sedona-land it might wrap a Sedona `Envelope`, *etc.*.
- In data format land: GeoParquet differs from GeoJSON. If we wanted to support GPKG it would be different again, *etc.*

With `BaseModel`-derived types, every code generator can mechanically derive the format representation by just traversing the structure tree of the `BaseModel` and applying rules for primitives when it hits the leaf nodes (primitive values). For primitives, you need target-specific rules, and that's what I see as common to `Geometry` and `BoundingBox`.

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
sethfitz: Small feature adds _don't_ produce big diffs unless they're dramatic small features that cross multiple types. The diff noise is probably the diff **context** produced by `deepdiff` (and we may be able to turn that down, or just don't use `-vv` until you need to).

Once I'm confident we have better ways to identify downstream changes, I'll be more confident about getting rid of the generated schemas. (We'll also have them as build artifacts, but that's a bit indirect and awkward.)

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: I have both primitive and target-specific rules in my codegen branch, so I'm able to map `BBox` to `Envelope` (etc.). I'm already doing this for `Geometry`.

All I really want here is to replace the data class with `BaseModel` so that introspection (including for doc generation from the last day, designed to match the data people are looking at, which has (some of) the fields defined here; representing GeoJSON would require a pivot similar to how the JSON Schema is transformed and could handle the same translation) works the same across all models. What's your specific objection to `BaseModel` over `@dataclass`?

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: This is why we may want `strict=False` (which can *also* be configured at parse time, see https://github.com/OvertureMaps/schema/blob/fa176639fec6a8e26a97cfa424393288012eae1c/packages/overture-schema-core/src/overture/schema/core/parser.py#L45 and [`TypeAdapter.validate_python`](https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter.validate_python).

The conceptual type (and in Parquet / readers) is `float`; by declaring that it _might_ be an integer value we create confusion. It shouldn't matter how Python handles it.

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: I don't feel strongly about this, other than following GeoParquet in the field definitions if we're moving past 2 dimensions.

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: See above. `BaseModel` should be equivalent to `@dataclass`, other than including some additional functionality like `model_fields` for introspection of what users will see in the Parquet. This is the only use of `@dataclass` (and I don't see where else we'd use it), so I'd prefer to be consistent and facilitate introspection.

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: We should or we'll run into problems if/when we create a Parquet reader that plugs directly into Pydantic validation for small-scale jobs.

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: Unless we want model/type-specific behavior, we should handle this globally in `parse_feature`. I considered turning it on, mainly to avoid `str` → `number` false positives, but I'm totally happy with `int` → `number` conversions happening (see below).

https://github.com/OvertureMaps/schema/blob/fa176639fec6a8e26a97cfa424393288012eae1c/packages/overture-schema-core/src/overture/schema/core/parser.py#L9-L53

Python:

```console
>>> 4 == 4.0
True
```

Scala:

```console
scala> 4 == 4.0
val res0: Boolean = true
```

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
vcschapp: I want to push back against this pretty strongly.

My personal history is I've included things like this in projects multiple times and multiple times I've had to had someone talk sense in me to remove them.

My problems with them:

1. They are completely derivative, so they can be obtained by just checking out a specific commit and generating them.  So they just noise up the repo.
2. I have only yet experienced false positives from them and every false positive is frustration-inducing. It hasn't actually caught a single bug/issue for me.
3. The testing that we really want to get is better obtained by direct unit testing, or other types of more specific tests.

Note when I say the diff noise, I mean when I make a one-line change to add a feature, it can touch 8 different derivative JSON files, so the actual patch gets bloated.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: But:

```
>>> x = 4
>>> y = 4.0
>>> x is y
False
>>> x == y
True
>>> type(4)
<class 'int'>
>>> type(4.0)
<class 'float'>
```

We should IMO annotate accurately. If we only allow `float` we should assert that it can't be an `int`/raise `TypeError` if it is, but doing that is annoying because then it means your user isn't allowed to write `4` when he means `4.0` anymore.

vcschapp:
Based on a conversation Seth and I had yesterday, I'm going to resolve a number of comments.

In rough terms, we agreed on the following:

- We will drop the `Dims` and reduce `BBox` down to 2 dimensions for now, to keep it aligned with the dimensionality of bounding boxes we are supporting in Overture.
    - We will either raise an exception or drop higher dimensions when converting from GeoJSON. In choosing between these two I'll try to match what Shapely would do if it got too many GeoJSON dimensions.
    - Our stance for the time being is that Overture data has a 2D bounding box and if a user has higher-dimension data they should flatten the envelope to become Overture compliant.
- We will treat `BBox` as a primitive type with the following ramifications:
    - It will have custom logic for each code generation target.
    - We won't over-index on (Geo)Parquet, which is just one format among many, albeit an important one.
    - In the generated reference docs, the type will be `bbox`.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: I'll remove `Dims` - see my longer comment [here](https://github.com/OvertureMaps/schema/pull/381#issuecomment-3201618301).

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Resolving this per [my longer summary comment below](https://github.com/OvertureMaps/schema/pull/381#issuecomment-3201618301).

We'll keep this as a `@dataclass` because it's meant to behave like primitive, not a "struct". I do intend to "upgrade" `Geometry` to `@dataclass` at the next chance I get.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Slated for next revision.

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
vcschapp: FWIW, will restore them in the next revision as it was an accidental delete and we don't need to resolve this question now.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Hmmmm, on further reflection, I don't think the premise is correct.

> We should or we'll run into problems if/when we create a Parquet reader that plugs directly into Pydantic validation for small-scale jobs.

IMO, Parquet should **not** be handled by the Pydantic models, it should live 100% in code generation world. If that's the idea behind the [similar features in `Geometry`](https://github.com/OvertureMaps/schema/blob/pydantic/packages/overture-schema-core/src/overture/schema/core/geometry.py#L211-L216) then they are suspect also.

I'll add it in for the time being, but I'd like to revisit the premise comprehensively at some point.

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
vcschapp: Restored.

packages/overture-schema-core/src/overture/schema/core/bbox.py
vcschapp: Added.

packages/overture-schema-addresses-theme/tests/test_address_json_schema_baseline.py
sethfitz: I have had similar experiences and would also like to get rid of derivatives in the medium-term (my issue is usually drift). However, needing to check out a specific commit and run the tests within that context is too much effort *right now* relative to how commonly I've actually needed to do that. Thank you for restoring it for the time being.

packages/overture-schema-core/src/overture/schema/core/bbox.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.xmin, float | int):
+            raise TypeError(f"`xmin` must be a `float` or `int`; but {repr(self.xmin)} is a `{type(self.xmin).__name__}`")
+        elif not isinstance(self.ymin, float | int):
+            raise TypeError(f"`ymin` must be a `float` or `int`; but {repr(self.ymin)} is a `{type(self.ymin).__name__}`")
+        elif not isinstance(self.xmax, float | int):
+            raise TypeError(f"`xmax` must be a `float` or `int`; but {repr(self.xmax)} is a `{type(self.xmax).__name__}`")
+        elif not isinstance(self.ymax, float | int):
+            raise TypeError(f"`ymax` must be a `float` or `int`; but {repr(self.ymax)} is a `{type(self.ymax).__name__}`")
```
sethfitz, line 77: Another way to deal with this validation (since the internal implementation can be opaque) would be to replace the 4 fields with a typed array (and let Pydantic handle it, although it's not (going to be) a `BaseModel`). However, that seems like it would have the same issue with combinations of floats and ints.

(I still don't understand your perspective on that problem; I get that once Python has them `is` and `==` differ, but I'm unclear on the behavior when they're passed in. However, since this is going to be a `@dataclass`, it doesn't matter because I'm not introspecting on its fields.)

packages/overture-schema-core/src/overture/schema/core/bbox.py
sethfitz: *Parquet* wouldn't be handled (directly) by the Pydantic models. Instead, I'm envisioning some code (:waves hands:) that knows how to read Parquet into Python dicts (flat, not GeoJSON-style) and pass them to `validate_python` without any transformation other than type coercions between Parquet and Python. (So, `geometry` will be `bytes` and `bbox` will be a 4-element dict.)

sethfitz (approved):
I added some comments for posterity, but since `bbox` is going to be an opaque type for declarative purposes, none of it matters for the code and doc generation.

PR #384, opened 2025-08-19
Annotate ID refs with target feature type and relationship type

vcschapp:
# Description

This change adds support for annotating `Id` fields that refer to other features with the target feature type and the relationship type.

Expected use cases:

1. Documentation generation, especially generation of UML and UML-like diagrams.
2. Automation: With these annotations in place, it is now possible to write automated data validation at scale to do things like "verify that the GERS ID reference in this `Id` field actually refers to an existing feature *and* that features has the expected type".

# Testing

New unit tests.


packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py
```diff
@@ -48,6 +50,12 @@
 Width = Annotated[float, Field(gt=0)]
 
 
+def _connector_type() -> type[Feature]:
+    from overture.schema.transportation.connector import Connector
+
+    return Connector
```
vcschapp: This is used to break the circular reference: `ConnectorReference` (`transportation/models.py`) now depends on `Connector`, but `Connector` also depends on `transportation/models.py`.

packages/overture-schema-buildings-theme/src/overture/schema/buildings/building_part/models.py
```diff
@@ -4,9 +4,11 @@
 
 from pydantic import Field
 
+from overture.schema.buildings.building.models import Building
```
sethfitz:
```suggestion
from ..building.models import Building
```

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,34 @@
+from dataclasses import dataclass
+from enum import Enum
+
+from overture.schema.core.models import Feature
```
sethfitz:
```suggestion
from .models import Feature
```

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
```diff
@@ -12,10 +12,12 @@
     Named,
     Names,
 )
+from overture.schema.core.ref import RefersTo, Relationship
 from overture.schema.core.types import CountryCode, Id, RegionCode
 from overture.schema.core.validation import (
     exactly_one_of,
 )
+from overture.schema.divisions.division.models import Division
```
sethfitz:
```suggestion
from ..division import Division
```

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_boundary/models.py
```diff
@@ -19,6 +20,7 @@
     exactly_one_of,
     not_required_if,
 )
+from overture.schema.divisions.division import Division
```
sethfitz:
```suggestion
from ..division import Division
```

packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py
```diff
@@ -48,6 +50,12 @@
 Width = Annotated[float, Field(gt=0)]
 
 
+def _connector_type() -> type[Feature]:
+    from overture.schema.transportation.connector import Connector
```
sethfitz:
```suggestion
    from .connector import Connector
```

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,34 @@
...
+    value, the one that is referred to, as the referee; then this value represents the relationship
+    from the perspective of the referent.
+    """
+
+    def __init__(self, value: str, doc: str) -> None:
+        self._value_ = value
+        self.__doc__ = doc
+
+    BELONGS_TO = "belongs_to", "The referent belongs to the referee"
+    BOUNDS = "bounds", "The referent bounds the referee"
```
sethfitz: Spatially bounded? Type bounded?

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,34 @@
...
+    CONNECTED_TO = "is_connected_to", "The referent is connected to the referee"
```
sethfitz: Has one / has many? Or is this spatial / node / vertex-based?

mojodna:
This feels very ORM-like in terms of mapping foreign keys and describing relationships. When I was noodling on this I discovered that Pydantic's author is also the author of SQLAlchemy, so I expect some design synergy beyond them having been designed to work together.

Since one purpose of extensions is to link together datasets in a structured way, I can imagine that we might want to use existing ORM components to achieve that and somehow to generate DML that creates views over joins. As a result, I'd love to explore how ORM syntax can inspire us.

packages/overture-schema-buildings-theme/src/overture/schema/buildings/building_part/models.py
vcschapp: Yeah that's nicer.

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: It's a good question. The statement as it is kind of vague.

What about "The referent is a ~~spatial~~ geometric boundary of the referee" ?

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: It's topologically based I guess. Any thoughts on a better name and/or description?

vcschapp: `CONNECTS_TO` is parallel to `BELONGS_TO`.

Seth found that in multiple places - RDF, OWL.

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: `BOUNDARY_OF` would be much more clear. (Huddling w Seth now.)

vcschapp: "The referent is a physical boundary of the referee"

packages/overture-schema-core/src/overture/schema/core/ref.py
sethfitz: The string value should be `connectsTo` to follow RDF.

(The realization here is that `RefersTo` encodes triples.)

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
```diff
@@ -70,6 +72,7 @@ class DivisionArea(Feature[Literal["divisions"], Literal["division_area"]], Name
         Field(
             description="Division ID of the division this area belongs to.",
         ),
+        RefersTo(referee=Division, relationship=Relationship.BELONGS_TO)
```
sethfitz: Long term, `@belongsTo(Division)` is both more compact and reads as a sentence. (Where we could register the decorators according to the list of relationships.)

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_boundary/models.py
```diff
@@ -69,7 +71,10 @@ class DivisionBoundary(Feature[Literal["divisions"], Literal["division_boundary"
         ),
     ] = None
     division_ids: Annotated[
-        list[Id],
+        list[Annotated[
+            Id,
+            RefersTo(referee=Division, relationship=Relationship.BOUNDS),
```
vcschapp: Rename to `Reference`. Flip parameters. Could read as:

```python
Reference(Relationship.BELONGS_TO, Division)
```

vcschapp: Suggestion: `BelongsTo[Division]`

Need to have something like:

```python
from typing import TypeVar, Generic

class Relationship(enum):
   FOO = "foo"
   BAR = "baR"

T = TypeVar('T', bound=Feature)

class BelongsTo(Relationship, Generic[T]):
    pass
```

Above is the general gist but there has to be an enum and a class separate.

Or: `Reference(BELONGS_TO, Division` if file scope has alias `BELONGS_TO=Relationshiop.BELONGS_TO`.

packages/overture-schema-core/src/overture/schema/core/ref.py
sethfitz: I think we landed on `isBoundaryOf` (as the RDF-style enum value).

packages/overture-schema-buildings-theme/src/overture/schema/buildings/building_part/models.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_boundary/models.py
vcschapp: I just updated a PR with a change to this syntax:

```
Reference(Relationship.BELONGS_TO, Division)
```

That's most of what we hoped to get, and it's definitely an improvement over the first version.

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area/models.py
vcschapp: One upside of the (updated based on our convo, but only incrementally) syntax `Reference(Relationship.BELONGS_TO, Division)` is that it makes it (IMO) more obvious that you are annotating a primitive value (of type `Id`) to show that it contains a reference, basically that it is a foreign key.

We're kind of conflating two concepts here: the foreign key reference to an ID, and the conceptual relationship that the foreign key is crudely modeling. The downside of the more compact `@belongs_to(Division)` syntax is that it reads like "this `Id` field belongs to division" rather than "this `Id` field is a reference modeling a belongs to relationship to a division"......................

packages/overture-schema-core/src/overture/schema/core/ref.py
vcschapp: Just updated the PR with a commit that applies this change.

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,67 @@
+from dataclasses import dataclass
+from enum import Enum
+
+from overture.schema.core.models import Feature
```
sethfitz:
```suggestion
from .models import Feature
```

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,67 @@
...
+    from the perspective of the relator.
+    """
+
+    def __init__(self, value: str, doc: str) -> None:
+        self._value_ = value
+        self.__doc__ = doc
+
+    BELONGS_TO = "belongs_to", "The referent belongs to the referee"
+    BOUNDARY_OF = "boundary_of", "The referent is a boundary of the referee"
+    CONNECTS_TO = "connects_to", "The referent connects to the referee"
```
sethfitz: Consistency:

```suggestion
    BELONGS_TO = "belongs_to", "The relator belongs to the relatee"
    BOUNDARY_OF = "boundary_of", "The relator is a boundary of the relatee"
    CONNECTS_TO = "connects_to", "The relator connects to the relatee"
```

packages/overture-schema-core/src/overture/schema/core/ref.py
sethfitz: Not sure it got pushed, or if it did, it disappeared somehow.

packages/overture-schema-core/src/overture/schema/core/ref.py
```diff
@@ -0,0 +1,67 @@
...
+
+from overture.schema.core.models import Feature
+
+
+class Relationship(Enum):
+    """
+    Category of relationship between a value that refers to another value.
+
+    If we call the first value, the one that holds the reference, the relator; and the second value,
+    value, the one that is referred to, as the relatee; then this value represents the relationship
```
jenningsanderson, line 12: It's relator / relatee here and then referee / referent below. 

Is it possible to simplify these to the classic "A" and "B", such that the Relationship is "the relationship between A & B" where convention dictates that A is the subject and B is the object (or another subject).

PR #391, opened 2025-09-13
[Pydantic] Doc improvements

mojodna:
READMEs are still a work in progress (I need to tackle the remaining ones).

I feel pretty good about the guide.

### DIFF
--- a/README.pydantic.md
+++ b/README.pydantic.md
@@ -1,10 +1,24 @@
 # Overture Schema
 
-Pydantic schemas for Overture Maps data structures.
+[Pydantic](https://docs.pydantic.dev/latest/) schemas for [Overture Maps data](https://docs.overturemaps.org/guides/).
 
 ## Overview
 
-This project provides type-safe Python models for validating and working with [Overture Maps Foundation](https://overturemaps.org) data. The library uses a multi-package workspace architecture with theme-based namespaces, enabling modular development and extensibility.
+This project provides type-safe Python models for validating and working with [Overture Maps Foundation](https://overturemaps.org/) data. Overture Maps is an open geospatial dataset containing buildings, places, addresses, transportation networks, and administrative boundaries curated from multiple sources.
+
+Use these schemas to:
+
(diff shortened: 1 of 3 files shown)


README.pydantic.md
```diff
@@ -1,168 +1,197 @@
...
-    network_operator: str | None = Field(default=None)
-
-# Register the new type
-register_model("places", "ev_charger", EVCharger)
+# Convert to GeoJSON format
+geojson_output = building.model_dump(mode="json")
 ```
 
-#### Value Expansion
+## Schema Extension
```
stepps00, line 134: In this context, are we considering the sources schema an extension? Might be worth linking to that here as an example.

packages/overture-schema/README.md
```diff
@@ -1,106 +1,93 @@
 # overture-schema
 
-Overture Maps schema collection with implemented themes.
+Type-safe Python models for [Overture Maps Foundation](https://overturemaps.org/) data.
 
-This package provides testing infrastructure for Overture Maps Pydantic schemas. It includes a pytest-based test harness that validates schema models against curated examples and counterexamples.
+This package provides Pydantic models for validating and working with Overture Maps data, including buildings, places, addresses, transportation networks, and administrative boundaries.
```
stepps00, line 5: Add reference to base?

stepps00 (approved):
A couple nits, but this looks great.
Appreciate all the detail in `PYDANTIC_GUIDE.md`.

packages/overture-schema/README.md
sethfitz: I'm struggling to characterize the data that lives in base in a useful way to include here. I figure we'll probably touch this a few times, and "including" does a bunch of work.

README.pydantic.md
sethfitz: I view sources as a separate schema (although it does kind of extend through Feature's `sources`). I left this as a placeholder for when we have our extension story more fully baked.

PR #393, opened 2025-10-01
Fix up overture-schema-sources

mojodna:
- add support for models without a `type` field to the dynamic union generator in `overture.schema:parse`; it now produces `<discriminated union of models with 'type'> | <regular union of models without 'type'>` to maintain performance while adding flexibility for peripheral Overture models
- remove non-existent `PatternPropertiesDictConstraint` in favor of `json_schema_extra=` kwarg
- remove deprecated `format` kwargs (redundant with `HttpUri`)
- avoid transforming non-Features to GeoJSON
- add missing typing information
- align YAML formatting with existing examples
- add missing baseline sources JSON Schema


mojodna:
This was from the wrong origin, so credential fetching doesn't work properly.

PR #397, opened 2025-10-03
Pydantic package organization episode 1: "Making huge mess"

vcschapp:
# tl;dr

This is part #1 only. Part #2 will follow fast.

Yes, it leaves things in a mess (temporarily). This is mainly to produce a manageable diff and let us get some alignment on the destination.

Let's ship this quickly so the process doesn't bog down. The end goal is a full and final organization of the packages and the end of churn.

# Destination

1. **Packages**.
   - ▶️ The package setup will be organized as diagrammed below.
2. **Documentation**.
    - ⬜️ Package-level README content will migrate into the package's `__init__.py` docstring.
    - ⬜️ Our most foundational `BaseModel` base class in `system` will be given an optional `description` property that can be used to provide class-level documentation for the schema user persona. (See [this conversation](https://github.com/OvertureMaps/schema/pull/397#discussion_r2400446749) in the comments on this PR.)
3. **Modules**.

# Packaging Deep Dive

The target layout is as shown below.

2025-10-02-overture-schema-pydantic-architecture-Ideal package arch 9_29 drawio

## overture-schema-system

The `system` package is at the base of the pyramid, and will contain the stuff, otherwise known as "George", that is maximally reusable in any data modeling environment, not just Overture.

-  The name `system` was chosen after some reflection to try to symbolize something "more basic than `core`".
- The previous name candidate, `foundation`, has the unfortunate characteristic of being part of the name Overture Maps Foundation, so was dropped.
- Much of the `validation` package will end up in `system`, but some will also migrate up into `core`.
- Some of the `core` package will migrate down into `system`.

## overture-schema-core

The `core` package contains the shared types that are specific to Overture data. It will continue more or less unchanged, but some of its contents will migrate down to `system`.

## overture-schema-annex

The `annex` package will contain "all the other Overture schemas" part from the main geospatial data. The content from the `sources` package will move in here.

## overture-schema-***-theme

The `theme` packages will remain as-is, subject to:

- Module standardization.
- Documentation standardization.


packages/overture-schema-core/src/overture/schema/core/primitives/__init__.py
```diff
@@ -1,28 +0,0 @@
-"""Primitive data types.
```
vcschapp, line 1: Content moved to the `system` package, see its `__init__.py`.

temp-jail-broken-package/overture-schema-sources/src/overture/__init__.py
```diff
@@ -0,0 +1 @@
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
vcschapp, line 1: I put the `sources` package in jail temporarily b/c it was broken.

In part #2 or part #3 it will come out and be reborn as the `annex` package.

packages/overture-schema-system/src/overture/schema/system/primitive/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    GeometryTypeConstraint,
+)
+
+uint8 = NewType("uint8", Annotated[int, Field(ge=0, le=255)])  # type: ignore [type-arg]
+"""
+Portable 8-bit unsigned integer.
+
+This is an `int` at runtime, but using `uint8` for Pydantic model fields instead of `int` makes them
+portable across different serialization and validation platforms.
+"""
```
vcschapp, line 32: 1. These types need to go in `__init__.py` to place well with `pdoc`. (I spent a looong and unfruitful time evaluating the sorry state of automatic doc generation in Python and landed on `pdoc` as the best hope. Sphinx is a nightmare of anti-automation and other systems have various issues.)
2. Technically this isn't a "real" docstring due to the rejection of [PEP 224](https://peps.python.org/pep-0224/), but [pdoc has special logic to process it anyway](https://pdoc.dev/docs/pdoc.html#document-variables).

packages/overture-schema-system/src/overture/schema/system/string.py
```diff
@@ -0,0 +1,152 @@
...
+    Annotated[
+        str,
+        HexColorConstraint(),
+        Field(
+            description="A color represented as an #RRGGBB or #RGB hexadecimal string, for example #ff0000 for pure red."
+        ),
+    ],
+)  # type: ignore [type-arg]
+HexColor.__doc__ = """
+HexColor : NewType
```
vcschapp, line 28: These will be changed to work the way the numeric ones do, using that `pdoc` feature.

packages/overture-schema-validation/src/overture/schema/validation/types.py
```diff
@@ -4,59 +4,15 @@
...
-    LanguageTagConstraint,
-    NoWhitespaceConstraint,
-    PhoneNumberConstraint,
-    RegionCodeConstraint,
-    WhitespaceConstraint,
-    WikidataConstraint,
 )
 
 # String types with constraints
 PlaceCategory = NewType("PlaceCategory", Annotated[str, CategoryPatternConstraint()])
```
vcschapp, line 15: This will be replaced with `SnakeCaseConstraint`/`SnakeCaseString` in the `system` package.

packages/overture-schema-validation/src/overture/schema/validation/types.py
```diff
@@ -4,59 +4,15 @@
...
-        WikidataConstraint(),
-        Field(
-            description="A wikidata ID if available, as found on https://www.wikidata.org/."
-        ),
-    ],
-)
-
 # Numeric types with constraints
 ConfidenceScore = NewType(
     "ConfidenceScore",
```
vcschapp, line 18: This is going to migrate up into `core`.

packages/overture-schema-addresses-theme/src/overture/schema/addresses/address/models.py
```diff
@@ -8,12 +8,13 @@
...
 )
-from overture.schema.core.geometry import (
+from overture.schema.core.types import CountryCode
+from overture.schema.system.primitive import (
     Geometry,
     GeometryType,
     GeometryTypeConstraint,
 )
-from overture.schema.core.types import CountryCode, TrimmedString
+from overture.schema.system.string import StrippedString
```
vcschapp, line 17: `WhitespaceConstraint`/`TrimmedString` became `StrippedConstraint`/`StrippedString` partly for consistency and partly because the LLMs assure me that `strip` is the Python idiom for trimming whitespace.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
```diff
@@ -0,0 +1,376 @@
...
+            validate_string, python_schema
+        )
+
+
+class PatternConstraint(StringConstraint):
+    """Generic pattern-based string constraint."""
+
+    def __init__(self, pattern: str, error_message: str, flags: int = 0):
+        self.pattern = re.compile(pattern, flags)
+        self.pattern_str = pattern
```
sethfitz, line 39: This looks like it can go away, esp. since it's not actually a string representation.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
```diff
@@ -0,0 +1,376 @@
...
+                        ctx={"error": self.error_message.format(value=value)},
+                    )
+                ],
+            )
+
+    def __get_pydantic_json_schema__(
+        self, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> dict[str, Any]:
+        json_schema = handler(core_schema)
+        json_schema["pattern"] = self.pattern_str
```
sethfitz, line 62:
```suggestion
        json_schema["pattern"] = self.pattern
```

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
sethfitz, line 39:
```suggestion
```

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
```diff
@@ -0,0 +1,376 @@
...
+        self, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> dict[str, Any]:
+        json_schema = handler(core_schema)
+        json_schema["pattern"] = self.pattern.pattern
+        json_schema["description"] = "IETF BCP-47 language tag"
+        return json_schema
+
+
+class CountryCodeConstraint(StringConstraint):
+    """ISO 3166-1 alpha-2 country code constraint."""
```
sethfitz, line 115: We should consider renaming this to reflect that it's an `alpha-2` code (to distinguish from the use of `alpha-3` codes elsewhere).

packages/overture-schema-system/src/overture/schema/system/primitive/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    Geometry,
+    GeometryType,
+    GeometryTypeConstraint,
+)
+
+uint8 = NewType("uint8", Annotated[int, Field(ge=0, le=255)])  # type: ignore [type-arg]
+"""
+Portable 8-bit unsigned integer.
+
+This is an `int` at runtime, but using `uint8` for Pydantic model fields instead of `int` makes them
```
sethfitz, line 30: Depending where this docstring surfaces, what is "runtime"?

packages/overture-schema-system/src/overture/schema/system/string.py
sethfitz: It'd be nice to share the docstring with the Pydantic field description if appropriate (while allowing them to diverge if necessary).

packages/overture-schema-system/src/overture/schema/system/string.py
```diff
@@ -0,0 +1,152 @@
...
+        NoWhitespaceConstraint(),
+        Field(description="A string that contains no whitespace characters"),
+    ],
+)  # type: ignore [type-arg]
+NoWhitespaceString.__doc__ = """
+NoWhitespaceString : NewType
+    A string that contains no whitespace characters.
+"""
+
+
```
sethfitz, line 102: Not sure if this is intentional, but there's a mix of 1 line between types and 2 lines.

packages/overture-schema-system/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+dependencies = [
+    "pydantic>=2.0.0",
+    "shapely>=2.0.0",
+]
+
+[dependency-groups]
+dev = [
+    "pytest>=7.0",
+    "ruff",
+    "mypy",
```
sethfitz, line 21: These should be in the root `pyproject.toml` (as you've done). I'll have a PR to remove references from the other packages up shortly.

Makefile
```diff
@@ -13,18 +13,27 @@ test-all:
 test:
 	@uv run pytest packages/ -x
 
+docformat:
+	@find packages/*/src -name "*.py" -type f -not -name "__*" \
+		| xargs uv run pydocstyle --convention=numpy --add-ignore=D105
```
sethfitz, line 21: Did you look at `docformatter` at all? That's already in as a package dependency (and all of the existing docstrings were formatted with it).

Makefile
```diff
@@ -13,18 +13,27 @@ test-all:
...
-		-p overture.schema.places \
-		-p overture.schema.transportation \
-		-p overture.schema.validation
+	@# $$ escapes $ for make - sed needs literal $ for end-of-line anchor
+	@find packages -maxdepth 1 -type d -name "overture-schema*" \
+		| sort \
+		| sed 's:-theme$$::' \
+		| tr - . \
+		| sed 's:^packages/\|:-p :' \
+		| xargs uv run mypy --no-error-summary
```
sethfitz, line 39: Thank you.

This looks like a side-effect though:

```console
mypy: can't read file 'packages/overture.schema': No such file or directory
```

mojodna (changes requested):

Makefile
```diff
@@ -13,18 +13,27 @@ test-all:
...
-		-p overture.schema.divisions \
-		-p overture.schema.places \
-		-p overture.schema.transportation \
-		-p overture.schema.validation
+	@# $$ escapes $ for make - sed needs literal $ for end-of-line anchor
+	@find packages -maxdepth 1 -type d -name "overture-schema*" \
+		| sort \
+		| sed 's:-theme$$::' \
+		| tr - . \
+		| sed 's:^packages/\|:-p :' \
```
sethfitz: This fixes it on macOS:

```suggestion
		| sed -E 's:^(packages/)?:-p :' \
```

Makefile
vcschapp: Looks like the proximate issue is that BSD `sed` on MacOS doesn't support alternation.

But I think the root cause is that alternation isn't required in this case, so I've removed it. (About to push commit.)

Makefile
vcschapp: Not yet, but it's on my agenda. I'll add that explicitly into the *Destination* section. I think `docformatter` should ideally be run just like `ruff` as part of `make check` assuming it supports a "check for reformats" mode.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
vcschapp: Agree. I'll put this on the agenda.

Are we using alpha-3 anywhere though? I thought I'd managed to keep it consistent...

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
vcschapp: I didn't fully grok this one. Can you elaborate?

packages/overture-schema-system/src/overture/schema/system/string.py
vcschapp: Agreed.

Summary of what we discussed together today:

1. There are two different target audiences, let's call them the "**schema user**" persona and the "**builder/extension author**" persona.
2. The schema user is more interested in higher-level conceptual documentation.
3. The builder/extension author is more interested in Python API reference.
4. In many cases, the way to keep everyone happy is:
   1. Target the docstrings to the builder/extension author.
   2. Target the Pydantic field descriptions to the schema user.
5. There are some edge cases around point 4.
   1. For model classes, there's no way to add a Pydantic field description so the docstring for model classes needs to be aimed at schema users, so higher-level and more conceptual.
   2. For model class *attributes*, Pydantic has a feature called [`use_attribute_docstrings`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.use_attribute_docstrings) that will pull docstrings for attributes in as field descriptions automatically. We want to turn this on. Therefore, docstrings on model class attributes should be higher-level and more conceptual.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
sethfitz: `self.pattern_str` is unnecessary; the lone reference to it (see the next comment) can be changed.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
sethfitz: We're not. I'm thinking about prospective George users who don't follow our specific flavor of consistency.

mojodna (approved):

packages/overture-schema-system/src/overture/schema/system/string.py
vcschapp: Not intentional. I'm basically relying on `ruff format` and `ruff check --fix`.

packages/overture-schema-system/pyproject.toml
vcschapp: Cool. Resolving as I rebased on top of that change.

Makefile
vcschapp: Added into *Destination* section. It won't be the next PR, which is imminent. I'm expecting it to be in episode 4: "Almost There".

packages/overture-schema-system/src/overture/schema/system/primitive/__init__.py
vcschapp: Good question, "while Python is running" is the answer.

This docstring will only surface in the generated Python API reference so I'm OK with it despite the ambiguity/hand-wavy-ness for now. We can improve these though.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
vcschapp: i made this change. It'll be in the upcoming episode 2 PR.

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
vcschapp: Change coming in the next PR, episode 2.

vcschapp: Done in PR #400.

Issue #359, opened 2025-07-12
Documentation for using cateories taxonomy .csv


lilcorey:
https://docs.overturemaps.org/schema/concepts/by-theme/places/#categories

`categories` column in Places does not contain the full taxonomy tree, only `primary` and `alternate`. E.g. `{'primary': 'indian_restaurant', 'alternate': ['restaurant', 'fast_food_restaurant']}`. In order to find all `eat_and_drink` I could build a metadata table from the `.csv` and use it to join `eat_and_drink` to rows containing more granular categories. 
1) can the full taxonomy tree be included in categories column ?
2) alternatively, can examples be added to documentation to show how the full taxonomy can be leveraged in Places query filters?

an-rose:
@lilcorey the `categories` property will be deprecated from the schema in the Sept 2026 release:  https://github.com/OvertureMaps/schema/pull/434.  `basic_category` and `taxonomy` properties were introduced to replace `categories`.  Here's a description of the new properties:  https://docs.overturemaps.org/guides/places/taxonomy/#comparing-the-original-and-new-properties.

Issue #329, opened 2025-03-04
Category definitions


Cj-Malone:
Are the categories defined anywhere? What's the difference between the following?

https://github.com/OvertureMaps/schema/blob/618ba0b7214bc28074b70ef901ef542c82e75dd2/docs/schema/concepts/by-theme/places/overture_categories.csv#L1159-L1160

Cj-Malone:
Also, are the category names translated anywhere?

stepps00:
Overture does not actively maintain category name translations or category definitions today, but both would be helpful for data consumers. Could you tell me a bit more about your use case for both? Would these generally be helpful in understanding the category taxonomy?

As for `cannabis_clinic` and `cannabis_collective` - in the February release, only two places have a primary category of  `cannabis_collective`, but ~9100 place records have a primary category of `cannabis_clinic`. I suspect these may be joined into a single category in the future. When we investigate the taxonomy and propose fixes, we will look into each of the issues you've raised and update here.

Issue #328, opened 2025-03-04
Invalid taxonomy in overture_categories.csv


Cj-Malone:
https://github.com/OvertureMaps/schema/blob/618ba0b7214bc28074b70ef901ef542c82e75dd2/docs/schema/concepts/by-theme/places/overture_categories.csv#L116

There is an inconsistency between the ﻿category code (`serbo_croation_restaurant`) and the taxonomy column (`serbo_croatian_restaurant`), this breaks the tree. `tion` vs `tian`.

`serbo_croation_restaurant` is the one present in the 2025-02-19 data so the quick and easy fix is to update the Overture Taxonomy column.

```patch
 serbo_croation_restaurant; [eat_and_drink,restaurant,serbo_croation_restaurant]
```

However they both probably should be `serbo_croatian_restaurant`, fixing that here is equally as easy, but it doesn’t seem that this is the file used in production. You'd need to fix it somewhere else as well.

stepps00:
Thanks for raising this, @Cj-Malone. There are a handful of other mismatches/fixes we're tracking in the Overture category taxonomy today - in this case, `serbo_croatian_restaurant` should be used. We will investigate and flag this issue once a fix is in place.

jonahadkins:
this was fixed in #343

ericgodwin:
Closing as Jonah referenced a fix.

Issue #39, opened 2023-08-06
Provide category headings for places


bjperson:
Variable names speak for themselves in most cases, but it would still be useful. 

A proposal for a simple, usable format : [categories.json.txt](https://github.com/OvertureMaps/schema/files/12268457/categories.json.txt)

RobSoetewey-TomTom:
Hi,

We're not completely understanding the issue you're bringing. Can you clarify?
There's a Places Category property in the model here: https://github.com/OvertureMaps/schema/blob/main/schema/places/place.yaml#L23-L36

Here are the categories used in overture: https://github.com/OvertureMaps/schema/blob/main/task-force-docs/places/overture_categories.csv

bjperson:
Hi

Right now i’m using :

```
category = category.replace("_", " ")

category = category[0].upper() + category[1:]
```

To get the categories names.

It would be nice to have them directly in github. The icing on the cake would be to have translations as well.

jwass:
Hi @bjperson, I'm not exactly sure if this is what you're looking for but I posted https://github.com/OvertureMaps/schema/issues/40#issuecomment-1774972538 how to find the complete list of categories including alternates with an attached csv.

bjperson:
Hi there,

I was just thinking of something like an "assets" directory in the repository that would contain mappings between category IDs and their human-readable names.

That's the first need I had to display them to my users.

I've almost finished translating it into French, so I can put it somewhere :-)

It could take the form of a "fr.json" file in the format suggested in my first post.

But I don't know whether it should go in the "data" repository or the "schema" repository.

PR #399, opened 2025-10-07
[draft][Places] add basic level category

jonahadkins:

- Adds a new `string` column called `basic_category`.

tbd

# Description

As part of our proposal to improve the current place categorization system in the Places theme, we are introducing a new column called`basic_category`.  

 The basic level category of a place. At present this is a mapping of the `categories.primary` entry to a new, simplified name. This mapping can be 1:1 or M:1 from the existing `categories.primary` entry. If the entry is currently empty in `categories.primary`, this entry will be empty. The full list of basic level categories is available at:(todo)

This new system reduces the current list of ~2100 category names down to ~430 categories.

> This type of categorization is a cognitive science model that is relevant for taxonomy and ontology development that shows the most broadest and general category name that is most often found in the middle of a general-to-specific hierarchy, with generalization proceeding upward and specialization proceeding downward. This category level will be in the hierarchy but can also be pulled out and used for data visualization, matching, grouping similar places for benchmarking

Additionally, a comprehensive analysis of the existing taxonomy and several 3P taxonomies was completed to generate the current list.

# Reference

#359 
#329 
#328 
#39 

# Testing

this is a simple string column addition

TODO.

# Checklist
1. [x] Add relevant examples.

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/399)

### DIFF
--- a/counterexamples/places/bad-basic-category.yaml
+++ b/counterexamples/places/bad-basic-category.yaml
@@ -10,7 +10,7 @@ properties:
   theme: places
   type: place
   version: 1
-  category: woof__
+  basic_category: woof__
   confidence: 0.1
   names:
     primary: bad bad basic category
--- a/counterexamples/places/bad-basic-category.yaml
+++ b/counterexamples/places/bad-basic-category.yaml
@@ -14,3 +14,4 @@ properties:
   confidence: 0.1
   names:
     primary: bad bad basic category
+  operating_status: open 
--- a/counterexamples/places/bad-basic-category.yaml
+++ b/counterexamples/places/bad-basic-category.yaml
@@ -1,0 +1,16 @@
+---
+id: counterexample:place:bad-categories-values
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  ext_expected_errors:
+    - "'woof__' does not match pattern"
+  theme: places
+  type: place
+  version: 1
+  category: woof__
+  confidence: 0.1
+  names:
+    primary: bad bad basic category
--- a/schema/places/defs.yaml
+++ b/schema/places/defs.yaml
@@ -8,9 +8,3 @@ description: Common schema definitions for places theme
       type: string
       minLength: 1
       pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
-    basic_category:
-      description: basic level category of a place.
-      type: string
-      minLength: 1
-      pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
-
--- a/schema/places/defs.yaml
+++ b/schema/places/defs.yaml
@@ -8,3 +8,9 @@ description: Common schema definitions for places theme
       type: string
       minLength: 1
       pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
+    basic_level_category:
+      description: basic level category of a place.
+      type: string
+      minLength: 1
+      pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
+
--- a/schema/places/defs.yaml
+++ b/schema/places/defs.yaml
@@ -8,7 +8,7 @@ description: Common schema definitions for places theme
       type: string
       minLength: 1
       pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
-    basic_level_category:
+    basic_category:
       description: basic level category of a place.
       type: string
       minLength: 1
--- a/schema/places/place.yaml
+++ b/schema/places/place.yaml
@@ -41,6 +41,23 @@ properties:
             items:
               "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
             uniqueItems: true
+      basic_level_category:
+        description: |
+          The Basic-level categorization is a cognitive science model that is relevant
+          for taxonomy and ontology development that shows the most broadest and general
+          category name that is most often found in the middle of a general-to-specific
+          hierarchy, with generalization proceeding upward and specialization
+          proceeding downward. The full list of basic level categories is available at:(todo)
+        type: object
+        required: [basic]
+        properties:
+          basic:
+            description: The basic level category of the place.
+            "$ref": "./defs.yaml#/$defs/typeDefinitions/basic_level_category"
+            type: array
+            items:
+              "$ref": "./defs.yaml#/$defs/typeDefinitions/basic_level_category"
+            uniqueItems: true
       confidence:
         description: The confidence of the existence of the place. It's a number between 0 and 1. 0 means that we're sure that the place doesn't exist (anymore). 1 means that we're sure that the place exists. If there's no value for the confidence, it means that we don't have any confidence information. Places with operating_status set to 'closed' will have a confidence score of 0
         type: number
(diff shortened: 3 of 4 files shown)


schema/places/defs.yaml
```diff
@@ -8,3 +8,9 @@ description: Common schema definitions for places theme
       type: string
       minLength: 1
       pattern: ^[a-z0-9]+(_[a-z0-9]+)*$
+    basic_level_category:
```
vcschapp: My suggestion would be call it `basic_category` and document it as "Basic level category"

schema/places/place.yaml
```diff
@@ -41,6 +41,23 @@ properties:
...
+        description: |
+          The Basic-level categorization is a cognitive science model that is relevant
+          for taxonomy and ontology development that shows the most broadest and general
+          category name that is most often found in the middle of a general-to-specific
+          hierarchy, with generalization proceeding upward and specialization
+          proceeding downward. The full list of basic level categories is available at:(todo)
+        type: object
+        required: [basic]
+        properties:
+          basic:
```
vcschapp: This is introducing multiple levels in the hierarchy. You're going to get something like this in JSON:

```json
{
  "properties": {
    "basic_level_category": {
      "basic": [
        "coffee_shop",
        "pet_shop"
      ]
    }
  }
}
```

And in tabular format you're going to end up with a STRUCT column that contains an ARRAY of STRING.

I'm almost certain that's at least one level of nesting more than you are intending.

---

Questions:

1. Does a Place have one basic category or many?
2. If it has many, are you ending to have structured data for each category, or just a single category ID?

schema/places/place.yaml
```diff
@@ -41,6 +41,23 @@ properties:
             items:
               "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
             uniqueItems: true
+      basic_level_category:
```
vcschapp: This is an array property (at least indirectly), so there are multiple of them. If you are really intending multiple values may I suggest we call this `basic_categories`?

vcschapp (changes requested):
Some comments in the diff.

In addition, would you please:

- Replace all TODO items with actual descriptions, in the PR description and in the diff text.
- Add representative examples and a counterexamples.

schema/places/place.yaml
```diff
@@ -41,6 +41,15 @@ properties:
             items:
               "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
             uniqueItems: true
+      basic_category:
```
vcschapp, line 44: This is a lot simpler than previous and looks a lot closer to what I thought was your end goal. Now it's just a single string. I just want to confirm that that's the intention, that any place regardless of how fancy and weird it is will always just have one `basic_category`.

schema/places/place.yaml
```diff
@@ -41,6 +41,15 @@ properties:
...
+        description: >-
+          The Basic-level categorization is a cognitive science model that is relevant
+          for taxonomy and ontology development that shows the most broadest and general
+          category name that is most often found in the middle of a general-to-specific
+          hierarchy, with generalization proceeding upward and specialization
+          proceeding downward. The full list of basic level categories is available at:(todo)
+        type: string
+        uniqueItems: true
```
vcschapp: `uniqueItems` is only applicable to `type: array` so this should be deleted.

schema/places/place.yaml
```diff
@@ -41,6 +41,15 @@ properties:
...
               "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
             uniqueItems: true
+      basic_category:
+        description: >-
+          The Basic-level categorization is a cognitive science model that is relevant
+          for taxonomy and ontology development that shows the most broadest and general
+          category name that is most often found in the middle of a general-to-specific
+          hierarchy, with generalization proceeding upward and specialization
+          proceeding downward. The full list of basic level categories is available at:(todo)
+        type: string
```
vcschapp, line 51: I think your intention is to reuse the validation rules from `defs.yaml`, right?

In this case, there should be a `"$ref"` out to the `basic_category` item. Here's [an example](https://github.com/OvertureMaps/schema/blob/dev/schema/places/place.yaml#L9-L10) for how these are done:

```yaml
properties:
  id:
    "$ref": ../defs.yaml#/$defs/propertyDefinitions/id
```

counterexamples/places/bad-basic-category.yaml
```diff
@@ -0,0 +1,16 @@
+---
+id: counterexample:place:bad-categories-values
+type: Feature
+geometry:
+  type: Point
+  coordinates: [0, 0]
+properties:
+  ext_expected_errors:
+    - "'woof__' does not match pattern"
```
vcschapp, line 9: This counterexample isn't failing because of the stated reason, it's failing because the required property `operating_status` is missing.

The counterexample check workflow's failing. ...

schema/places/place.yaml
jonahadkins: yes, thats correct 👍

schema/places/place.yaml
jonahadkins: im thinking we can reuse the existing categories one?

schema/places/place.yaml
```diff
@@ -41,6 +41,15 @@ properties:
             items:
               "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
             uniqueItems: true
+      basic_category:
+        description: >-
```
jensgoossens-tomtom, line 45: What's not entirely clear to me is the relationship with 'categories' - which afaics is not required.
Can we have a basic_category when the categories is null?

jonahadkins: neither are required, and when the data is populated it will be from of map of the existing overture primary category to the new basic category.

candy_store -> candy_shop
empty -> empty
sports_and_recreation_venue -> sport_fitness_facility
sports_and_fitness_instruction - > sport_fitness_facility

jensgoossens-tomtom: Shouldn't we add that to the description to make that clear, or express this somehow in the schema that they are either both null or both filled in?

vcschapp (approved):
image

*a.k.a.* SHIP IT!

schema/places/place.yaml
vcschapp: Seems to work!

schema/places/place.yaml
vcschapp: Correct me if I'm wrong, but I understood the goal is to deprecate and remove the old `categories` column. Is that accurate?

jonahadkins: i wasnt sure if we are deprecating this release or in a future one

jwass (approved):

schema/places/place.yaml
vcschapp: Project plan from [2025-10-01 meeting agenda](https://github.com/OvertureMaps/schema-wg/issues/384):

> WIP: October 2025, adding basic category column; November 2025, no schema changes; December 2025, adding new Overture categories property; March 2026, deleting old Overture categories property

It sounds like we'll put a `deprecated` marking in, in December, with final removal is planned for March. Ish.

PR #400, opened 2025-10-09
Pydantic package organization episode 2: "The mess unfolds"

vcschapp:
# tl;dr

This is part #2. Part #3 should drop by Friday 8/10 latest. I'm now expecting 5 total parts.

# Squash Message

```
Re-org. Pydantic packages (2/N)

The destination is as described in PR #397 (1/N) and PR #400. In this
commit, the following value is added:

1. 🐞 Fix bug: The `@any_of` decorator was not working with transporation
   destinations. The bug was discovered while refactoring the `mixin`
   module and the fix is reflected in the updated transportation schema
   baseline.
2. 🐞 Fix bug: The `make doctest` target exited with code 0 (success)
   even if it found doctest failures. This bug was added in #397 when
   the `doctest` target was added and is now fixed.
3. Apply multiple suggestions from PR #397 including renaming
   `CountryCode` to `CountryCodeAlpha2` and dropping an unnecessary
   attribute from `PatternConstraint`.
4. Rename the sources package to annex in alignment as shown in the
   destination package diagram.
5. Begin refactoring `mixin` module to fix some known issues and polish
   it for final release. Changes made here:
      - Underlying implementation is replaced with the Pydantic function
        `create_model` (as was done in the scoping POC) which eliminates
        the need for `ConstrainedBaseModel` and should also eliminate
        MRO issues.
      - Underlying implementation uses some basic "JSON Schema algebra"
        to ensure that multiple decorators can coexist without breaking
        each others' JSON Schema changes.
      - The module name `mixin` is replaced by `model_constraint`, which
        feels more use case-oriented.
      - Initially one main decorator is migrated from `mixin` in the
        validation package to `model_constraint` in the system package.
        The rest will come next PR.
6. Migrate much of the validation package `README.md` into the system
   package's main module `__init__.py` and migrate the examples into
   doctests.
7. Re-home a few other bits of functionality from validation to core.
8. Bring the https://github.com/vcschapp/overture-schema-pydantic
   scoping into core as `scoping.py` and fix up the linter errors. This
   is not wired up yet, but that's Coming Soon, likely in episode 4.

Stay tuned for episode 3, where we will finish re-homing the validation
package.
```

# Journey

- Part 1 is in PR #397.

# Destination

1. **Packages**.
   - ▶️ The package setup will be organized as diagrammed below.
2. **Documentation**.
    - ⬜️ ~~Our most foundational `BaseModel` base class in `system` will be given an optional `description` property that can be used to provide class-level documentation for the schema user persona. (See [this conversation](https://github.com/OvertureMaps/schema/pull/397#discussion_r2400446749) in the comments on this PR.)~~ I no longer think this is necessary as there's an evolving clean separation between fields and models (user-facing descriptions or docstrings) and system types (builder-facing docstrings).
3. **Modules**.

# Packaging Deep Dive

Recapping the target package organization from part 1 (PR #397):

2025-10-02-overture-schema-pydantic-architecture-Ideal package arch 9_29 drawio

### DIFF
--- a/packages/overture-schema-core/src/overture/schema/core/types.py
+++ b/packages/overture-schema-core/src/overture/schema/core/types.py
@@ -17,7 +17,7 @@ from overture.schema.system.field_constraint import (
 from overture.schema.system.field_constraint.string import (
     CountryCodeAlpha2Constraint,
 )
-from overture.schema.system.primitive import float64, int32, pct
+from overture.schema.system.primitive import float32, int32, pct
 from overture.schema.system.string import (
     HexColor,
     JsonPointer,


packages/overture-schema-addresses-theme/src/overture/schema/addresses/address/models.py
```diff
@@ -66,7 +67,7 @@ class Address(Feature[Literal["addresses"], Literal["address"]]):
                 Note: when a level is not known - most likely because the data provider has not supplied it and we have not derived it from another source, the array element container must be present, but the "value" field should be omitted""",
         ),
     ] = None
-    country: CountryCode | None = None
+    country: CountryCodeAlpha2 | None = None
```
vcschapp, line 70: Implementing a change [requested in the previous PR](https://github.com/OvertureMaps/schema/pull/397#discussion_r2402696378).

packages/overture-schema-system/src/overture/schema/system/constraint/string.py
```diff
@@ -36,7 +38,6 @@ class PatternConstraint(StringConstraint):
 
     def __init__(self, pattern: str, error_message: str, flags: int = 0):
         self.pattern = re.compile(pattern, flags)
-        self.pattern_str = pattern
```
vcschapp, line 39: Implementing a change [requested in the previous PR](https://github.com/OvertureMaps/schema/pull/397#discussion_r2402691161).

packages/overture-schema-system/src/overture/schema/system/model_constraint/__init__.py
```diff
@@ -0,0 +1,12 @@
+from .model_constraint import ModelConstraint, apply_alias
```
vcschapp, line 1: The rest of the mixins will migrate here in the next PR in the series.

packages/overture-schema-system/src/overture/schema/system/model_constraint/json_schema.py
```diff
@@ -0,0 +1,110 @@
+from collections.abc import Callable
```
vcschapp, line 1: Internal module used to do "JSON Schema algebra" so the mixin constraints can write JSON Schemas that don't stomp on each other if they are both decorating the same base model.

packages/overture-schema-core/src/overture/schema/core/scoping.py
```diff
@@ -0,0 +1,208 @@
+from collections.abc import (
```
vcschapp, line 1: First version is a straight port from the POC with minor amendments to pass `ruff` and `mypy`, and integrate `@require_any_of`. Will develop it further in an upcoming PR, planned to be episode 4/N.

packages/overture-schema-core/src/overture/schema/core/types.py
```diff
@@ -35,6 +40,38 @@
     ],
 )
 
+
+class ConfidenceScoreConstraint(FieldConstraint):
```
vcschapp, line 44: Re-homed from validation package.

packages/overture-schema-core/src/overture/schema/core/types.py
```diff
@@ -47,6 +84,78 @@
         ),
     ],
 )
+
+
+class LinearReferenceRangeConstraint(CollectionConstraint):
```
vcschapp, line 89: Re-homed from validation package.

packages/overture-schema-system/src/overture/schema/system/model_constraint/model_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+        Raises
+        ------
+        RuntimeError
+            If the config is in an unacceptable state
+        """
+        pass  # noqa: B027
+
+    @final
+    @classmethod
+    def get_model_constraints(
```
vcschapp, line 180: This is the introspection interface the codegen can call to find all the `ModelConstraint` values that were attached to the base model.

packages/overture-schema-core/src/overture/schema/core/models.py
```diff
@@ -333,7 +335,9 @@ class CartographicallyHinted(BaseModel):
     cartography: Annotated[CartographicHints | None, Field(title="cartography")] = None
 
 
-class Address(StrictBaseModel):
+# TODO - vic - move this into Places
```
sethfitz, line 338: 👍 

FWIW, it's only here because this matches where it was located in the JSON Schema.

packages/overture-schema-core/src/overture/schema/core/ext.py
```diff
@@ -0,0 +1,62 @@
...
+
+def allow_extension_fields() -> Callable:
+    """Decorator to allow only ext_* prefixed extension fields."""
+
+    def decorator(cls: type[BaseModel]) -> type[BaseModel]:
+        constraint = ExtensionPrefixValidator()
+        register_constraint(cls, constraint)
+        return cls
+
+    return decorator
```
sethfitz, line 20: Very nice.

packages/overture-schema-core/src/overture/schema/core/scoping.py
sethfitz: Should I come back and review this once it's wired up? I haven't spent the time to properly grok it, but I suspect that I have questions that are better answered by comments, how it's used by models.

packages/overture-schema-core/src/overture/schema/core/types.py
```diff
@@ -35,6 +40,38 @@
...
+        json_schema["minimum"] = 0.0
+        json_schema["maximum"] = 1.0
+        json_schema["description"] = "Confidence score between 0.0 and 1.0"
+        return json_schema
+
+
+ConfidenceScore = NewType(
+    "ConfidenceScore",
+    Annotated[
+        float64,
```
sethfitz:
```suggestion
        float32,
```

`double` precision is probably unnecessary here.

Same for `LinearlyReferencedPosition`, if that was previously a `float64`.

However, we need to consult the Parquet types for both and match them, otherwise we have an unexpected breaking change.

packages/overture-schema-system/src/overture/schema/system/__init__.py
```diff
@@ -1,21 +1,119 @@
...
+  decorated onto Pydantic model classes to add cross-field validation rules, for example "these two
+  fields are mutually-exclusive" or "if this field is set, then that field must also be set".
+- :mod:`string <overture.schema.system.string>` String types with built-in validation to conform to
+  well-known patterns, for example a country code, a hexadecimal color code, a language tag, or just
+  a string that doesn't contain whitespace.
 
 
+Features
+--------
+- Integrates with Overture code generation tools, making Pydantic models build using these types
```
sethfitz:
```suggestion
- Integrates with Overture code generation tools, making Pydantic models built using these types
```

packages/overture-schema-system/src/overture/schema/system/__init__.py
```diff
@@ -1,21 +1,119 @@
...
+  portable across different programming languages (*e.g.*, Java) and serialization formats (*e.g.*,
+  Parquet.)
+- Tightly integrated with Pydantic's JSON Schema system, providing rich JSON Schemas and maximum
+  parity between Pydantic, generated JSON Schemas, and Overture's code generation tools.
+- First-class support for geospatial data using the geometry primitives.
+- Conditional fields and validation on relationships between fields (*e.g.* if the type field
```
sethfitz:
```suggestion
- Conditional fields and validation on relationships between fields (*e.g.*, if the type field
```

packages/overture-schema-system/src/overture/schema/system/string.py
```diff
@@ -24,18 +49,16 @@
...
+"""
+A color represented as an #RRGGBB or #RGB hexadecimal string.
 
-    For example:
+For example:
 
-    - "#ff0000" for pure red 🟥
-    - "#ffa500" for bright orange 🟧
-    - "#000000" or "#000" for black ⬛
+- `"#ff0000"` for pure red 🟥
```
sethfitz:
```suggestion
- `"#ff0000"` or `#f00` for pure red 🟥
```

If we're being consistent ;-)

packages/overture-schema-system/src/overture/schema/system/model_constraint/model_constraint.py
sethfitz: Thank you!

packages/overture-schema-system/tests/model_constraint/test_no_extra_fields.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+from pydantic import BaseModel, ConfigDict, create_model
+
+from overture.schema.system.model_constraint import (
+    NoExtraFieldsConstraint,
+    no_extra_fields,
+)
+
+sys.path.insert(0, str(Path(__file__).parent))  # Needed to import `util` peer module.
```
sethfitz, line 12: This may be fixable within `pyproject.toml`, setting up `tests/` to behave as modules... Did you try anything else?

packages/overture-schema-validation/tests/test_mixin_constraints.py
```diff
@@ -915,57 +713,6 @@ class FeatureModel(BaseModel):
...
-                is_variable=True,
-                conditions=None,
-            )
-        assert "Field 'conditions' is required when is_variable = True" in str(
-            exc_info.value
-        )
-
 
 if __name__ == "__main__":
     pytest.main([__file__])
```
sethfitz, line 718: Whoops. This is probably extraneous too.

packages/overture-schema-core/src/overture/schema/core/types.py
vcschapp: Yeah.

Agreed on the Parquet model. I'm sold on `float32` for confidence score. I think we might need to use `float64` for LR though - one of the downsides of the percentage model we're using is that the real available precision (cm, mm) varies based on the length of the segment - and having more mantissa bits can help with this. Frankly if we had a standard fixed-point number it would be better.

packages/overture-schema-system/tests/model_constraint/test_no_extra_fields.py
vcschapp: No. I took the lazy man's way out. (Well lazy and time-constrained.)

We should fix it for sure. But being a test I'm thinking it's a bit farther down for now.

packages/overture-schema-core/src/overture/schema/core/scoping.py
vcschapp: I think it's worth coming back to review this.

In theory, it will result in a large net loss of code because a lot of explicit modeling will be replaced with `@scoping.whatever` decorations, but it might be easier to grok what it does once the decorations are tied to JSON Schema output and unit tests.

PR #403, opened 2025-10-16
Pydantic package organization episode 3: "System update complete"

vcschapp:
# tl;dr

This is part `#3`. Part `#4` is expected early in the week of 10/20 if all goes according to plan.

At this point the `system` package (George) is pretty much finished, documented, and tested, and there's nothing more that can obviously move into it. It's my hope that the majority of the refactoring is done now. Part `#4` will focus on organizing the `core` package.

# Squash Message

```
The destination is as described in PR #397 (1/N), PR #400, and PR #403.
In this commit, the following value is added:

1. 🐞 Fix bug: The `not_required_if` mixin wasn't meeting the original
   schema intent, which result in the `DivisionBoundary` feature type
   having the wrong JSON Schema. The expected behavior is "if the
   boundary is at the country level, then the `country` pair should be
   omitted" but instead it did two variations: in the Python validator,
   it required the country pair to be present; and in the JSON Schema,
   it required the country pair to be present, but only if the condition
   was not true. Both were subtly wrong. The bug is fixed in this commit
   and the mixin is renamed to `forbid_if`, which I hope will make the
   semantics more obvious.
2. 🐞 Fix bug: Very subtle, but the mixin model constraint validators
   did not work correctly if added directly to a `Feature` model,
   because they aren't aware of the quirky multi-tiered GeoJSON JSON
   Schema where most properties get pushed down into "properties", but
   some do not. This is (hopefully ... 🤞) fixed in the new `Feature`
   model in the system package.
3. 🐞 Fix bug: Very subtle, but the previous `Feature` model's JSON
   Schema did not prohibit fields named `"id"`, `"bbox"`, and
   `"geometry"` within the properties object. This created a weird
   quirk where in a poisoned JSON object containing those values under
   properties, the properties version would overwrite the root version. 
   This is fixed by explicitly prohibiting those fields in the JSON
   Schema for the `Feature` class' properties object.
4. 🐞 Fix bug: The `ExtensibleBaseModel`'s `patternProperties` got put
   both at the top level of the GeoJSON Feature schema and in the
   "properties" object schema. This would have allowed the input JSON
   to contain `ext_*` at two levels, which doesn't meet the original
   schema intent. As part of this bug fix, `ExtensibleBaseModel` is
   collapsed directly into `OvertureFeature` and the `ext_*` feature
   is documented as being on a deprecation path.
5. The `validation` package is officially deleted as its essential
   George functions have all been re-homed either into `system` or
   `core`.
6. All model constraint mixins have been refactored along the lines
   started in PR #400. They are now called `@forbid_if`,
   `@min_fields_set`, `@no_extra_fields`, `@radio_group`,
   `@require_any_of`, and `@require_if`.
7. A type hint, `Omitable[T]` has been added to the system package.
   This type hint allows a Pydantic model to "naturally" get JSON
   Schema semantics for omitted values - it de-conflates optionality
   and nullability, which Pydantic currently conflates. In an
   upcoming episode, I hope to replace the `X | None` unions in the
   existing model with `Omitable[X]`.
8. The existing `Feature` class from the `core` package has been split
   into two: (i) `Feature` in the `system` package now acts as a
   generic feature and owns all the funky GeoJSON logic, allowing
   George enthusiasts to work with a nice ergonomic feature type without
   having to use an "Overture"-flavored feature; (ii) `OvertureFeature`
   in the `core` package adds the Overture schema specific flavoring
   on top of `Feature`.
9. The `ref` module from `core` has been pushed down into `system` and
   the dependency on `(Overture)Feature` has been factored out in favor
   of a generic dependency on an `Identified` model, which is basically
   just a `BaseModel` with a mandatory ID.

Stay tuned for episode 4, where we will re-organize the core package.
```

# Journey

- Part 1 is in PR #397.
- Part 2 is in PR #400.

# Destination

1. **Packages**.
2. **Documentation**.
3. **Modules**.

### DIFF
--- a/packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
+++ b/packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
@@ -77,6 +77,7 @@
   "description": "Addresses are geographic points used for locating businesses and individuals. The\nrules, fields, and fieldnames of an address can vary extensively between locations.\nWe use a simplified schema to capture worldwide address points.  This initial schema\nis largely based on the OpenAddresses (www.openaddresses.io) project.\n\nThe address schema allows up to 5 \"admin levels\". Rather than have field names that\napply across all countries, we provide an array called \"address_levels\" containing\nthe necessary administrative levels for an address.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
+++ b/packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
@@ -128,6 +128,7 @@
   "description": "Boundaries represent borders between divisions of the same subtype.\n\nSome boundaries may be disputed by the divisions on one or both sides.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/bathymetry_baseline_schema.json
@@ -96,6 +96,7 @@
   "description": "Topographic representation of an underwater area, such as a part of the ocean\nfloor.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
@@ -434,6 +434,7 @@
   "description": "Various features from OpenStreetMap such as bridges, airport runways, aerialways,\nor communication towers and lines.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/land_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_baseline_schema.json
@@ -308,6 +308,7 @@
   "description": "Physical representations of land surfaces.\n\nGlobal land derived from the inverse of OSM Coastlines. Translates `natural` tags from OpenStreetMap.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/land_cover_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_cover_baseline_schema.json
@@ -113,6 +113,7 @@
   "description": "Representation of the Earth's natural surfaces.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/land_use_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/land_use_baseline_schema.json
@@ -386,6 +386,7 @@
   "description": "Land use features from OpenStreetMap.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-base-theme/tests/water_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/water_baseline_schema.json
@@ -269,6 +269,7 @@
   "description": "Physical representations of inland and ocean marine surfaces.\n\nTranslates `natural` and `waterway` tags from OpenStreetMap.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-buildings-theme/tests/building_baseline_schema.json
+++ b/packages/overture-schema-buildings-theme/tests/building_baseline_schema.json
@@ -391,6 +391,7 @@
   "description": "A building is a man-made structure with a roof that exists permanently in one\nplace.\n\nBuildings are compatible with GeoJSON Polygon features.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-buildings-theme/tests/building_part_baseline_schema.json
+++ b/packages/overture-schema-buildings-theme/tests/building_part_baseline_schema.json
@@ -277,6 +277,7 @@
   "description": "A single building part.\n\nParts describe their shape and color and other properties. Each building part must\nrefer to the building to which it belongs.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-core/tests/test_models.py
+++ b/packages/overture-schema-core/tests/test_models.py
@@ -448,7 +448,6 @@ def test_feature_json_schema() -> None:
             {
                 "type": "Feature",
                 "id": "foo",
-                "bbox": None,
                 "geometry": {
                     "type": "Point",
                     "coordinates": [-1, 1],
--- a/packages/overture-schema-divisions-theme/tests/division_area_baseline_schema.json
+++ b/packages/overture-schema-divisions-theme/tests/division_area_baseline_schema.json
@@ -236,6 +236,7 @@
   "description": "Division areas are polygons that represent the land or maritime area covered by a\ndivision.\n\nEach division area belongs to a division which it references by ID, and for which\nthe division area provides an area polygon. For ease of use, every division area\nrepeats the subtype, names, country, and region properties of the division it\nbelongs to.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-divisions-theme/tests/division_baseline_schema.json
+++ b/packages/overture-schema-divisions-theme/tests/division_baseline_schema.json
@@ -337,6 +337,7 @@
   "description": "Divisions are recognized official or non-official organizations of people as seen\nfrom a given political perspective.\n\nExamples include countries, provinces, cities, towns, neighborhoods, etc.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-places-theme/tests/place_baseline_schema.json
+++ b/packages/overture-schema-places-theme/tests/place_baseline_schema.json
@@ -300,6 +300,7 @@
   "description": "A Place is a point representation of a real-world facility, service, or amenity.\n\nPlace features are compatible with GeoJSON Point features.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-transportation-theme/tests/connector_baseline_schema.json
+++ b/packages/overture-schema-transportation-theme/tests/connector_baseline_schema.json
@@ -62,6 +62,7 @@
   "description": "Connectors create physical connections between segments.\n\nConnectors are compatible with GeoJSON Point features.",
   "properties": {
     "bbox": {
+      "description": "An optional bounding box for the feature",
       "items": {
         "type": "number"
       },
--- a/packages/overture-schema-transportation-theme/tests/segment_baseline_schema.json
+++ b/packages/overture-schema-transportation-theme/tests/segment_baseline_schema.json
@@ -622,6 +622,7 @@
       "description": "Rail Segment Properties.",
       "properties": {
         "bbox": {
+          "description": "An optional bounding box for the feature",
           "items": {
             "type": "number"
           },
@@ -890,6 +891,7 @@
       "description": "Road Segment Properties.",
       "properties": {
         "bbox": {
+          "description": "An optional bounding box for the feature",
           "items": {
             "type": "number"
           },
@@ -1559,6 +1561,7 @@
       "description": "Water Segment Properties.",
       "properties": {
         "bbox": {
+          "description": "An optional bounding box for the feature",
           "items": {
             "type": "number"
           },


packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
```diff
@@ -75,13 +75,9 @@
   },
   "additionalProperties": false,
   "description": "Addresses are geographic points used for locating businesses and individuals. The\nrules, fields, and fieldnames of an address can vary extensively between locations.\nWe use a simplified schema to capture worldwide address points.  This initial schema\nis largely based on the OpenAddresses (www.openaddresses.io) project.\n\nThe address schema allows up to 5 \"admin levels\". Rather than have field names that\napply across all countries, we provide an array called \"address_levels\" containing\nthe necessary administrative levels for an address.",
-  "patternProperties": {
-    "^ext_.*$": {
-      "description": "Additional top-level properties must be prefixed with `ext_`."
-    }
-  },
```
vcschapp, line 82: This is one of the subtle 🐞 bugs addressed: `patternProperties` shouldn't exist at the top level because that would allow extra properties to be put into the root of the GeoJSON document, like this:

```json
{
  "type": "Feature",
  "geometry": {},
  "properties": {},
  "ext_foo": "bar",
  "ext_bar": "baz"
}
```

The extension properties would then get lost by the model validator that imports the GeoJSON into the Pydantic model class instance.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
```diff
@@ -129,9 +125,16 @@
     },
     "properties": {
       "additionalProperties": false,
+      "not": {
```
vcschapp, line 128: This fixes another subtle 🐞 bug: if these properties were allowed in under "properties", then you'd be open to this scenario in a GeoJSON input:

```json
{
  "type": "Feature",
  "id": "outer-id",
  "bbox": [0, 0, 0, 0],
  "geometry": {},
  "properties": {
    "id": "inner-id",
    "bbox": [1, 1, 1, 1],
    "geometry": "anything"
  }
}
```

The model validator would then copy everything from "properties" up to the root (overwriting the values in the root).

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
```diff
@@ -221,19 +224,18 @@
         "type",
         "version"
       ],
-      "type": "object",
-      "unevaluatedProperties": false
```
vcschapp, line 225: Redundant - because `"additionalProperties": false` is stricter...

packages/overture-schema-base-theme/src/overture/schema/base/bathymetry/models.py
```diff
@@ -6,7 +6,7 @@
 
 from overture.schema.base.types import Depth
 from overture.schema.core import (
-    Feature,
+    OvertureFeature,
```
vcschapp, line 9: This change will become more obvious later on, but I split old `Feature` in two:

1. new `Feature` in the `system` package: the basic GeoJSON-enabled feature model.
2. `OvertureFeature` in the `core` package: same as old `Feature` but with the basic GeoJSON stuff factored out into the base `Feature`.

Full circle, I know. I went from saying we shouldn't call it `OvertureFeature` to saying we should. We could go with just `Feature`, but I thought it might be good to disambiguate the names.

packages/overture-schema-core/src/overture/schema/core/models.py
```diff
@@ -144,100 +136,48 @@ class Feature(ExtensibleBaseModel, Generic[ThemeT, TypeT], ABC):
...
-                "id": data.pop("id"),
-                **({"bbox": data.pop("bbox")} if "bbox" in data else {}),
-                "geometry": data.pop("geometry"),
-                "properties": data,  # All remaining fields go into properties
-            }
-        else:
-            # Return flattened structure for Python output (info.mode == "python")
-            return data
+    @model_validator(mode="after")
+    def validate_model(self) -> Self:
```
vcschapp, line 142: This is just applying the `ext_*` validation and can be dropped when we drop that.

packages/overture-schema-core/src/overture/schema/core/models.py
```diff
@@ -144,100 +136,48 @@ class Feature(ExtensibleBaseModel, Generic[ThemeT, TypeT], ABC):
...
-            json_schema_top_level_required.append("properties")
-
-        # Add the `"type": "Feature"` GeoJSON property at the top level
-        json_schema_top_level_properties["type"] = {
-            "type": "string",
-            "const": "Feature",
+        cls,
+        schema: core_schema.CoreSchema,
+        handler: GetJsonSchemaHandler,
+    ) -> JsonSchemaValue:
```
vcschapp, line 162: This is just applying the `ext_*` schema and can be dropped when we drop that.

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
```diff
@@ -240,8 +236,31 @@
...
+              }
+            }
+          },
+          "then": {
+            "not": {
+              "required": [
+                "country"
+              ]
+            }
+          }
```
vcschapp, line 254: This is another bug 🐞 fix.

The old decoration was:

```
@not_required_if("subtype", PlaceType.COUNTRY, ["country"])
```

This only captured half the [original constraint](https://github.com/OvertureMaps/schema/blob/dev/schema/divisions/division_boundary.yaml#L22-L29) (`if-then` but not `if-then-else`).

packages/overture-schema-system/src/overture/schema/system/metadata.py
```diff
@@ -0,0 +1,319 @@
+from collections.abc import (
```
vcschapp, line 1: The metadata is a formalization and generalization of the `model_constraint` metadata introduced in #400 track the model constraints applied on a model for code generation support.

packages/overture-schema-system/src/overture/schema/system/create_model.py
```diff
@@ -0,0 +1,56 @@
...
+    /,
+    *,
+    __config__: pydantic.ConfigDict | None = None,
+    __doc__: str | None = None,
+    __base__: type[ModelT] | tuple[type[ModelT], ...] | None = None,
+    __module__: str | None = None,
+    __validators__: dict[str, Callable[..., Any]] | None = None,
+    __cls_kwargs__: dict[str, Any] | None = None,
+    __qualname__: str | None = None,
+    __metadata__: Metadata | None = None,
```
vcschapp, line 22: #It's the addition of this line that makes our `create_model` different from Pydantic's.

packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
```diff
@@ -0,0 +1,146 @@
+from collections.abc import Callable
```
vcschapp, line 1: I'm expecting one more pass may be necessary to make `@forbid_if`, `@radio_group`, `@require_any_of`, and `@require_if` aware of `Omitable` and behave differently if the field is omitable or nullable (nullable being a union with `None`).

packages/overture-schema-system/src/overture/schema/system/feature.py
vcschapp, line 1: This is the factored-out GeoJSON-enabled feature model.

Everything line 388 and below is in support of "on-the-fly refactoring" the JSON Schema to support `model_constraint` decorators to support three cases:

1. Constraints that refer only to the top-level fields `"id", `"bbox"`. and `"geometry"`. 
2. Constraints that refer only to non-top-level fields.
3. Constraints that refer to a mix of top-level and non-top-level fields.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
sethfitz: Mmm, so it wasn't getting pivoted into `properties` correctly. Good catch.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
sethfitz: What do you mean by "stricter"?

IIRC, `unevaluatedProperties` applies to sub-schemas (and structures introduced through `allOf`, etc.), which is probably what we want at the `OvertureFeature` level (to prevent people from accidentally making their schemas more open than intended, as is the case in a few parts of our current JSON Schema).

packages/overture-schema-base-theme/src/overture/schema/base/bathymetry/models.py
sethfitz: Nah, this makes sense. The split is good.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
sethfitz: Does `not.required` achieve this? It seems more like it would say that those fields aren't required, which was already the case. `additionalProperties: false` will prevent it, but is contingent on the shape of `properties` **filtering out** references to `id`, `bbox`, `geometry`.

packages/overture-schema-core/src/overture/schema/core/types.py
```diff
@@ -281,6 +252,8 @@ def __get_pydantic_json_schema__(
     ],
 )
 
+# FIXME: Use of `default` on this "floating" type declaration results in a Pydantic warning that the
+#        default has no effect. Default value should be migrated to site usage in the actual models.
```
sethfitz, line 256: Or removed, since Pydantic defaults only apply to programmatic use of the models (see https://github.com/OvertureMaps/schema/pull/354#discussion_r2325747594). A default of 0 for an `int` field isn't weird though.

(This is present as a side-effect of the port from JSON Schema.)

packages/overture-schema-system/src/overture/schema/system/model_constraint/radio_group.py
```diff
@@ -0,0 +1,155 @@
...
+    Decorates a Pydantic model class with a constraint requiring that exactly one field in a group
+    of `bool` fields has the value `True`.
+
+    This function is the decorator version of the `RadioGroupConstraint` class.
+
+    Historical node for nerds: the term radio group, meaning a group of radio buttons, was first
+    used in software user interfaces design as an analogy to the mechanical push buttons on the
+    physical home and automobile radio sets of the 1950s-1970s. These radio sets had preset station
+    buttons where pushing one button would physically pop out any other buttons, ensuring only one
+    station could be selected at a time.
```
sethfitz, line 24: My first reaction when I saw the decorator in action was (yes, knowing that it's an analogy) "huh, using UI to describe this is weird," immediately followed by "but that's so much clearer than trying to communicate the effect" (which is what I was struggling with when I was first modeling this). Nice use of cross-domain terms.

packages/overture-schema-system/src/overture/schema/system/ref/id.py
```diff
@@ -0,0 +1,58 @@
...
+    ...     '''A room within a house.'''
+    ...     name: str = Field(description = 'Name of the room')
+    ...     house_id: Annotated[
+    ...         Id,
+    ...        Reference(Relationship.BELONGS_TO, House)
+    ...     ] = Field(description = "Unique ID of the house the room belongs to.")
+
+    When combining `Identified` with another Pydantic model that has an `id` field, such as a
+    :class:`~overture.schema.system.feature.Feature`, you must derive from `Identified` first in
+    order to correctly the *mandatory* `id` field.
```
sethfitz, line 48:
```suggestion
    order to correctly <what> the *mandatory* `id` field.
```

link?

sethfitz, line 48: How is this enforced?

packages/overture-schema-system/src/overture/schema/system/optionality.py
```diff
@@ -0,0 +1,96 @@
...
+    ```
+
+    In the above JSON Schema, the property `"bar"` is allowed to be omitted, but if present it must
+    be an integer. Under no circumstances can it contain the value `null`. The `Omitable` type
+    allows this to be modeled in Pydantic:
+
+    >>> from pydantic import BaseModel
+    >>> class MyModel(BaseModel):
+    ...     foo: str
+    ...     bar: Omitable[int]
```
sethfitz, line 60: I saw this approach in one of the Pydantic issues that discussed its, um, omission from Pydantic. I opted for the current approach to handling optional (nullable) values as omittable (where the schema gets rewritten) for a few reasons:

1. Omittable and nullable are indistinguishable everywhere except for JSON Schema and JSON (or document) representations. Omittable and nullable fields are both treated as nullable in tabular representations.
2. I didn't see a case where we'd ever need to distinguish between them.
3. The extra syntax needed *everywhere* (which is a hurdle for schema authors who are already unfamiliar with Pydantic and the conventions we've adopted; I'm also thinking of `x: int = Field(...)` vs. `x: Annotated[int, Field(...)]`, where I opted to standardize on latter): "I hope to replace the `X | None` unions in the existing model with `Omitable[X]`."
4. Authors don't need to do anything to get the right behavior.

I still think these points are valid and that the relative clarity of `X | None` outweighs fixing it for a very narrow use.

packages/overture-schema-system/src/overture/schema/system/optionality.py
```diff
@@ -0,0 +1,96 @@
...
+from types import NoneType, UnionType
+from typing import Annotated, Any, Generic, TypeVar, Union, get_args, get_origin
+
+from pydantic import Field
+from pydantic.experimental.missing_sentinel import MISSING
+
+T = TypeVar("T")
+
+
+class Omitable(Generic[T]):
```
sethfitz, line 11: According to various dictionaries:

```suggestion
class Omittable(Generic[T]):
```

(and references)

packages/overture-schema-system/src/overture/schema/system/model_constraint/min_fields_set.py
```diff
@@ -0,0 +1,130 @@
...
+            )
+
+    @override
+    def edit_config(self, model_class: type[BaseModel], config: ConfigDict) -> None:
+        super().edit_config(model_class, config)
+
+        json_schema = get_static_json_schema(config)
+
+        try:
+            prev = json_schema["minProperties"]
```
sethfitz, line 120: I'm torn on the `@min_fields_set` name, but I think I'm persuaded that the presence of "set" as a way to more clearly indicate its purpose outweighs following JSON Schema terminology because assuming that schema authors have JSON Schema experience is likely incorrect (and for those that do, we can provide a mapping in the guide; I've left myself a note there).

packages/overture-schema-system/src/overture/schema/system/create_model.py
```diff
@@ -0,0 +1,56 @@
...
+
+from .metadata import Metadata
+
+ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
+
+
+def create_model(
+    model_name: str,
+    /,
+    *,
```
sethfitz, line 14: TIL about `/` and `*` as positional- and keyword-only separators. Neat!

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
sethfitz: _Not required_ is the default though, yes? I agree that creating explicit rules is good (particularly to override something that's been marked as required, but that feels like a modeling mistake depending on guidance for semi-required fields), but I don't see the bug.

packages/overture-schema-system/tests/model_constraint/test_require_any_of.py
```diff
@@ -11,27 +11,27 @@
     require_any_of,
 )
 
-sys.path.insert(0, str(Path(__file__).parent))  # Needed to import `util` peer module.
+sys.path.insert(0, str(Path(__file__).parent.parent))  # Needed to import `util` module.
```
sethfitz, line 14: I'm still curious if adding `__init__.py` to `tests/` and here (making them modules) will make this work (but don't have the bandwidth to try it out).

README.pydantic.md
```diff
@@ -90,7 +90,7 @@ This workspace contains the following packages:
   convenient usage
 - **`overture-schema-core`** - Base classes, geometry models, and common structures
   shared across all themes
-- **`overture-schema-validation`** - Validation system with constraints and mixins
+- **`overture-schema-system`** - Foundational system of primitivef types and constraints
```
sethfitz, line 93:
```suggestion
- **`overture-schema-system`** - Foundational system of primitive types and constraints
```

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
vcschapp: It was getting cloned into `properties`, but was somehow also getting replicated at the top level.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
vcschapp: `required`  is a funny beast.

The positive version means "it must be the case that it is there" and the negative version means "it must not be the case that it is there".

So if you check out this mini-schema at https://www.jsonschemavalidator.net/...

```json
{
  "type": "object",
  "not": { "required": ["foo" ] },
  "required": ["bar"]
}
```

You find that this passes: `{"bar": 42}`, but this fails: `{"foo": "anything", "bar": 42}`.

---

*Edit*: Fixed an example that wrongly ended in a bracket `]` instead of a brace `}`.

packages/overture-schema-addresses-theme/tests/address_baseline_schema.json
vcschapp: What I mean by stricter is that `"additionalProperties": false` will prohibit any properties from being included if they are **not** either explicitly listed in `properties` or covered by an allowed pattern in `patternProperties`. So the fact that it won't traverse sub-schemas introduced through aggregators `allOf`, `anyOf`, *etc.*) makes it stricter.

*e.g.* consider this JSON Schema:

```json
{
  "type": "object",
  "additionalProperties": false,
  "properties": {
     "foo": { "type": "integer" } 
  },
  "patternProperties": {
    "bar.*": { }
  },
  "allOf": [{
    "properties": {
      "baz": { }
    }
  }]
}
```

The above will work for `{"foo":42,"bark": "anything"}` but not `{"foo":42,"bark": "anything", "baz":"anything else"}`.

The reason this makes sense to me in the Pydantic context is that Pydantic models always have enumerated properties in the `properties` section. The one exception so far is `ext_*`, which is covered by `patternProperties`. All of our other applicable constraints don't *add* allowed properties, they just constrain the values the existing properties are allowed to have.

---

*Edit*: Added a missing **not** that was super confusing.

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
vcschapp: I think, IIRC, that `@forbid_if` won't let you override something that is required. Meaning, if it is required then `@forbid_if` will throw.

It's limited to the case of "this already *optional* field is also *prohibited* if this other condition holds".

packages/overture-schema-system/src/overture/schema/system/model_constraint/model_constraint.py
```diff
@@ -195,29 +211,309 @@ def get_model_constraints(
...
+    def validate_class(self, model_class: type[BaseModel]) -> None:
+        super().validate_class(model_class)
+
+        required_fields = [
+            f for f in self.field_names if model_class.model_fields[f].is_required()
+        ]
+        if required_fields:
+            raise TypeError(
+                f"`{self.name}` expects all the fields to be optional, but at least one is required in the model class `{model_class.__name__}`: {', '.join(required_fields)}"
+            )
```
vcschapp, line 337: (Comment for linking purposes.)

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
vcschapp: > I think, IIRC, that `@forbid_if` won't let you override something that is required. Meaning, if it is required then `@forbid_if` will throw.
> 
> It's limited to the case of "this already _optional_ field is also _prohibited_ if this other condition holds".

Yup it inherits this behavior from `OptionalFieldGroupConstraint`: https://github.com/OvertureMaps/schema/pull/403/files#r2445351121.

packages/overture-schema-system/tests/model_constraint/test_require_any_of.py
vcschapp: I likewise am short on bandwidth. Hopefully one of us can do it.

packages/overture-schema-system/src/overture/schema/system/model_constraint/min_fields_set.py
vcschapp: I used *fields* in *lieu* of properties because it is the Pydantic term and I think we should assume Pydantic-oriented authoring.

---

I think "set" is a Pydantic term. It's used in the property `model_fields_set`.

```python
>>> class Foo(BaseModel):
...     bar: int | None = None
... 
>>> Foo().model_fields_set
set()
>>> Foo(bar=42).model_fields_set
{'bar'}
>>>

File: packages/overture-schema-addresses-theme/tests/address_baseline_schema.json (hunk 2 above)
Comment (sethfitz) [reply packages/overture-schema-addresses-theme/tests/address_baseline_schema.json:128]: TIL, wow. That's worth capturing (if it's not already) in the code that produces expressions like that.

File: packages/overture-schema-addresses-theme/tests/address_baseline_schema.json (hunk 3 above)
Comment (sethfitz) [reply packages/overture-schema-addresses-theme/tests/address_baseline_schema.json:225]: Aie, that's confusing.

Given the behavior you're showing, we might do well to avoid ever creating aggregators that introduce sub-schemas.

File: packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py [hunk 21]
```diff
@@ -0,0 +1,146 @@
...
+    condition: Condition,
+) -> Callable[[type[BaseModel]], type[BaseModel]]:
+    """
+    Decorates a Pydantic model class with a constraint forbidding any of the named fields from
+    holding an explicitly-assigned value, but only if a field value condition is true.
+
+    To ensure parity between Python and JSON Schema validation, a field's value must be explicitly
+    set to violate the constraint. This means in particular that fields whose value was set by
+    Pydantic using a default value do not count as having a set value, and fields containing the
+    value `None`, if this value was explicitly set rather than being inherited by default, do count.
```
sethfitz, line 25: Missing: a comment that this won't let you override something that is required (and will `raise`...when?).

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
sethfitz: Based on ☝️, my understanding is that `not.required` is **not** the default (because it actively blocks). "Not required" (or optional) *is*. Following from that, this means that the `not.required` in the annotated code means "shouldn't be present" and that's the bug fix, correct?

The raising behavior is nice. Will it raise at definition time (runtime), so it'll pop up when you try to generate JSON Schema from it? I stuck a note on `@forbid_if` to document the "won't let you override something that is required" behavior.

packages/overture-schema-system/src/overture/schema/system/model_constraint/min_fields_set.py
sethfitz: > I think we should assume Pydantic-oriented authoring.

I agree! (This feeds into the meta-issue about open vs. closed behavior, where we're falling into the trap of paying attention to the JSON Schema behavior over Pydantic (my bad).)

That's not the meaning of "set" that I was expecting ("make sure to _set_ at least _n_ fields"), but good that it matches the other meaning too!

packages/overture-schema-system/tests/model_constraint/test_require_any_of.py
sethfitz: Fixed by adding a `pytest` configuration: #407

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
vcschapp: It raises at the time of applying the decorator, so basically you'll have to cure the error before you can go on to generate the JSON Schema from the model.

packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
vcschapp: Yeah this is a good idea. Could even be a doctest example. I'll try to wedge it in the next iteration.

packages/overture-schema-divisions-theme/tests/division_boundary_baseline_schema.json
sethfitz: Perfect!

mojodna:
> I responded to some comments about not.required and additionalProperties. Let me know what you think. For not.required, I can always pull the baseline schema from before and after and show both how it differs from the hand-written JSON Schema and also how the behavior differs, but hopefully my explanation is clear enough.

Yes, thanks for clarifying. Definitely confusing behavior on the JSON Schema side, but not something we need to surface to Pydantic builders (esp. if we get it right).

> I agree with this. I think today that is quasi-handled by the ext_* enabling code which already puts on "additionalProperties": false but allows through the narrow exception of ext_*. The ext_* exception is problematic, but when we remove it we should be left with all OvertureFeatures behaving as if @no_extra_fields was applied.
>
> Let me know what your thoughts are on this.

Re: `OvertureFeature` directly, that's also my understanding/expectation.

I was referring to *other* `BaseModel`s that get attached to `OvertureFeature` as fields. When modeling our particular flavor of tabular data (including potentially-nested lists, dicts, and structs (schemas/models)), "closed" (no additional fields) should be the default, as there's no mechanism in the formats used to serialize this data (other than JSON) to support "open" lists of fields. This is in contrast to API definitions (because they use JSON), where supporting extra fields is no big deal, and could support undocumented functionality in a useful way.

Pydantic optimizes for the API definition use case, so it's open by default. I believe that our "flavo[u]r" of Pydantic should default to "closed", even when schema authors don't include `@no_extra_fields` in their sub-schemas. That might lead to confusion for people used to Pydantic, but they'll find that the behavior is wrong otherwise when used with real data.

> FWIW, it was a deliberate spelling choice so I'd argue not a typo. The word wasn't in the dictionary so I picked a spelling.

Fair enough ;-) This is from the New Oxford American Dictionary, which is the one that ships with macOS (and is neither Canadian nor old):

image

> The choice of word is to differentiate it from Python Optional even though that's being phased out.

Yep, that was the same reason I remember from the GitHub issue I can't find. Makes sense.

> Agree that the primary purpose is to give the Pydantic model the same amount of richness as JSON Schema by de-conflating "nullable" from "may be omitted". It would also be useful in any other context in which these two concepts are separated - more may arise.
>
> I think it's valuable. The reason I prefer it is that there IS a difference between a required field with { "anyOf": [ { "type": "string" }, { "type": "null" ] } and an optional field of type "string". That's why our current code base has a custom JSON Schema serializer that tries to make these edits on the fly. The problem in my mind is that means the Pydantic does not specify the JSON Schema, as it should, it specifies one JSON Schema that isn't intended and if you want to get to the one that is intended, you have to know to use a custom schema serializer. That seems wrong to me... In my mind, the Pydantic model should automatically embed the correct JSON Schema.

From a JSON/JSON Schema point of view, I agree with you. However, I think that's the wrong lens; coming from JSON Schema doesn't mean that we need to support its semantics, especially if they don't have value for the type of data we're describing. I think we should start from the Pydantic perspective (which is what people will be authoring, even if it doesn't fully support the richness(?) of JSON Schema) and allow that to influence us vs. letting an implementation that we're moving away from (in part because it's confusing) drive the bus.

Alternately, I'd be ok saying that Overture *only* has **optional** fields (nothing can be omitted, and removing the custom serializer), which maps cleanly to tabular representations. It has the downside of producing more verbose JSON Schema (which also differs from our hand-written version) and making all of our JSON/YAML representations invalid, eliminating the convenience we've derived from them. I think the convenience wins though, since it doesn't change the actual validation behavior.

What's a concrete example where `Omitable`'s semantics make sense in our world other than the convenience of JSON/YAML representations?

vcschapp:
> > There's an argument that we're over-thinking this some.
> 
> Probably ;-)
> 
> > When looking at code-generated tools it may be mostly a non-issue. For something like a Spark dataframe writer, we can just generate the code to ignore all the "extra" fields, and I would be tempted to have the generator emit an optional warning/error if the model allows extra fields. For a reader, again we can just ignore any extra columns (top-level model) or again have the code generator emit an optional warning/error if the model allows extra fields. I'd go this way: allow authors to do whatever they say they want, but warn them if what they say they want would produce unexpected results in, say, Spark reading.
> 
> Agreed. Spark Datasets will ignore columns that aren't present in the class definition (`X` for `Dataset[X]`), although they'll be accessible when accessing the underlying DataFrame. Same when working with JSON (or `dict`s) and Pydantic / JSON Schema validation. I have some lingering discomfort about not flagging unexpected columns that "shouldn't" be present but that will end up getting dropped when serialized.
> 
> > which I have an aesthetic preference for
> 
> As do I. It's downright convenient to be able to work with sparse JSON.
> 
> > The one thing Pydantic is missing is "full omit[t]ability" without a default, which is semantically the same as the others but just represented differently at the JSON level. This is what [EnhancedJsonSchemaGenerator](https://github.com/OvertureMaps/schema/blob/pydantic/packages/overture-schema-core/src/overture/schema/core/json_schema.py#L11-L57) is meant to solve, and my complaint has been that the solution is an optional post-processing step you have to know about and apply, rather than allowing the desired semantics to be embedded in the model. The other problem with it is that it can cause the JSON Schema to diverge from Pydantic behavior since if you do `Foo().model_dump_json()` in my above example, you will get `{"bar":null}` (unless you set `exclude_none=True`).
> 
> `parse_feature` currently sets `exclude_unset=True` (which I think is more appropriate than `exclude_none`; in addition to using `EnhancedJsonSchemaGenerator`) for that reason:
> 
> https://github.com/OvertureMaps/schema/blob/941c8df854294b16e7becdef1e7dd38239f97e5b/packages/overture-schema-core/src/overture/schema/core/parser.py#L50
> 
> > 1. Drop `Omitable`.
> > 2. Re-home `EnhancedJsonSchemaGenerator` into the system package (maybe with a more meaningful name).
> > 3. Have the our CLI default behavior be "full omission" using `EnhancedJsonSchemaGenerator` and `exclude_none=True` so that the Overture schema continues to work the way it always has, but allow this default to be turned off via some command line option.
> 
> Aligned, other than switching to `exclude_unset=True`. This is already how the CLI behaves.

I think we have the path forward!

And agreed that `exclude_unset=True` is the right way, not `exclude_none`.

PR #406, opened 2025-10-19
[Pydantic] overture-schema CLI for type listing, JSON Schema generation, and validation

sethfitz:
```console
Usage: overture-schema [OPTIONS] COMMAND [ARGS]...

  Overture Schema command-line interface.

  Provides validation, schema generation, and type discovery for Overture Maps
  data.

  Examples:
    # Validate a file
    $ overture-schema validate data.json

    # Validate from stdin
    $ overture-schema validate - < data.json

    # List available types
    $ overture-schema list-types

    # Generate JSON schema
    $ overture-schema json-schema --theme buildings

    # Validate specific types
    $ overture-schema validate --theme buildings data.json

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  json-schema  Generate JSON schema for Overture Maps types.
  list-types   List all available types grouped by theme with descriptions.
  validate     Validate Overture Maps data against schemas.
```

`overture-schema list-types`:

`overture-schema validate --type division --show-field id --show-field names`:

![─ \[0\] (Division) id=23e81262-d6ed-45a3-a1a0-4bc6a2... ──────────────────────────────────────────────────────────────── 
       ... 
        id "23e81262-d6ed-45a3-a1a0-4bc6a2a887d8" 
      bbox      {xmin: ..., xmax: ..., ymin: ...} 
   country                              <missing> ← Input should be a valid string 
   version                                      1 
    names   primary: "Amundsen–Scott South Pole"
       ... 
  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────](https://github.com/user-attachments/assets/9811c27d-4b2c-44f1-9cda-f09be1be8f38)

I anticipate introducing a `[parquet]` variant in the future that can parse Parquet directly, but in the meantime, this works:

```shell
duckdb -c <<-'SQL'
	INSTALL spatial;
	LOAD spatial;
	INSTALL httpfs;
	LOAD httpfs;
	SET s3_region='us-west-2';
	COPY (
	  SELECT ST_AsGeoJSON(geometry) as geometry, * EXCLUDE geometry
	  FROM read_parquet('s3://overturemaps-us-west-2/release/2025-09-24.0/theme=divisions/type=division/*.parquet', hive_partitioning=true)
	  LIMIT 100
	) TO '/dev/stdout' (FORMAT JSON, ARRAY false);
SQL
 | uv run overture-schema validate --type division --show-field type --show-field id -
```


vcschapp:
> > 1. I see the CLI only depends on the "core" package. If we migrate the model discovery stuff down into system, does that mean we can drop the direct dependency on core and just have the CLI depend on system?
> 
> Yes. The full list of dependencies from core are https://github.com/OvertureMaps/schema/pull/406/files#diff-bcbfe867ab7a1405a6384886b8ed2975dc659d798cb72af5fdc18a71e5617298R5-R13:
> 
> ```python
> from overture.schema.core import parse_feature  # sets exclude_unset=True
> from overture.schema.core.discovery import discover_models  # model discovery mechanism
> from overture.schema.core.json_schema import json_schema  # this uses EnhancedJsonSchemaGenerator
> from overture.schema.core.parser import (
>     # list[BaseModel] variant
>     parse_features,
>     # validate-only variants
>     validate_feature,
>     validate_features,
> )
> from overture.schema.core.unions import create_union_from_models  # dynamically creates unions from discovered models for use by `*_feature[s]`.
> ```

Awesome! 🤯

vcschapp:
> > 2\. Functionally, it would be ideal if the CLI can be PIP-installed and just detect all the entry-points available from the CWD including globally-installed packages and workspace/env packages. How far are we from that?
> >     
> >     * I suspect entry-point discovery will discover from globally-installed packages.
> >     * But can we also discover from the current venv/workspace?
> 
> I don't know. I haven't tried this yet. Within a specific Python environment, it will detect all entry points. That _may_ extend to anything in `PYTHONPATH`, but setuptools' hooks may only apply to packages that are "installed."

I'm pretty sure it should work and we should aim for that deployment mode. Ideally the CLI is generic functionality that can be applied to any schema and doesn't have to depend on a specific schema or schema version.

I would ideally love to have all the following 8 configurations work the same ... Where the CLI can "discover at the same level and downward"...


2025-10-31-overture-schema-cli-package-architecture drawio

I think we can achieve this roughly. I haven't been able to capture all the nuance yet, but it's definitely possible for a Python program to figure out what "tier" it is running at (system, user, venv) and also to detect if the CWD has a venv.

vcschapp:
> > 4\. Are the theme and type that the CLI depends on just values parsed from the entry-points?
> > 
> > * I assume yes.
> > * A yes mean the CLI doesn't depend on `OvertureFeature` or any of the models in core, which would be ideal.

> Yes. The only Overture-specific implementation is the `--overture-types` flag (e.g., https://github.com/OvertureMaps/schema/pull/406/files#diff-5c898d587fe2fbc5e7d402913e372a553edfbe979182578cf3d8176381dfbf35R457-R461), which sets the namespace to use when discovering models: https://github.com/OvertureMaps/schema/pull/406/files#diff-5c898d587fe2fbc5e7d402913e372a553edfbe979182578cf3d8176381dfbf35R66-R67

Awesome. 👍

vcschapp (approved):
Approved. I merged PR #412 so there may be some minor rebase conflicts - sorry!

packages/overture-schema-core/src/overture/schema/core/parser.py
```diff
@@ -6,6 +6,87 @@
 from ._cache import get_type_adapter
 
 
+def _flatten_feature(feature: dict[str, Any]) -> dict[str, Any]:
```
vcschapp: Now that the core GeoJSON feature is down in `system`, the GeoJSON custom parsing, if needed, should go down there.

---

I don't think we should need a special parser assuming that the existence of `"properties"` with `"type": "Feature"` makes it GeoJSON though. ... We already have `BaseModel.model_validate` for Python mode and `BaseModel.model_validate_json` for JSON mode (or more accurately `TypeAdapter.validate_python` and `TypeAdapter.validate_json`). ...

I think we should expect that users know the context or mode (Python or JSON) in which they are calling the method. If they do, they just need `TypeAdapter.validate_python` or `TypeAdapter.validate_json` and the GeoJSON handling built into the `Feature` model will take care of the rest on the JSON side.

IMO we shouldn't need to repeat the GeoJSON rotations anywhere outside the `Feature` class - if we do then we probably need to embark on some more refactoring....

PR #411, opened 2025-10-23
Add ROLLER_COASTER enum to enums.py

jenningsanderson:
Adds `roller_coaster` to list of enums for infrastructure

Goes with https://github.com/OvertureMaps/tf-data-platform/pull/1979/files

### DIFF
--- a/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
+++ b/packages/overture-schema-base-theme/tests/infrastructure_baseline_schema.json
@@ -237,12 +237,20 @@
           "$ref": "#/$defs/NameVariant"
         }
       },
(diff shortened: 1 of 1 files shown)


mojodna (changes requested):
Please re-baseline the generated schema to reflect this addition.

vcschapp (commented):
LGTM (once re-baselined per Seth's request).

mojodna (commented):
Did you manually edit/reformat `infrastructure_baseline_schema.json`? The idea is that you delete it, re-run the test (to re-create the file), confirm that the git diff matches your expectations, and then commit the change.

I'm puzzled by the differences...

PR #412, opened 2025-10-28
Pydantic package organization episode 4: "Scope is in (scope)"

vcschapp:
# tl;dr

This is part `#4`. Part `#5` is hoped for in the next week or two if all goes according to plan.

Done in this part: re-organized the addresses and places packages to their (hopefully) final form, finish dynamic scoping implementation (`@scoped` decorator) and apply it to names and transportation.

# Merge guidance: Rebase

Unlike parts 1, 2, and 3, this iteration has a well-written commit history. It should ideally be rebase-merged onto the `pydantic` branch rather than squashed.

# Journey

- Part 1 is in PR #397.
- Part 2 is in PR #400.
- Part 3 is in PR #403.

# Destination

1. **Packages**.
2. **Documentation**.
3. **Modules**.
   - ▶️ In theme packages, each feature type will be within its own non-exported module within the package root.


packages/overture-schema-places-theme/pyproject.toml
```diff
@@ -27,4 +27,4 @@ path = "src/overture/schema/places/__about__.py"
 packages = ["src/overture"]
 
 [project.entry-points."overture.models"]
-"places.place" = "overture.schema.places.place.models:Place"
+"places.place" = "overture.schema.places:Place"
```
sethfitz, line 30: Note that I made a similar change in the CLI PR. In that case, it may refer to an indirect alias, but either way, the intent is to register the shortest useful import for the class (because that path will be displayed when listing available types).

PYDANTIC_GUIDE.md
sethfitz, line 1: Thank you for updating this!

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+>>> from pydantic import BaseModel
+>>> from overture.schema.system.string import StrippedString
+>>> @scoped(Scope.TEMPORAL)
+... class WiFiHotspot(BaseModel):
+...     '''A public WiFi hotspot'''
+...     ssid: StrippedString
+...
+>>> hotspots=[
+...     WiFiHotspot(ssid="always_on"),
+...     WiFiHotspot(ssid="daytime_use_only", when=WiFiHotspot.When(during="06:00-22:00"))
```
sethfitz, line 170: This is confusing. Where does `WiFiHotspot.When()` come from? (Update: found it below)

Reading the description above, I would expect this to be:

```python
WiFiHotspot(ssid="daytime_use_only", when=dict(during="06:00-22:00"))
```

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+...     version=1,
+...     geometry=Geometry.from_wkt('LINESTRING (0 0, 1 1)'),
+...     transformers=[Transformer(at=0.73, power_capacity=167)]
+... )
+
+
+Geometric range scope (linear events):
+--------------------------------------
+Geometric range scoping allows a value to be tied to a sub-segment of a linear path using linear
+referencing. When a model is decorated with geometric range scoping, a `between` field is
```
sethfitz, line 103: Why are point and linear events and side treated differently from other scopes? Specifically, `at` / `between` / `side` at the top level where the others are all nested under `when`.

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+...     id='island_001',
+...     theme='bus_terminal',
+...     type='loading_island',
+...     version=1,
+...     geometry=Geometry.from_wkt('LINESTRING (0 0, 1 1)'),
+...     stops=[
+...         BusStop(route='15', side=Side.LEFT, at=0.10),
+...         BusStop(route='B-Line', side=Side.RIGHT, at=0.25),
+...     ]
+... )
```
sethfitz, line 241: I'm thinking through how I'd expand on this by modeling standalone bus stops... It seems like I'd need a different, non-scoped `BusStop` that extends from `OvertureFeature` if I want to model them directly. (I might mix in fields that are common to all bus stops, like `route`.)

How would I model bus stops along road segments (without changing the `RoadSegment` model, unless there's a generic `point_events` field or similar to capture these)? I suppose I could reference the road segment with an `at` (to indicate where on the road segment the stop is). It would kind of look like the equivalent of a point event even though it's not.

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+
+>>> @scoped(Scope.TEMPORAL, Scope.GEOMETRIC_RANGE, Scope.HEADING)
+... class MyModel(BaseModel):
+...     pass
+
+
+Optional and required scopes
+============================
+When using the `@scoped` decorator, scopes are optional by default but some or all scopes may be
+made required.
```
sethfitz, line 302: How does this translate to current behavior where at least 1 scope needs to be fulfilled (but it doesn't matter which one)?

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+...    STOP_SIGN = 'stop_sign'
+...
+>>> @scoped(required=(Scope.GEOMETRIC_POSITION, Scope.HEADING))
+... class TrafficSignal(BaseModel):
+...    signal_type: SignalType
+
+The following example mixes an optional scope (temporal) with two required scopes (geometric
+position and heading).
+
+>>> @scoped(Scope.TEMPORAL, required=(Scope.GEOMETRIC_POSITION, Scope.HEADING))
```
sethfitz, line 318: What are the geometric position and heading scopes relative to in this example?

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+... class TrafficSignal(BaseModel):
+...    signal_type: SignalType
+
+
+The `when` clause
+=================
+For historical reasons, some scope fields are added directly to the decorated model, while others
+are added as children of a synthetic `when` field.
```
sethfitz, line 326: How should we resolve this inconsistency?

Is the ideal state moving `at`, `between`, and `side` under `when`?

Would it be worth allowing authors to model using the ideal state while introducing a backward-compatibility shim that allows us to migrate our distributed data independently of modeling? (kinda similar to how we pivot JSON Schema to match GeoJSON, but probably at a higher level than JSON Schema generation)

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+| `Scope.TEMPORAL`           | `when.during`     |
+| `Scope.TRAVEL_MODE`        | `when.mode`       |
+| `Scope.PURPOSE_OF_USE`     | `when.using`      |
+| `Scope.RECOGNIZED_STATUS`  | `when.recognized` |
+| `Scope.SIDE`               | `side`            |
+| `Scope.VEHICLE`            | `when.vehicle`    |
+
+
+If a `when` field is added to the model, the model is also decorated with a nested `When` class to
+simplify instantiating values for the `when` field, for example:
```
sethfitz, line 342: These answered questions I had above. I'd open with this to provide more of an introduction before digging into specific concepts.

What is the nested When class? Reading this, I don't know what the right mental model is, which is relevant because I might try to duck type my way through with a dict (see above).

Another important thing to include may be that the generated/nested `When` class will include fields for *all* scopes attached to the model (presumably marked as required or optional as appropriate).

packages/overture-schema-core/src/overture/schema/core/scoping/lr.py
```diff
@@ -0,0 +1,104 @@
...
+        return json_schema
+
+
+LinearlyReferencedRange = NewType(
+    "LinearlyReferencedRange",
+    Annotated[
+        list[LinearlyReferencedPosition],
+        LinearReferenceRangeConstraint(),
+        Field(
+            description="Represents a non-empty range of positions along a path as a pair linearly-referenced positions. For example, the pair [0.25, 0.5] represents the range beginning 25% of the distance from the start of the path and ending 50% of the distance from the path",
```
sethfitz, line 101: Are ends inclusive or exclusive? Presumably inclusive, if we think about `0.0` and `1.0`. It may not matter, although it means that multiple ranges will overlap at common points rather than being contiguous.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
```diff
@@ -0,0 +1,495 @@
...
+                raise self._unexpected_scope()
+
+    def _field_description(self, parent: str) -> str:
+        match self:
+            case Scope.GEOMETRIC_POSITION:
+                return (
+                    "The linearly-referenced position on the geometry, "
+                    "specified as a percentage displacement from the start "
+                    f"of the geometry, that the containing {parent} applies to."
+                )
```
sethfitz, line 236: Can / should these pull from enum docstrings or related doc properties? It would help centralize the strings.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
```diff
@@ -0,0 +1,495 @@
...
+from .lr import LinearlyReferencedPosition, LinearlyReferencedRange
+from .opening_hours import OpeningHours
+from .purpose_of_use import PurposeOfUse
+from .recognized_status import RecognizedStatus
+from .side import Side
+from .travel_mode import TravelMode
+from .vehicle import VehicleSelector
+
+
+class Scope(str, Enum):
```
sethfitz, line 37: Should this be a `DocumentedEnum`?

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
```diff
@@ -0,0 +1,495 @@
...
+
+def scoped(
+    *optional: Scope,
+    required: Scope | Iterable[Scope] | None = None,
+) -> Callable:
+    """
+    Returns a decorator to decorate a Pydantic model class with one or more scoping attributes.
+
+    At least one `Scope` must be given either in `optional` or `required` or both. A scope may be
+    optional or required, but not both.
```
sethfitz, line 313: How does one indicate that *a* scope must be provided without indicating which one?

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
```diff
@@ -0,0 +1,495 @@
...
+            when_fields: dict[str, Any] = {
+                s._field_name: s._field(parent, s in required) for s in when_scopes
+            }
+            when_class = _make_when_class(model_class, when_fields, description)
+            if has_required:
+                # If any `when` field is required, `when` is itself required.
+                scoped_fields["when"] = (when_class, Field(description=description))
+            else:
+                # If the `when` has no required fields but contains multiple optional fields, it is
+                # optional, but if present we require that at least one of its fields be set.
```
sethfitz, line 445: Aha! Here it is. This behavior should be documented above.

mojodna (approved):
Nothing blocking, but I have lots of questions about scopes (surprise, surprise).

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: Historical reasons.

The scoping concept came together somewhat organically (inductive reasoning). The existence of `when` can be explained by the fact that some fool thought "it read nicer in JSON", *i.e.* `{ "between": [0, 0.5], "when": { "mode": ["foot"] } }`...

In retrospect, the existence of `when` seems regrettable. We should probably phase it out. But for now, it is what it is...

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: Modeling them along road segments might be done in several ways but they're all variations of two options.

If you can change the road segment model then you probably add a `bus_stops` field similar to how it's down in the example.

If you can't change the road segment model, it probably means that bus stops are independent point features. If we want them to reference the road network you end up with something like (maybe):

```python
@scoped(Scope.GEOMETRIC_POSITION)
class BusStop(OvertureFeature[Literal["transportation"], Literal["bus_stop"]]):
    geometry: Annotated[
        Geometry,
        GeometryTypeConstraint(GeometryType.POINT)
    ]

    road_segment_id: Annotated[
        Id | None, 
        Reference(Relationship.CONNECTS_TO, RoadSegment)
    ]
```

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: Does that behavior exist anywhere?

If it does, I might have to tweak the logic a bit.

Operating mostly from memory, if all the scoping fields are omitted that roughly means "this fact applies all the time".

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: Start of the geometry.

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: I think the ideal state is probably getting rid of `when`.

At the moment it's hard for me to imagine a shim that would quite get the job done at all the required levels. I'm tempted to punt and hope for a beautiful tomorrow where we just get rid of `when` with an MV bump.

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
vcschapp: > for *all* scopes attached to the model

Excepting the ones that don't go in `when` (at/between/side).

packages/overture-schema-core/src/overture/schema/core/scoping/lr.py
vcschapp: They're inclusive. It's funny but I think the fact that we even wonder about that is an artifact of the fateful decision to use percentages instead of absolute numbers. If they were absolute numbers we wouldn't wonder!

In practice some common sense is probably needed when coding against the data but that's true regardless of how you describe it. For *e.g.*, if the speed limit is 80 from 0 to 0.25 and 100 from 0.25 to 1.0, it obviously can't be both 80 and 100 at 0.25, so you have to pick one, but in real life it's not a material distinction.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
vcschapp: These get added as Pydantic descriptions on the fields that are generated by the `@scoped` decorator. So code generation software just has to traverse the model as it normally does and it'll see the descriptions.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
vcschapp: Possibly, but here's my logic for not.

I see `@scoped` and `Scope` as predominantly builder-oriented constructs. Tools for schema authors. For the builder/author persona, the API reference is probably their go-to documentation source, so it makes sense to write the docs in a way that will be pull into the API reference. (`pdoc` knows how to find the pseudo-docstrings below the enum members and document the members as if they were real docstrings in its generated API reference).

On the other hand, the data consumer persona just wants to understand the structure of the data. The place to go to get that understanding is the schema reference documentation that we're code-generating. Our documentation "code" generator will traverse the model and see the Pydantic descriptions, which are derived from the properties in the `Scope` enum and are pretty comprehensive. So hopefully that persona will be well-served already.

The `DocumentedEnum` is intended to provide documentation for the data consumer personal in cases of consumer-facing enums like `RoadClass`. The `RoadClass` isn't really for builders/authors it's for data consumers. Since we don't have a mechanism to attach documentation to the enumeration members that our code generators can understand, the intention is for `DocumentedEnum` to be that mechanism.

But, importantly,, `pdoc` doesn't know about `DocumentedEnum` and doesn't look for `__doc__` on enum members so using `DocumentedEnum` in this case instead of pseudo-docstrings would cause us to lose docs for the persona we care about, the builder/author.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
vcschapp: That's currently not possible.

I didn't think it was needed for any existing use case, but if I'm wrong let me know so I can fix it.

(Assuming we don't use that construct today: I can imagine it's something we'll need in the future. The best thing to do would be to drop `when` and just use `@require_any_of` to indicate this constraint. With `when` in the mix, it's messy.)

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
vcschapp: But I don't think this does what you're asking about [above](https://github.com/OvertureMaps/schema/pull/412#discussion_r2479050060) 👆

Above, the use case you're asking about is "I want to stipulate that at least one of the possible scoping fields must be set". 

This comment is just saying, if all the fields in `when` are optional then the `when` itself is optional. But if you provide a `when` you have to provide at least one field. This is just a sanity requirement to reduce the number of possible ways of saying the same thing. Since `"when": null` means the same as `"when": {}`, it's prohibiting you from saying it the second way just to reduce the number of synonym states.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
sethfitz: [`SpeedLimitWhenClause`](https://github.com/OvertureMaps/schema/blob/pydantic/packages/overture-schema-transportation-theme/src/overture/schema/transportation/models.py#L447-L477) is what I was thinking of, with `@min_fields_set(1)`

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
sethfitz: Start of which geometry? I think I was more reacting to the lack of a linear `OvertureFeature` (in this example) that `TrafficSignal`s would be attached to.

Semi-related, is there (should there be) any sort of validation when using geometric position scopes on non-linear geometries (or MultiLineStrings)?

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
vcschapp: Got it. Yeah, that was a faithful modeling of [this hand-written JSON Schema](https://github.com/OvertureMaps/schema/blob/dev/schema/transportation/segment.yaml#L590-L599). At least for now it's still just trying to [block multiple ways of the `when` clause saying the same thing](https://github.com/OvertureMaps/schema/pull/412#discussion_r2479847874).

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
sethfitz: I'm not sure how to phrase it, but we should create an issue (and probably a label for breaking changes we want to make generally) to capture phasing this out, even if we don't have a good idea for how to migrate it yet.

packages/overture-schema-core/src/overture/schema/core/scoping/__init__.py
sethfitz: That's what I was envisioning, but I was thinking 1 step further: a **linear** reference to the road segment where it not only describes what it connects to, but where.

We could do something like this:

```python
@scoped(Scope.GEOMETRIC_POSITION)
class BusStop(OvertureFeature[Literal["transportation"], Literal["bus_stop"]]):
    geometry: Annotated[
        Geometry,
        GeometryTypeConstraint(GeometryType.POINT)
    ]

    road_segment_id: Annotated[
        Id | None, 
        Reference(Relationship.CONNECTS_TO, RoadSegment)
    ]

    road_segment_at: LinearlyReferencedPosition | None = None
```

but the combination of the `CONNECTS_TO` and linear references implies to me that some form of nested structure would be valuable, especially if both are required at the same time.

packages/overture-schema-core/src/overture/schema/core/scoping/scoped.py
sethfitz: I was thinking more about the maintenance benefits of centralization than code generation.

PR #414, opened 2025-11-04
[WIP] Update divisions subtype naming convention

stepps00:
# Major change release plan

Dec 2025

TODO

- [ ] [Theme guide](https://docs.overturemaps.org/guides/divisions/) updates
- [ ] Documentation in the release notes

# Description

The divisions theme has discussed moving to an adminN (admin0, admin1, admin2, admin3) subtype naming convention for top-level divisions, while keeping more semantic subtypes (locality and below) intact.

Some rationale for this change:

- Every top-level division would fit cleanly into a numbered level.

- Easier to query and maintain, users understand what admin1 or admin2 mean and these terms are interoperable with other open administrative datasets (geoBoundaries partnership?).

- Frees up words like “region” for other feature types.

- Works for any country or future level without redefining new types for existing features.

**Proposed mapping**

Existing subtype | New subtype
-- | --
country | admin0
dependency | admin0
macroregion | N/A
region | admin1
macrocounty | N/A
county | admin2
localadmin | admin3
locality | locality
borough | borough
macrohood | macrohood
neighborhood | neighborhood
microhood | microhood

Notes:

- Both country and dependency records would be classified as admin0. Dependencies would maintain a `parent_division_id` value equal to the GERS ID of the administering country.
- The theme does not (and has not) maintained macroregion or macrocounty features. These subtypes would not match one-to-one to a new adminN subtype
- This change simply renames the subtypes - it will not impact what features appear at these subtype levels - ie all regions become admin1, all localadmin become admin3, etc.

All references and examples / counterexamples have been updated with these new adminN subtype names.

# Reference

TODO

# Testing

TODO

# Checklist

1. [x] Add relevant examples.

# Documentation website


counterexamples/divisions/division_boundary/missing-country.yaml
```diff
@@ -8,10 +8,10 @@ properties:
   theme: divisions
   type: division_boundary
   version: 0
-  subtype: region
+  subtype: admin1
   is_territorial: true
   is_land: false
   class: land
-  division_ids: ["example:division:country:left", "example:division:country:right"]
+  division_ids: ["example:division:admin0:left", "example:division:admin0:right"]
```
sasastanojkov: Seems like nobody cared about these ids to begin with :D
```suggestion
  division_ids: ["example:division:admin1:left", "example:division:admin1:right"]
```

schema/divisions/division.yaml
```diff
@@ -25,12 +25,19 @@ properties:     # JSON Schema: Top-level object properties.
       - "$ref": ../defs.yaml#/$defs/propertyContainers/cartographyContainer
       - if:
           properties:
-            subtype: { enum: [country] }
+            subtype: { enum: [admin0] }
         then:
           required: [names, subtype, country, hierarchies]
-          not: { required: [parent_division_id] }
```
sasastanojkov, line 31: Why remove this line. From what I gather, it is not required for admin0 to have parent, but it can have parent right?

schema/divisions/division.yaml
```diff
@@ -25,12 +25,19 @@ properties:     # JSON Schema: Top-level object properties.
...
         else:
           required: [names, subtype, country, hierarchies, parent_division_id]
+      - if:
+          properties:
+            subtype: { enum: [admin0] }
+            parent_division_id: {}
+        then:
+          properties:
+            parent_division_id:
+              pattern: ":admin0:"
```
sasastanojkov: Can you elaborate on this part all together?

schema/divisions/division.yaml
```diff
@@ -139,7 +146,7 @@ properties:     # JSON Schema: Top-level object properties.
         description:
           Division ID of this division's parent division.
           
-          Not allowed for top-level divisions (countries) and required
+          Not allowed for top-level divisions (admin0) and required
```
sasastanojkov: This is now just not true.

sasastanojkov (changes requested):

schema/divisions/division.yaml
stepps00: Sure, the intention here is to check if an admin0 feature has the parent_division_id property, and if so, the schema now checks that the ID literally contains `:admin0:`  (ex: `example:division:admin0:us`).

That way only admin0 parents are allowed, so a dependency can’t accidentally point at some admin1 (or random) record.

schema/divisions/division.yaml
stepps00: > From what I gather, it is not required for admin0 to have parent, but it can have parent right?

Yeah that's right.

The way I understand it, that old `not: { required: [parent_division_id] }` block rejected any admin0 record that included the field, which would have prevented us from modeling dependency -> country.

Removing it lets the field exist when needed, and the new pattern check makes sure that it points to another admin0 (if it exists).

sasastanojkov: Ah yea. I had that some time ago. Yea that is ugly but I understand the logic behind it.

schema/divisions/division.yaml
sasastanojkov: So we are expected to have admin0 gers ids with `:admin0:` or how does it know to check subtype?

stepps00: Ah you're right, I misunderstood what this was testing. I've removed this if/then statement.

Related - as it stands, the JSON Schema still requires the `parent_division_id` for subtypes other than admin0, but for admin0 features the field is optional. If present, we're only checking the GERS-format (defined in defs.yaml).

vcschapp (commented):
What's the underlying motivation for this change, and are you aware of some of the design choices in having named subtypes?

---

- One of the nice parts of the current design is that the terms `country` and `region` are used consistently in various contexts (`country` and `region` subtypes have a corresponding `country` and `region` property or column)...
- As well there's just general consistency: are subtypes descriptive names, or are they numbers?
- And the "admin level" abstraction can be derived today from the primary hierarchy or following the `parent_division_id` tree to the root (which can be written up in SQL).

---

**Motivation**: Is the main motivation to give people a simple "1, 2, 3..." level abstraction?

Like I said this *can* be derived from the data today, but if we view that as too user-unfriendly, may I suggest the alternative of just adding a `rank` property that has the 1-based level of the feature in the primary hierarchy tree?

vcschapp:
One extra thought: the advantage of a `rank` or similar column is that it contains an integer value, and you have more ready query options with integers.

*e.g.*

```
WHERE rank >= 2
```

Is a lot easier than:

```
WHERE subtype IN ('admin2', 'admin3', 'admin4', 'admin5', 'admin6')
```

Or

```
WHERE CAST(REGEXP_EXTRACT(subtype, 'admin(\\d+)') AS INTEGER) >= 2
```

vcschapp:
It was noted in the [schema meeting this morning](https://github.com/OvertureMaps/schema-wg/issues/391)  that:

1. Calling things "admin0", ... "admin3" is an industry standard. 
2. Dependencies like Puerto Rico and Guam are "admin0" even though they conceptually are subject to the administration of another country that is also going to be "admin0".

For #1, for my education it would be super helpful to understand the players who do (and do not) use this scheme. (We know that WOF and OSM do not.)

For #2, this seems odd to a layman like me.

- Why is Puerto Rico "admin0" if it is a "child" of the USA which is also "admin0"?
- Since it is not a country, and this PR is proposing to delete the information that it is a dependency, are we sure the benefits of the change make up for the reduction in information content?

TristanDiet-TomTom (commented):
The Industry standard feels very much tailored on the USA 'official' hierarchical structure:
country > state > county > 'local admin' (aka 'go figure')

Also the proposed mapping reflects this perception: county maps to an admin2.
This might be different for counties in other countries.

Many other countries have a different official structures, e.g.
country > province > city > district

or sometimes a non-formal structure (dare I say, non-administrative) of
country > province > city > neighborhood.
This is not something you want to show as default hierarchy, but it is good to know that your neighborhood 'belongs' to a certain city.

The latter is what "hierarchies" wants to achieve.
The choice was at the time made to make the subtype based on the opinionated place type categorization of OSM.
At best these types give an idea of what a division represents, but it did not try to represent any administrative level let alone a hierarchical index level.

If 'the want' is to have administrative levels more clear, it feels like introducing a numeric 'level' property, could be more target leading.
This allows for more variations between countries while still keeping the approach globally consistent.

Right now I'm worried that it will not even fulfill worldwide that need, as there are some countries which have more levels between country and 'local_admin'.

vcschapp:
@jonahadkins Another interesting example is the Azure Maps search API because they actually have a "get the boundary polygon" API, [Get Polygon - Search](https://learn.microsoft.com/en-us/rest/api/maps/search/get-polygon?view=rest-maps-2025-01-01&tabs=HTTP). As far as I can tell, they also use a textual rather than a numeric system. Theirs is called [BoundaryResultType](https://learn.microsoft.com/en-us/rest/api/maps/search/get-polygon?view=rest-maps-2025-01-01&tabs=HTTP#boundaryresulttypeenum) and the allowed values are:

image

It's not super obvious what "country *or* region" means, but of note that they do use that term at the top level.

stepps00:
Closing, as #419 replaced this work.

PR #415, opened 2025-11-05
Pydantic package organization episode 5: "Only mostly dead"

vcschapp:
# tl;dr

This is part `#5`. Part `#6` will happen ... sometime, and it is hoped it will be the last, and the best, of the package re-orgs. Our last best hope, if you will.

Done in this part:

- The following themes are (mostly) fully re-organized: base, buildings (and places and addresses from part the fourth).
- Core package is partly re-organized and `json_schema` module is promoted to system.
- A number of bugs are fixed including several big GeoJSON-related Pydantic parsing issues and a subtle bug with the `json_schema` generation.
- Parsing modules are aligned more with Pydantic conventions, which is made possible by the more robust GeoJSON parsing coming from the bug fixes, especially the GeoJSON-aware discriminated union fix for `Feature`.
- Pydantic warnings are eliminated and asserted against in `make check`.
- Doc format errors are reduced to less than a terminal page and only errors that are in scope for episode 6 remain to be fixed.
- Various other clean-up.

# Merge guidance: Rebase

Unlike parts 1-3, this iteration has an adequate commit history. It should ideally be rebase-merged onto the `pydantic` branch rather than squashed.

# Journey

- Part 1 is in PR #397.
- Part 2 is in PR #400.
- Part 3 is in PR #403.
- Part 4 is in PR #412

# Destination

1. **Packages**.
   - ✅ ~~Package dependencies and `pyproject.toml` will be standardized.~~ Seth already did this in an earlier PR.
2. **Documentation**.
    - ▶️ `pydocstyle` will be enforced by `make check`.
    - ✖️ ~~`docformatter` will be run by `make check`.~~ Incompatible with `pydocstyle` and NumPy-style documentation.
3. **Modules**.
   - ▶️ In theme packages, each feature type will be within its own non-exported module within the package root.
   - ▶️ In theme packages, if a field contains a theme-specific "deep model" then it will be given its own module, *e.g.* `perspectives`.

# TODO list for episode 6

1. **Promote model discovery from core into system.**
2. Finish documenting `land` and `land_cover` in base based on discussion in `#working-group-schema`.
3. Re-organize divisions, transportation, and annex.
4. Move perspectives into scoping.
5. Finish re-organizing core, which is almost there.
6. Eliminate all `make docformat` errors and add `pydocstyle` into `make check`.
7. Add `make api-docs-live` and `make api-docs-html` targets to `Makefile`.
8. Add doctest examples in system package `__init__.py` for main `Feature` use cases:
   - Make a feature with constrained geometry and roundtrip GeoJSON
   - Make a discriminated union of features
9. Generify `FieldConstraint` so validate method can be type hint valid as much as possible, along the lines sketched below.

```python
from typing import TypeVar, Generic

T = TypeVar('T')

class Parent(Generic[T]):
    def method(self, x: T) -> None:
        ...

class ChildStr(Parent[str]):
    def method(self, x: str) -> None:  # ✓ OK - matches Parent[str]
        ...

class ChildInt(Parent[int]):
    def method(self, x: int) -> None:  # ✓ OK - matches Parent[int]
        ...
```


packages/overture-schema-base-theme/src/overture/schema/base/land.py
```diff
@@ -0,0 +1,150 @@
...
+    In Overture data releases, land features are sourced from OpenStreetMap. TODO. Finish this when
+    I get more info from Jennings.
+
+
+
+    Physical representations of land surfaces.
+
+    Global land derived from the inverse of OSM Coastlines. Translates `natural` tags from OpenStreetMap.
+
+    TODO: Update this description when the relationship to `land_cover` is better understood.
```
vcschapp, line 117: Coming back to finish the descriptions for `Land` and `LandCover` is on the TODO list for episode `#6` that's listed in the PR description....

packages/overture-schema-buildings-theme/src/overture/schema/buildings/building.py
```diff
@@ -0,0 +1,201 @@
...
+    Stacked,
+    Appearance,
+):
+    """
+    Buildings are man-made structures with roofs that exists permanently in one place.
+
+    A building's geometry represents the two-dimensional footprint of the building as viewed from
+    directly above, looking down. Fields such as `height` and `num_floors` allow the
+    three-dimensional shape to be approximated. Some buildings, identified by the `has_parts` field,
+    have associated `BuildingPart` features which can be used to generate a much more realistic 3D
```
vcschapp: The words "much" and "3D" are questionable here. Should I delete "much" and change "3D" to "2.5D"?

packages/overture-schema-core/tests/test_models.py
```diff
@@ -434,61 +436,77 @@ def test_feature_json_schema() -> None:
...
+    ),
 )
-def test_feature_json(feature: OvertureFeature, expect: dict[str, Any]) -> None:
-    assert feature.model_dump(mode="json") == expect
+
+
+@pytest.mark.parametrize("feature, geo_json", FEATURE_WITH_GEO_JSON)
+def test_feature_dump_json(feature: OvertureFeature, geo_json: dict[str, Any]) -> None:
+    """Ensure GeoJSON serialization inherited from `Feature` continues to work correctly."""
+    assert feature.model_dump(mode="json") == geo_json
```
vcschapp, line 497: This is intended to prevent regression of the bug fix in the commit [Fix bug: OvertureFeature overrode and broke Feature GeoJSON parse 🦟](https://github.com/OvertureMaps/schema/pull/415/commits/8e7cbf119e7b300d6885f473fd688d7453f7f970).

packages/overture-schema-divisions-theme/src/overture/schema/divisions/enums.py
```diff
@@ -4,8 +4,10 @@
 
 
 class PlaceType(str, Enum):
```
vcschapp, line 6: NB: To be renamed in the divisions package re-org. to remove the prefix "Place" to prevent any confusion with the places theme. The historical tie to WOF will still be documented obviously.

packages/overture-schema-system/src/overture/schema/system/__init__.py
```diff
@@ -98,7 +100,7 @@
 >>> try:
 ...    MyModel(**{"osm_id": "foo"})
 ... except ValidationError as e:
-...    assert "Invalid OSM ID format: foo. Must be n123, w123, or r123." in str(e)
```
vcschapp, line 101: I'm trying to standardize all exception messages to begin with lowercase.

---

I'm basing this convention on multiple (hopefully unbiased) LLM prompts to different models that basically land on this as the exception message convention. (The below summary is from ChatGPT, which claims it has examined the Python standard library in generalizing them.)

| Element | Convention | Example |
| -- | -- | -- |
| Capitalization | lowercase unless proper noun/type | `"invalid configuration option 'mode'"` |
| Ending punctuation | no period | `"missing required field 'name'"` |
| Formatting | short, plain English phrase | `"expected int, got str"` |
| Including data | quote or `repr()` values | `"invalid value: {!r}".format(val)` |

packages/overture-schema-system/src/overture/schema/system/feature.py
```diff
@@ -84,6 +90,82 @@ class Feature(BaseModel):
     ...     ]
 
     .. _GeoJSON format: https://datatracker.ietf.org/doc/html/rfc7946
+
```
vcschapp, line 93: 👉 This block and its example tests are the best summary of the new GeoJSON-aware dynamic discriminator feature.

Implementation is in the `field_discriminator` method. Usage is in the `__init__.py` from "schema" package and the `segment/models.py` package from "transportation" package.

packages/overture-schema-transportation-theme/src/overture/schema/transportation/segment/models.py
```diff
@@ -127,7 +127,14 @@ class WaterSegment(TransportationSegment):
 
 
 Segment = Annotated[
-    RoadSegment | RailSegment | WaterSegment, Field(discriminator="subtype")
+    Annotated[RoadSegment, Tag(Subtype.ROAD)]
```
vcschapp, line 130: 👉 Usage of GeoJSON-aware dynamic discriminator.

packages/overture-schema-system/src/overture/schema/system/primitive/bbox.py
```diff
@@ -105,28 +109,48 @@ def __str__(self) -> str:
     @property
     def xmin(self) -> float | int:
         """
-        float | int: Minimum X-coordinate
+        Minimum X-coordinate of the bounding box.
+
+        Returns
+        -------
+        float | int
+            Minimum X-coordinate
```
vcschapp, line 117: Correct NumPy-style documentation that `pydocstyle` approves of.

Big favor ask! Please use `make docformat` to make sure your changes don't regress the `pydocstyle` warnings and you are hopefully killing all the warnings in files you're touching!

packages/overture-schema/src/overture/schema/__init__.py
```diff
@@ -1,76 +1,171 @@
...
-    return parse_feature(feature, model_union, mode)
+        return Annotated[
+            reduce(
+                or_,
+                (
+                    Annotated[f, Tag(cast(str, _typeliteral(f)))]
+                    for f in feature_classes
+                ),
+            ),
+            Field(discriminator=Feature.field_discriminator("type", *feature_classes)),
```
vcschapp, line 123: 👉 The other use of GeoJSON-aware dynamic discriminator.

packages/overture-schema/src/overture/schema/__init__.py
```diff
@@ -1,76 +1,171 @@
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
vcschapp, line 1: Calling this file out because:

1. It is a significant re-imagining of the "parsing features" concept.
2. There's some overlap with the peripheries of https://github.com/OvertureMaps/schema/pull/406.

---

The path to get here wasn't entirely linear, but if we pretend it was then I asked myself:

1. What do `parse` and `parse_feature` do?
2. What value do they add above the parsing features Pydantic already has?

My answers were basically that **if our models integrate properly with Pydantic** then:

1. `parse_feature` does not add any real value because the same results can be obtained from calling `BaseModel.model_validate` and `BaseModel.model_validate_json` on the feature class.
2. The value added by `parse` is  that it can see the union of all discovered Overture models, so it can validate against the entire union.†

My actions then were to delete `parse_feature` and to re-imagine the interface of `parse` to:

1. Be consistent as possible with Pydantic conventions. This is where the rename from `parse` to `validate*` comes from, as well as the split between `validate` and `validate_json` and the return value being clearly expressed as `BaseModel`.
2. Leverage existing Pydantic features, such as the ability to validate a model from raw JSON, and the ability to validate a Python data structure, either a `dict` or a model class, against a model.

Some of the effects of this re-imagining are:

- The entire "GeoJSON rotation" of a feature is now completely encapsulated within `Feature` in the system package, and, I hope, it will not leak out of it.
- All GeoJSON validation is just JSON validation of a feature using `BaseModel.model_validate_json`.
- There is no longer any concept of "a GeoJSON `dict`" independent of a `Feature` but if someone wants that they just have to dump the feature with `BaseModel.model_dump_json` and parse the JSON into a `dict`.
- There may be some "technical" performance regressions in places because while Pydantic has all the facilities it needs, there edge cases where you need to go through serde with the `json` package to achieve your goal, as is always true with Pydantic. I think this is fine. I doubt JSON optimization is really needed because at-scale use cases shouldn't use JSON anyway, but if some optimized path is desired by someone we should evaluate the benefits against the cons of complexity and mission creep.

---

†: *I'm still not sold on validating the union being a feature that is particularly needed by customers or particularly scalable, as [some of my comments](https://github.com/OvertureMaps/schema/pull/406#issuecomment-3484213760) on #406 show, but that's a discussion for another day.*

packages/overture-schema-system/src/overture/schema/system/json_schema.py
```diff
@@ -0,0 +1,204 @@
...
+      },
+      "title": "FooModel",
+      "type": "object"
+    }
+
+
+    ⚠️ Warning
+    ----------
+    When using this class to generate a JSON Schema, you must dump your model JSON carefully to
+    ensure that it matches the JSON Schema.
```
vcschapp, line 46: The need for this warning is why I think `Omitable` is needed.

It's weird for us to say "Overture uses this omit convention but if you use our system you have to think carefully about matching your JSON Schema and JSON dumps of your model"...

packages/overture-schema-system/src/overture/schema/system/json_schema.py
```diff
@@ -0,0 +1,204 @@
...
+    ---------------
+    This class makes the following changes to the default Pydantic JSON Schema for nullable optional
+    fields only:
+
+    1. The JSON Schema type `null` is removed as an allowed type from the field's schema, which will
+       usually result in eliminating the `"anyOf"` composition keyword from the field's schema.
+    2. If the field's default value is `None` (JSON `null`), the default value is removed.
+    """
+
+    @override
```
vcschapp, line 118: This is a slight refactoring of how things were achieved in the previous `EnhancedJsonSchemaGenerator` but there's also a subtle bug fix. Previously, it would eliminate the `null` option from *required* nullable fields.

So if you had something along the lines of this:

```python
class FooModel(BaseModel):
    foo: str | Null   # This is a required nullable field, you can't leave it unset.
```

Then Pydantic would give you a JSON Schema like this:

```json
{
  "required": ["foo"],
  "properties": {
    "foo": {
      "anyOf": [
        {"type": "string"},
        {"type": "null"}
      ]
    }
  }
}
```

And `EnhancedJsonSchemaGenerator` would remove the `"anyOf"`/`null` option, which **changed the meaning of the JSON Schema** from "you have to put something, but it can be a string or the value `null`" to "you have to put a string".

---

Apart from the ⚠️ warning above, it's subtle issues like this that leave me preferring the one-shot `Omitable` option where the schema itself encodes the semantics so that our users don't have to think about these things.

vcschapp (commented):
Added callouts of some of the more interesting parts in comments, as is the tradition.

packages/overture-schema-base-theme/src/overture/schema/base/_common.py
```diff
@@ -0,0 +1,89 @@
...
+)
+
+
+class SourcedFromOpenStreetMap(BaseModel):
+    """
+    Model derived from an OpenStreetMap entity and containing the entity's OSM tags and wikidata ID.
+    """
+
+    source_tags: SourceTags | None = None
+    wikidata: WikidataId | None = None
```
sethfitz, line 60: Should this be split out? It may be present in the OSM tags already, so `wikidata` is either redundant or applies to more than data sourced from OSM.

packages/overture-schema-base-theme/src/overture/schema/base/_common.py
sethfitz, line 1: Why `_common` (private) vs. `common` and not exported at the top level?

packages/overture-schema-base-theme/src/overture/schema/base/infrastructure.py
```diff
@@ -191,3 +221,45 @@ class InfrastructureClass(str, Enum):
...
+    Named,
+    Stacked,
+    SourcedFromOpenStreetMap,
+):
+    """
+    Infrastructure features provide basic information about real-world infrastructure entitites
+    such as bridges, airports, runways, aerialways, communication towers, and power lines.
+    """
+
+    model_config = ConfigDict(title="infrastructure")
```
sethfitz, line 237: These are still messy. We don't strictly need them, but I ported them over directly from JSON Schema, and they don't carry meaning right now. The only place they manifest is as the title of the schema in the generated JSON Schema.

packages/overture-schema-base-theme/src/overture/schema/base/land_cover.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Land cover features indicate the primary natural or artificial surface material covering a land
+    area on the earth, including vegetation types like forests and crops, built environments like
+    cities, and natural surfaces like wetlands or barren ground.
+
+    Land cover features relate to `LandUse` features in the following way: land cover is the
+    physical thing covering the land, while land use is the human use to which the land is being
+    put.
+
+    TODO: Explain relationship to `Land` features.
```
sethfitz, line 48: Also include representative sources, like being derived from remote sensing data as [WorldCover](https://esa-worldcover.org/en) or [Global Dynamic Land Cover](https://land.copernicus.eu/en/products/global-dynamic-land-cover). Knowing that it's _often_ derived from raster data is valuable context.

packages/overture-schema-base-theme/src/overture/schema/base/water.py
```diff
@@ -0,0 +1,167 @@
...
+    OvertureFeature[Literal["base"], Literal["water"]],
+    Stacked,
+    Named,
+    SourcedFromOpenStreetMap,
+):
+    """
+    Water features represent ocean and inland water bodies.
+
+    In Overture data releases, water features are sourced from OpenStreetMap. There are two main
+    categories of water feature: ocean and inland water bodies.
```
sethfitz, line 97: Future TODO: we shouldn't assume that it comes from OSM and that it might be transformed data from USGS, for example.

packages/overture-schema-buildings-theme/src/overture/schema/buildings/building.py
sethfitz:
```suggestion
    have associated `BuildingPart` features which can be used to generate a more representative 3D
```

Because `BuildingPart`s can describe things like roof shape, it's not strictly extruded. I think 3D is fine.

packages/overture-schema-core/src/overture/schema/core/unit.py
```diff
@@ -1,42 +1,46 @@
...
 
     # Imperial units.
-    IN = "in"  # Imperial: Inch.
-    FT = "ft"  # Imperial: Foot.
-    YD = "yd"  # Imperial: Yard.
-    MI = "mi"  # Imperial: Mile.
+    IN = ("in", "One inch in the imperial and US customary systems")
+    FT = ("ft", "One foot in the imperial and US customary systems (12 inches)")
+    YD = ("yd", "One yard in the imperial and US customary systems (three feet)")
+    MI = ("mi", "One mile in the imperial and US customary systems (1,760 yards)")
```
sethfitz, line 24:
```suggestion
    MI = ("mi", "One mile in the imperial and US customary systems (5,280 feet)")
```

packages/overture-schema-system/src/overture/schema/system/feature.py
sethfitz: This took me a while to understand, and I've spent time with Pydantic discriminators (`pydantic.Tag` was new to me though, and key to understanding how this works).

An English explanation of what the code does would go a long way, IMO.

Here's an attempt:

Discriminated unions are the most effective way for Pydantic to determine which model matches a given input. These usually work by targeting a specific field, where the model definitions make it clear what value that field has (`Literal["<type>"]`). Another name for this is a tagged union. The implementation below uses `pydantic.Tag` to explicitly define the discriminator values. However, due to the different structure of the GeoJSON and tabular representations, the field containing these values or tags exists in a different location. `Feature.field_discriminator` solves this problem by ... (I haven't read the implementation yet).

packages/overture-schema-system/src/overture/schema/system/feature.py
```diff
@@ -100,12 +182,128 @@ class Feature(BaseModel):
...
+        def get_discriminator_value(data: object) -> Any:
+            # Pydantic doesn't have a facility to tell the dynamic discriminator function whether
+            # the context is 'python' or 'json', so we just have to use heuristics. If the input is
+            # a `dict` with the mandatory attributes `"type": "Feature"`, `"geometry"`, and
+            # `"properties"`, we assume we're in GeoJSON-land, otherwise not.
+            #
+            # If the data doesn't contain the discriminator field at all, we return `None` to tell
+            # Pydantic proceed to try the next variant in the union, if there is one. This is
+            # equivalent to how Pydantic behaves with static unions.
```
sethfitz, line 281: Except that it targets a specific field (that may exist in 2 places) rather than testing the whole model? Otherwise, what's the benefit?

packages/overture-schema-system/src/overture/schema/system/json_schema.py
sethfitz: Sure, but that still feels like a later problem if we find that lots of people are using this with JSON data + JSON Schema (vs. other approaches that are more in line with how our data pipeline is going to use it).

packages/overture-schema-system/src/overture/schema/system/json_schema.py
```diff
@@ -0,0 +1,204 @@
...
+      "title": "FooModel",
+      "type": "object"
+    }
+
+    This default Pydantic behavior is reasonable, and aligns well with the Python internal
+    representation. However, it doesn't take advantage of a quiet advantage of JSON and JSON Schema,
+    which is that there's another concept of optionality that does not require the value `null` at
+    all. In this paradigm, if the field is there, it has a set value; and if the field is missing,
+    it does not. Pydantic does enable this alternative paradigm via the experimental `MISSING`
+    sentinel, but this sentinel is not yet a full-fledged feature and has some rough edges.
```
sethfitz, line 102: If dealing with this properly is indeed a later problem, do you think `MISSING` will provide us with a different, working solution if we wait?

packages/overture-schema-system/src/overture/schema/system/json_schema.py
sethfitz: I'm not aware of any uses of `| Null`, nor do I expect Overture to need it.

Strictly speaking, you're right, but I lean toward YAGNI until we learn that we do (and documenting the limitations is a great first start).

packages/overture-schema-system/src/overture/schema/system/json_schema.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Returns
+    -------
+    JsonSchemaValue
+        JSON Schema for the model or union of models
+
+    Raises
+    ------
+    TypeError
+        If `models` is not a Pydantic model or union of Pydantic models
```
sethfitz, line 163:
```suggestion
        If `models` is not a Pydantic model or (potentially nested) union of Pydantic models
```

packages/overture-schema-system/src/overture/schema/system/json_schema.py
```diff
@@ -0,0 +1,204 @@
...
+                    )
+                )
+                return None if not redacted else {**schema, "schema": redacted}
+            case _:
+                return None
+
+
+def json_schema(thing: object) -> JsonSchemaValue:
+    """
+    Generate JSON Schema for a Pydantic model or union of models.
```
sethfitz, line 148: YAGNI, but if we're truly being thorough, shouldn't this include support for dataclasses?

packages/overture-schema-transportation-theme/src/overture/schema/transportation/segment/models.py
```diff
@@ -63,7 +64,6 @@ class TransportationSegment(
     connectors: Annotated[
         list[ConnectorReference] | None,
         Field(
-            default=[],
```
sethfitz, line 66: Good; this was weird (but matched the JSON Schema).

packages/overture-schema/src/overture/schema/__init__.py
sethfitz: I agree with this. I think it's safe to remove. The primary consumer is currently the validation tests, IIRC, and it can be moved there.

Since you fixed the "GeoJSON rotation," I also think it's valid to say that consumers (like the CLI package) should create their own `TypeAdapter`s (or use a public version of `_union_type_adapter`), since there's no magic needed any longer.

packages/overture-schema/src/overture/schema/__init__.py
sethfitz: I was hoping this would also eliminate the Overture-specific `type` field reference (by using the explicit `Tag`s and pushing `Feature.field_discriminator` into the base class implementation, currying the field name), but I'm not sure how we would do that. The primary motivation for doing that would be to support union creation in `system`.

packages/overture-schema/tests/test_schema_validation.py
```diff
@@ -223,84 +167,82 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
-    if parsed_feature is not None:
-        # Parsed feature should be in GeoJSON format, so compare directly
-        is_equal, diff_report = deep_compare_dicts(test_feature, parsed_feature)
-        assert is_equal, (
-            f"Parsed feature differs from original (geojson): {example_file}\n"
-            f"Differences:\n{diff_report}"
-        )
+    # If validation passed and we have a parsed feature, serialize to JSON and compare with the
+    # original JSON.
+    json_dump = model.model_dump(exclude_unset=True, by_alias=True, mode="json")
```
sethfitz, line 186: Looking at this here makes me wonder if we should implement `Feature.model_dump`, providing different defaults for `exclude_unset` and `by_alias`. That would eliminate some of the gotchas you described above.

packages/overture-schema/tests/test_schema_validation.py
```diff
@@ -1,11 +1,12 @@
+import json
 from collections.abc import Generator
 from pathlib import Path
 from typing import Any
 
 import pytest
 import yaml
-from deepdiff import DeepDiff
```
sethfitz, line 7: Without this, how's the diff output for failures?

Should we remove it from the list of dependencies too?

packages/overture-schema-base-theme/src/overture/schema/base/_common.py
vcschapp: They're all exported from `__init__.py`. I guess the answer is just that I find the proliferation of multiple import options confusing and sometimes its easier to have files for code organization but just a single flat import namespace for usability...

packages/overture-schema-base-theme/src/overture/schema/base/_common.py
vcschapp: Almost definitely yes. It just ended up below my subjective refactoring line.

packages/overture-schema-base-theme/src/overture/schema/base/infrastructure.py
vcschapp: Yeah. I haven't checked how consistently they're applied across the different packages either, I'll try to remember that for the next one. They should be everywhere or nowhere.

packages/overture-schema-base-theme/src/overture/schema/base/water.py
vcschapp: This is the problem with the "base" theme, which is kind of really the "straight from OSM" theme.

Fundamentally we need to promote water to a real theme with a schema that's not just OSM dependent.

packages/overture-schema/src/overture/schema/__init__.py
vcschapp: That would be powerful - we should noodle on how to achieve it. It would be ideal if all that logic could be lifted up into system...

PR #419, opened 2025-11-26
Add admin_level to divisions

stepps00:
Will replace https://github.com/OvertureMaps/schema/pull/414.

## A. Expected release date

Feb 2026

# Description

This change:
- adds a new `admin_level` field (int value) representing the division’s position in its country’s hierarchy
- sets that new field as required for top-level subtypes; country, macroregion, region, macrocounty, and county
- keeps the existing subtype schema in tact
- updates relevant examples and counterexamples
- regenerates the pydantic baseline schema files

# Checklist

1. [x] Add relevant examples.
6. [x] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation.

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/419)


schema/divisions/division.yaml
```diff
@@ -149,6 +149,12 @@ properties:     # JSON Schema: Top-level object properties.
           parent divisions.
         allOf:
           - "$ref": "../defs.yaml#/$defs/propertyDefinitions/id"
+      admin_level:
+        description:
+          Integer representing this division's position in its country's administrative
+          hierarchy. Lower numbers correspond to higher-level administrative units.
+        type: integer
+        minimum: 0
```
vcschapp: We're not really consistent in the schema (*especially* within the divisions theme IIRC) about whether we want to start numbered sequences at `0` or `1`.

My mental model is `0` is the programmer way, `1` is the regular human way. For some reason I tend toward thinking `1` is better as the starting point here, but I don't have a strong justification for it.

Why zero, or why one?

-----

This value being repeated in 3 places, it probably makes sense to define it reusably in `schema/divisions/defs.yaml` and reuse it.

The parallel structure in the Pydantic code base would be a `NewType` within the divisions package named `AdminLevel`.

----

Can we be a bit more precise in the documentation?

Here's a stab at it:

```
Integer representing this division's position in its country's administrative hierarchy, where
lower numbers correspond to higher level administrative units.

In Overture data releases, this value is equal to the number of ancestor features in the
division's primary hierarchy (the one defined by following `parent_division_id`). Thus, a
country always has an `admin_level` of 0, a region has an  `admin_level` of 1, while further
subdivisions have `admin_level` values greater than 1.
```

schema/divisions/division_boundary.yaml
```diff
@@ -92,6 +92,12 @@ properties:     # JSON Schema: Top-level object properties.
           present on boundaries between different principal subdivisions or countries.
         allOf:
           - "$ref": "../defs.yaml#/$defs/propertyDefinitions/iso3166_2SubdivisionCode"
+      admin_level:
```
vcschapp, line 108: I had to think about this for a second given that it occurs on the boundary feature type, but I recognize it is right: boundaries are only meaningful at "shared" levels of the administrative hierarchy.

vcschapp (changes requested):
I'm aligned, this seems right!

I'm clicking request changes because ... shouldn't `admin_level` be a `required` field on all three feature types? If it is fully derived from the level in the feature hierarchy, which I think it is, then there's no reason why every feature shouldn't have it is there?

schema/divisions/division.yaml
jonahadkins: we might have lost [some context from the previous pr](https://github.com/OvertureMaps/schema/pull/414#issuecomment-3552803481). the accepted standard admin 0/1/2/3. ideally we dont want to create or own standard that doesnt align. as noted, osm has their own which does not follow the majority of providers of boundary data

stepps00: > Why zero, or why one?

Yeah, I think this is answered in Jonah's comment, but we're really just borrowing from the industry standard admin0/admin1/admin2 convention here.

> This value being repeated in 3 places, it probably makes sense to define it reusably in schema/divisions/defs.yaml and reuse it.

A more recent commit does just this - replaces these three values with a new entry in the `defs.yaml` file.

> Can we be a bit more precise in the documentation?

Thanks for this! I've updated the definition of this new property using your suggestion, with a minor tweaks.

schema/divisions/division_boundary.yaml
stepps00: Yep, that's correct.

jonahadkins (dismissed):

sasastanojkov (dismissed):

stepps00:
> I'm clicking request changes because ... shouldn't admin_level be a required field on all three feature types? If it is fully derived from the level in the feature hierarchy, which I think it is, then there's no reason why every feature shouldn't have it is there?

@vcschapp after reviewing the divisions data, specifically the number of entries in the hierarchy for each country's set of subtypes, we should be confident in using the number of hierarchy entries to derive the `admin_level` value for the top-level admin layers - admin0, admin1, and admin2 (minus features in countries with `X*` country codes).

Of these 42,974 features at the admin0, admin1, and admin2 levels, only 15 have an unexpected number of entries in their hierarchies. Meaning we'd see a small error rate when using this to calculate the `admin_level` value for this set of features.

Beyond this (admin3, admin4, etc), there are various data inconsistencies that we'd need to fix before supporting this value on _every_ division feature. For now, the theme is interested in supporting those top-level `admin_level` additions (0-2). I think this means this pull request is okay as-is.. but happy to discuss further.

vcschapp:
> Of these 42,974 features at the admin0, admin1, and admin2 levels, only 15 have an unexpected number of entries in their hierarchies. Meaning we'd see a small error rate when using this to calculate the `admin_level` value for this set of features.

Is the conclusion that we will use the hierarchy level method to calculate `admin_level` and accept that 15 entries are misclassified?

packages/overture-schema-divisions-theme/src/overture/schema/divisions/division/models.py
```diff
@@ -117,6 +124,13 @@ class Division(
 The default parent division is the parent division as seen from the default political perspective, if there is one, and is otherwise chosen somewhat arbitrarily. The hierarchies property can be used to inspect the exhaustive list of parent divisions.""",
         ),
     ] = None
+    admin_level: Annotated[
+        int32 | None,
```
vcschapp: May I suggest marking this as `uint8`? There's no conceivable world where an `admin_level` exceeds 255.

---

Since this is used in 3 places, there best would be to follow the reuse pattern for the current state of this package, which is to define a `NewType` in `types.py` named `AdminLevel`. Then you can just import `AdminLevel` and reuse it in each place.

stepps00: The most recent commit resolves this, using a `uint8` type `AdminLevel` here.

vcschapp (dismissed):
```
                            ______________________________________________
                         .-'                     _                        '.
                       .'                       |-'                        |
                     .'                         |                          |
                  _.'               p         _\_/_         p              |
               _.'                  |       .'  |  '.       |              |
          __..'                     |      /    |    \      |              |
    ___..'                         .T\    ======+======    /T.             |
 ;;;\::::                        .' | \  /      |      \  / | '.           |
 ;;;|::::                      .'   |  \/       |       \/  |   '.         |
 ;;;/::::                    .'     |   \       |        \  |     '.       |
       ''.__               .'       |    \      |         \ |       '.     |
            ''._          <_________|_____>_____|__________>|_________>    |
                '._     (___________|___________|___________|___________)  |
                   '.    \;;;Dani;;;o;;;;;o;;;;;o;;;;;o;;;;;o;;;;;o;;;;/   |
                     '.~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~   |
                       '. ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  |
                         '-.______________________________________________.'
```

packages/overture-schema-divisions-theme/src/overture/schema/divisions/types.py
```diff
@@ -2,10 +2,14 @@
 
 from pydantic import Field
 
+from overture.schema.system.primitive import uint8
+
 from overture.schema.system.field_constraint import UniqueItemsConstraint
 
 from .models import HierarchyItem
 
+AdminLevel = NewType("AdminLevel", uint8)
```
vcschapp: You can actually add in all the field annotations.

```
AdminLevel = NewType(
    "AdminLevel",
    Annotated[
        uint8 | None,
        Field(
            description="Integer representing the division's position in its country's administrative hierarchy, where lower numbers correspond to higher level administrative units.",
        ),
    ]
```

stepps00: Great - taken care of in https://github.com/OvertureMaps/schema/pull/419/commits/eaeb1858630cc75aed41bb6c2bb5d0722fca625c

TristanDiet-TomTom (approved):

sasastanojkov (approved):

vcschapp:
This PR has received approvals from all 4 Steering Members. (Jonah approved but approval was washed out by one of the final tweak updates last night.)

Merging.

PR #420, opened 2025-12-01
Initial commit of new places taxonomy

jwass:

- Add new taxonomy column

TODO.

# Description

Adds the new categorization description for Overture's new taxonomy. This includes
specifying the primary category and the hierarchy from most general to most specific. The
schema includes the ability to specify alternates, but those are only the strings.

# Reference

1. TODO.

# Testing

TODO.

# Checklist

1. [x] Add relevant examples.

# Documentation website

### DIFF
--- a/packages/overture-schema-places-theme/tests/place_baseline_schema.json
+++ b/packages/overture-schema-places-theme/tests/place_baseline_schema.json
@@ -318,7 +318,7 @@
           "uniqueItems": true
         },
         "hierarchy": {
-          "description": "The full primary hierarchy of categories known for this place, ordered from most general to most specific.\nAn example hierarchy might be: `[\"food_and_drink\", \"restaurant\", \"casual_eatery\", \"gas_station_sushi\"].\n\nThe rightmost, or most specific, value in the `hierarchy` must always be equal to the `primary` field.\nThe basic level category of the place will typically be found in the middle of the primary hierarchy.\nThe primary hierarchy does not include any of the alternate categories found in the `alternates` field.",
+          "description": "The full primary hierarchy of categories known for this place, ordered from most general to most specific.\nAn example hierarchy might be: `[\"food_and_drink\", \"restaurant\", \"casual_eatery\", \"gas_station_sushi\"]`.\n\nThe rightmost, or most specific, value in the `hierarchy` must always be equal to the `primary` field.\nThe basic level category of the place will typically be found in the middle of the primary hierarchy.\nThe primary hierarchy does not include any of the alternate categories found in the `alternates` field.",
           "items": {
             "description": "Category in snake_case format",
             "pattern": "^[a-z0-9]+(_[a-z0-9]+)*$",
@@ -419,7 +419,7 @@
           "type": "array"
         },
         "basic_category": {
-          "description": "The basic level category of a place.\n\nThis field classifies places into categories at a level that most people find\nintuitive. The full list of possible values it may hold can be found at (TODO).\n\nThe basic level category, or simply basic category, is based on a cognitive science\nmodel use in taxonomy and ontology development. The idea is to provide the category\nname at the level of generality that is preferred by humans in learning and memory\ntasks. This category to be roughly in the middle of the general-to-specific category\nhierarchy.",
+          "description": "The basic level category of a place.\n\nThis field classifies places into categories at a level that most people find\nintuitive. The full list of possible values it may hold can be found at (TODO).\n\nThe basic level category, or simply basic category, is based on a cognitive science\nmodel use in taxonomy and ontology development. The idea is to provide the category\nname at the level of generality that is preferred by humans in learning and memory\ntasks. This category to be roughly in the middle of the general-to-specific category\nhierarchy.\n\nThe full list of basic level categories is available at https://docs.overturemaps.org/guides/places/",
           "pattern": "^[a-z0-9]+(_[a-z0-9]+)*$",
           "title": "Basic Category",
           "type": "string"


sasastanojkov (approved):

TristanDiet-TomTom (dismissed):

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,50 @@ class Categories(BaseModel):
...
+    ]
+
+    hierarchy: Annotated[
+        list[SnakeCaseString],
+        Field(
+            min_length=1,
+            description=textwrap.dedent("""
+                The full taxonomy path from the most general to most specific category,
+                represented as an ordered array (e.g., [food_and_drink, restaurant,
+                casual_eatery, gas_station_sushi]).
```
vcschapp: *Uber* nits that we're certainly not being fully consistent about anyway, but hey, it's what I do...

- s/array/list in the docstring
- I'd be tempted to put backticks around the brackets and double quotes around the strings so you get

```
`["food_and_drink", "restaurant", ...]`
```

---

How I would p~~y~~edantically write the docs, feel free to take/leave/ignore:

```
The full primary hierarchy of categories known for this place, ordered from most general to most specific.

An example hierarchy might be: `["food_and_drink", "restaurant", "casual_eatery", "gas_station_sushi"].

The rightmost, or most specific, value in the `hierarchy` must always be equal to the `primary` field. The basic level category of the place will typically be found in the middle of the primary hierarchy. The primary hierarchy does not include any of the alternate categories found in the `alternates` field.
```

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,50 @@ class Categories(BaseModel):
...
+
+    # Optional
+
+    alternates: Annotated[
+        list[SnakeCaseString] | None,
+        Field(
+            description=textwrap.dedent("""
+                Additional categories that apply to this place but are not the primary
+                classification. Useful for places that fit multiple categories (e.g., a
+                gas station that also serves sushi).
```
vcschapp: Is this a good example when the example for hierarchies already has `"gas_station_sushi"`?

---

One conceptual issue that I don't think is clear: these alternate categories must come from other branches of the taxonomy. Are they the always the "leaf" branches (most specific)? If yes, we should say that. If no, we should have some kind of rational explanation for what level of the tree is being selected unless it's totally random....

---

Vic's attempt to document, which you can feel free to ignore, and also the wrapping is surely all wrong:

```
Unordered list of additional categories that are known for this
place but that are not part of the primary category hierarchy.

Alternate categories allow a more complete picture of the place to
be surfaced when it fits multiple unconnected branches in the taxonomy, 
or example <here put an example that doesn't clash with the hierarchies
example!>.

This list must never contain any member of the primary category
hierarchy. Therefore, it must not overlap with the `primary` field,
the `hierarchy` field, or the basic level category. Note as well that
this field is an unordered list of extra categories and does not
represent a hierarchy.
```

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,50 @@ class Categories(BaseModel):
...
+
+    Provides the primary classification, full hierarchy path, and alternate categories.
+    """
+
+    # Required
+
+    primary: Annotated[
+        SnakeCaseString,
+        Field(
+            description="The most specific category for this place within the taxonomy hierarchy."
```
vcschapp: I'd be tempted to amend the docstring slightly to say:

```
The primary, or most specific, category known about this place.

The `primary` category value must always equal the last or rightmost entry in the `hierarchy` field.
```

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,50 @@ class Categories(BaseModel):
...
+        list[SnakeCaseString],
+        Field(
+            min_length=1,
+            description=textwrap.dedent("""
+                The full taxonomy path from the most general to most specific category,
+                represented as an ordered array (e.g., [food_and_drink, restaurant,
+                casual_eatery, gas_station_sushi]).
+            """).strip(),
+        ),
+    ]
```
vcschapp, line 120: This should also have a `UniqueItemsConstraint` shouldn't it?

schema/places/place.yaml
```diff
@@ -44,12 +44,39 @@ properties:
...
+          A new representation of the place's category within the Overture taxonomy.
+          Provides the primary classification, full hierarchy path, and alternate categories.
+        type: object
+        required: [primary, hierarchy]
+        properties:
+          primary:
+            description: >-
+              The most specific category for this place within the taxonomy hierarchy.
+            "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
+          hierarchy:
```
vcschapp, line 64: `hierarchy` should also have `uniqueItems: true` shouldn't it?

schema/places/place.yaml
```diff
@@ -44,12 +44,39 @@ properties:
...
+            description: >-
+              The full taxonomy path from the most general to most specific category,
+              represented as an ordered array (e.g., [food_and_drink, restaurant, casual_eatery, gas_station_sushi]).
+            type: array
+            items:
+              "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
+            minItems: 1
+          alternates:
```
vcschapp, line 73: IMO alternates should have `minItems: 1` - either omit it, or provide it and don't leave it empty.

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,50 @@ class Categories(BaseModel):
...
+
+    # Optional
+
+    alternates: Annotated[
+        list[SnakeCaseString] | None,
+        Field(
+            description=textwrap.dedent("""
```
vcschapp, line 128: Ideally this has `min_length=1` so you either have to omit it or you have to provide at least one alternate value...

examples/places/place-with-taxonomy.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - casual_eatery
+      - gas_station_sushi
+    alternates:
+      - gas_station
+      - sushi_restaurant
+  categories:
+    primary: gas_station_sushi
+    alternates:
+      - gas_station
+      - sushi_restaurant
```
vcschapp, line 23: This example makes it super clear that apart from clearer documentation the main thing we're adding is the `hierarchy` - everything else was already there...

I'm curious then: was not having the primary hierarchy considered a major usability obstacle?

counterexamples/places/bad-taxonomy-missing-hierarchy.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
vcschapp, line 1: One sad and unfortunate aspect of the temporary dual regime of YAML JSON Schema and Pydantic is that we actually have to put the examples/counterexamples in two places.

I don't exactly remember why, @mojodna had to do this to make the examples/counterexamples work on Pydantic. But ... the short story is there's a `references/` directory that contains this second version. Would you mind duplicating the examples/counterexamples there and making sure they pass `$ make check`? 😬

examples/places/place-with-taxonomy.yaml
jwass: Actually that's a very good point. I think the main thing is that we're changing the categories themselves and in the meantime need to have both. The example here is bad because what will really happen is that categories.primary will be different than taxonomy.primary. And same for alternates.

So the structure of the two in the schema are the same with the exception of the new hierarchies field, but the values will be very different and we will have this period of time where they both need to exist because we don't want to flip a switch on all categories for all places.

counterexamples/places/bad-taxonomy-missing-hierarchy.yaml
jwass: Done.

schema/places/place.yaml
```diff
@@ -44,12 +44,41 @@ properties:
       basic_category:
         description: >-
           The basic level category of a place. At present this is a mapping of the categories.primary entry to a new, simplified name. This mapping can be 1:1 or M:1 from the
-          existing categories.primary entry. If the entry is currently empty in categories.primary, this entry will be empty. 
+          existing categories.primary entry. If the entry is currently empty in categories.primary, this entry will be empty.
           This type of categorization is a cognitive science model that is relevant for taxonomy and ontology development that shows the most broadest and general
-          category name that is most often found in the middle of a general-to-specific hierarchy, with generalization proceeding upward and specialization proceeding downward. 
+          category name that is most often found in the middle of a general-to-specific hierarchy, with generalization proceeding upward and specialization proceeding downward.
           The full list of basic level categories is available at:(todo)
```
danabauer: We need to add a link to this list of basic categories before we merge this PR. It can live in the [places data guide](https://docs.overturemaps.org/guides/places/) for now.

schema/places/place.yaml
danabauer, line 1: Is the link to the full list of "old" categories still accurate? It would more helpful to users to link to an explanation of the changes underway and a comparison of old categories to the new categories we're rolling out in the taxonomy property

```
    properties:
      categories:
        description: |
          The categories of the place. Complete list is available on
          GitHub: https://github.com/OvertureMaps/schema/blob/main/docs/schema/concepts/by-theme/places/overture_categories.csv
```

schema/places/place.yaml
```diff
@@ -44,12 +44,41 @@ properties:
...
         type: string
         "$ref": "./defs.yaml#/$defs/typeDefinitions/category"
+      taxonomy:
+        description: >-
```
danabauer, line 54: We need to link to more comprehensive documentation of the taxonomy work and a comparison of old categories to new categories. This can live in the [places guide](https://docs.overturemaps.org/guides/places/) for now.

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,72 @@ class Categories(BaseModel):
     ] = None
 
 
+@no_extra_fields
+class Taxonomy(BaseModel):
```
danabauer, line 86: see notes about necessary documentation in the yaml/json version of the schema

schema/places/place.yaml
jwass: Is that list even finalized at this point? @DrishT

jwass: Why is this a requirement for this PR which does not introduce basic category? It has been in the schema since #401 and #399

counterexamples/places/bad-taxonomy-missing-hierarchy.yaml
sethfitz: The "why" is that error paths + messages differ when validated by `jv` vs. Pydantic. Secondarily, Pydantic is a bit more pedantic about URLs. For the most part you should be able to just copy them into both locations.

schema/places/place.yaml
danabauer: All I'm asking is that you drop a link to the places guide for more information. https://docs.overturemaps.org/guides/places/

You can write: "Please see the places guide for more information."

packages/overture-schema-places-theme/src/overture/schema/places/place.py
```diff
@@ -82,6 +82,72 @@ class Categories(BaseModel):
...
+        ),
+    ]
+
+    hierarchy: Annotated[
+        list[SnakeCaseString],
+        Field(
+            min_length=1,
+            description=textwrap.dedent("""
+                The full primary hierarchy of categories known for this place, ordered from most general to most specific.
+                An example hierarchy might be: `["food_and_drink", "restaurant", "casual_eatery", "gas_station_sushi"].
```
vcschapp: Oh noes, missing backtick.

schema/places/place.yaml
vcschapp: Docs updated.

examples/places/place-with-taxonomy.yaml
vcschapp: Got it. I now have the whole picture in my head, at least temporarily!

Hopefully our messaging on this is good :)

PR #426, opened 2026-01-19
Fix ID typo in address examples

bact:
Make the IDs consistent with the IDs in other address examples.


vcschapp (approved):
Thanks @bact.

danabauer:
Good catch! Thanks, @bact

danabauer (approved):

PR #439, opened 2026-02-11
Fix mypy errors in test files

sethfitz:
## Summary

- Fix mypy errors across test subdirectories in both `overture-schema-core` and `overture-schema-system` packages
- Update `Makefile` mypy target to check test files recursively (`find` instead of glob) so subdirectories like `tests/model_constraint/` and `tests/scoping/` are covered

## Changes

- Add `# type: ignore[attr-defined]` for dynamically-generated `When` inner class access in scoping tests
- Add return type annotations to test functions missing `-> None`
- Fix type annotations: `dict` → `Mapping` for read-only params, `JsonDict` for Pydantic schema dicts, `bool | None` where `None` default was used
- Rename duplicate `TestModel` classes within the same module to satisfy mypy's no-redef rule
- Use regex patterns in `pytest.raises(match=...)` where class names vary across redefined test models
- Replace flat `packages/*/tests/*.py` glob in Makefile with recursive `find` to catch all test files

## Test plan

- [x] `make mypy` passes with no errors
- [x] `make test` still passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ mypy: uv-sync
 		| tr - . \
 		| sed 's|^packages/|-p |' \
 		| xargs uv run mypy --no-error-summary
-	@uv run mypy --no-error-summary packages/*/tests/*.py
+	@for d in packages/*/tests; do find "$$d" -name "*.py" | sort | xargs uv run mypy --no-error-summary || exit 1; done
 
 reset-baseline-schemas:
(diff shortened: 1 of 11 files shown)


vcschapp (approved):

RoelBollens-TomTom (approved):

jwass (approved):
lgtm

PR #440, opened 2026-02-11
Tags: extensible model grouping

sethfitz:
Draft design for replacing the hardcoded `namespace` concept with tags -- string labels declared by package authors and derived by tag providers.

Motivating goals:

- Move `discover_models` from `core` into `system`
- Support richer model categorization and filtering for CLI, code generation, and JSON Schema output
- Simplify the extension mechanism

Design doc: `docs/designs/tags.md`

### DIFF
--- a/docs/designs/tags.md
+++ b/docs/designs/tags.md
@@ -5,39 +5,57 @@ string labels declared by package authors and derived by tag providers. Tags
 become the filtering and grouping mechanism for model discovery, driven by
 package authors rather than central coordination.
 
-Theme remains a first-class structural field. It maps to data partitioning and
-entry point naming, distinct from tags which are descriptive metadata.
+Theme becomes a tag (`overture:theme=buildings`), not a structural field.
+`system` provides generic key-based grouping without understanding what
+"theme" means. Any package can declare theme tags (or any other structured
+tag) without special support in the discovery layer.
 
-## Phase 1: Basic Tagging
+## Phase 1: Structured Tags
 
 Replace namespace with tags, move discovery to `system`, update CLI.
 
+### Tag Format
+
+Tags are strings following the pattern `[namespace:]key[=value]`:
+
+- **Plain**: `overture`, `draft`
+- **Namespaced**: `system:extension` -- `:` separates ownership/reservation
+- **Namespaced k/v**: `overture:theme=buildings`, `codegen:schema_root=overture.schema`
+
+`:` signals ownership/reservation. `=` signals a dimension with a value
+(groupable via `--group-by`). One level of each -- no nested colons or
+multiple `=` signs.
+
 ### Data Model
 
-`ModelKey` drops `namespace`, gains `tags`:
+`ModelKey` drops `namespace` and `theme`, gains `name`:
 
 ```python
 @dataclass(frozen=True, slots=True)
 class ModelKey:
-    theme: str | None
-    type: str
-    tags: frozenset[str]
-    class_name: str
+    name: str              # friendly name from entry point key
+    class_name: str        # entry point value (module:Class)
+    tags: frozenset[str]   # plain and structured tags
 ```
+
+No `namespace` field, no `theme` field. Both are tags.
+
+`name` is the entry point key (before any `#` suffix). It serves as the
+model's friendly identifier in CLI output and codegen path generation.
 
 ### Entry Point Format
 
-Names change from `namespace:theme:type` to `theme:type` (or just `type` for
-non-themed models), with optional `#tag1,tag2` suffix. The group remains
+Entry point keys are `name` or `name#tag1,tag2`. No `theme:type` prefix --
+theme is a package-level keyword. The entry point group remains
 `overture.models`.
 
 ```toml
 [project]
-keywords = ["some-org-tag"]
+keywords = ["overture:theme=buildings"]
 
 [project.entry-points."overture.models"]
-"buildings:building" = "overture.schema.buildings:Building"
-"buildings:building_part#draft" = "overture.schema.buildings:BuildingPart"
+building = "overture.schema.buildings:Building"
+"building_part#draft" = "overture.schema.buildings:BuildingPart"
 ```
 
 A model's tags are the union of:
@@ -45,6 +63,7 @@ A model's tags are the union of:
 - `[project].keywords` from the distribution metadata (read via
   `entry_point.dist.metadata["Keywords"]`)
 - Per-model `#` tags from the entry point name
+- Tag providers (Phase 2)
 
 Note: `[project].keywords` is also used for PyPI search, so schema tags and
 PyPI keywords share a namespace. This was considered and accepted -- in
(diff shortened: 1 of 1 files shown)


vcschapp:
Before reading this (to avoid any kind of subtle anchoring bias), I want to make sure I channel the ideas I was taking out of yesterday's conversation.

# tl;dr

1. System package does discovery and assigns tags.
2. The following tags are reserved:
    1. `feature` - Automatically assigned by the system package if the model derives from `Feature` (not `OvertureFeature`).
    2. `extension` - Automatically assigned by the system package if the model meets the definition of extension.
    3. `attributes` (?) - Automatically assigned by the system package to non-models (???).
    4. `overture` - Automatically assigned by the system package to models discovered from a hard-coded list of designated Overture packages that lives in `system`.
3. System package provides a default tagger. Maybe it can be overridden, or maybe it is always used.
4. Additional taggers can be discovered via entry-points.
5. We will export Overture-flavored tagger(s):
   - OPTION 1. In the `core` package. This tagger understands `OvertureFeature`, and uses the `theme` property of a model to add a theme-based tag.
   - OPTION 2. In each theme package. Each such tagger just knows what the package exports and slaps the theme-based tag onto them.

# CLI interface

Short-form list (default). Basic FQCN.

```console
$ overture-schema list
overture.schema.buildings.Building
overture.schema.buildings.BuildingPart
overture.schema.divisions.Divisions
overture.schema.divisions.DivisionArea
...
overture.schema.ext.places.OpeningHours
overture.schema.ext.transportation.Lanes
...
foo.overture.bar.GeographicalArea
```

Note that if the entry-points specify a "friendly name", *e.g.* `building`, then the output becomes nicer.

```console
$ overture-schema list
address
connector
building
building_part
division
division_area
place
segment
```

Long-form list. One-liner table in a nicer format with tags.

```console
$ overture-schema list -l
MODULE                             CLASS             TAGS
overture.schema.buildings          Building          [overture] [feature] [buildings]
overture.schema.buildings          BuildingPart      [overture] [feature] [buildings]
overture.schema.divisions          Division          [overture] [feature] [buildings]
overture.schema.divisions          DivisionArea      [overture] [feature] [buildings]
...
overture.schema.ext.places         OpeningHours      [overture] [extension] [attributes] [places]
overture.schema.ext.transportation Lanes             [overture] [extension] [attributes] [transportation]
...
foo.overture.bar                   GeographicalArea  [community] [extension] [feature] [divisions]
```

Long-form list. One-liner table in a nicer format with tags.

```console
$ overture-schema list -l
MODULE                             CLASS             TAGS
overture.schema.buildings          Building          [overture] [feature] [buildings]
overture.schema.buildings          BuildingPart      [overture] [feature] [buildings]
overture.schema.divisions          Division          [overture] [feature] [buildings]
overture.schema.divisions          DivisionArea      [overture] [feature] [buildings]
...
overture.schema.ext.places         OpeningHours      [overture] [extension] [attributes] [places]
overture.schema.ext.transportation Lanes             [overture] [extension] [attributes] [transportation]
...
foo.overture.bar                   GeographicalArea  [community] [extension] [feature] [divisions]
```

Again, if we have the friendly name available from the entry-point maybe there's something nicer here:

```console
$ overture-schema list -l
MODEL              TAGS                                                  DESCRIPTION
building           [overture] [feature] [buildings]                      Man-made structure with a roof existing permanently ...
building_part      [overture] [feature] [buildings]
division           [overture] [feature] [buildings]
division_area      [overture] [feature] [buildings]
...
opening_hours      [overture] [extension] [attributes] [places]
lanes              [overture] [extension] [attributes] [transportation]
...
geographical_area  [community] [extension] [feature] [divisions]
```

Long-form list filtering on tags.

```console
$ overture-schema list -l --tag community --tag extension
MODULE                             CLASS             TAGS
foo.overture.bar                   GeographicalArea  [community] [extension] [feature] [divisions]
```

Full detail, filtering on tags.

```console
$ overture-schema document --tag overture --tag buildings
BUILDING
========
overture.schema.buildings.Building                                       [overture] [feature] [buildings]

Buildings are man-made structures with roofs that exist permanently in one place.

A building's geometry represents the two-dimensional footprint of the building as viewed from
directly above, looking down. Fields such as `height` and `num_floors` allow the three-dimensional
shape to be approximated. Some buildings, identified by the `has_parts` field, have associated
`BuildingPart` features which can be used to generate a more representative 3D model of the building.

<Maybe down here you list the feature geometry type and properties also?>

BUILDING PART
==============
overture.schema.buildings.BuildingPart                                   [overture] [feature] [buildings]

...
```

# Open issues

## Union models

Do we really need to or want to create **union models**, [as is currently done in the `overture-schema` package](https://github.com/OvertureMaps/schema/blob/dev/packages/overture-schema/src/overture/schema/__init__.py#L71) and perhaps other places also.

This is needed to supply a "blindly validate a blob" CLI command, like:

```bash
$ overture-schema validate-anything <anything.json
```

Personally, I think that the **complexity** we're taking on by doing this is **high** and the demonstrated **necessity** of this use case is **low**. Users can instead just specify the model to validate against:

```bash
$ overture-schema validate --type building <building.json
```

IMO we should remove the tagged union and associated behavior for now and tackle this issue in the future, when it bubbles up from P2+ to P0.

docs/designs/tags.md
```diff
@@ -0,0 +1,378 @@
...
+def approved_provider(
+    model_class: type[BaseModel], key: ModelKey, tags: set[str]
+) -> set[str]:
+    if _matches_manifest(key):
+        tags.add("approved")
+    return tags
+```
+
+Note: `approved` is an unprefixed tag (not `system:approved`) since this
+provider lives outside `system`.
```
danabauer, line 585: why does `approved` live outside of `system`?

sethfitz: It needs to be in `overture-schema-core` or downstream because it's an Overture concern (how we deem extensions or schemas as vetted) rather than `overture-schema-system` ("George", the parts of the schema system that we believe aren't specific to Overture).

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+    "feature": {"overture-schema-system"},
+}
+```
+
+Same enforcement: only authorized packages may produce these tags via tag
+providers. Unauthorized providers that attempt to add reserved tags see a
+warning and discard.
+
+This is the less-preferred mechanism. Reserving individual strings does not
+scale -- every new reserved word requires an update to the table. Prefix
```
vcschapp, line 153: > Reserving individual strings does not scale

Implicit question here: how much reserved-ness do we need?

Is reserved a thin bootstrapping slice or potentially a very big group?

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+### Why Prefix-Based Reservation
+
+The alternative is flat reservation only -- reserve specific strings
+(`overture`, `feature`, `extension`, `buildings`, `theme`) without prefix
+semantics. Flat tags keep the `key=value` dimension, and they're shorter.
+But making the prefix opaque has costs:
+
+**Structured tags encode relationships that flat tags cannot.** With
+`overture:theme=buildings`, `--group-by overture:theme` extracts the
+dimension and groups by its values. With flat tags `theme` and `buildings`,
```
vcschapp, line 168: I am arguing that there's no tag for theme. There's just `buildings`. Anyone can apply `buildings`, but what makes it the Overture buildings theme is that there's an `overture` tag on it.

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+grouping requires sniffing for pairings -- "does this model have both
+`theme` and `buildings`?" -- and the pairing is implicit. A model tagged
+`theme`, `buildings`, `residential` is ambiguous: are `buildings` and
+`residential` both theme values, or is `residential` a separate
+classification?
```
vcschapp, line 173: This is inherent in any tagging situation. Anytime multiple tags can be added, you can get ambiguity.

Unless you introduce a rigorous set of requirements of what tags you're allowed to add, which then becomes something other than a tagging system.

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+
+**Tags serve both endorsement and discovery.** A flat `buildings` tag in a
+marketplace context is ambiguous: does it mean "this package relates to
+buildings" (taxonomy) or "this package is the endorsed buildings schema"
+(endorsement)? `overture:theme=buildings` unambiguously signals taxonomy;
```
vcschapp, line 178: The same is true of `overture` (reserved tag) + `buildings`.

As well, why shouldn't an 3P extension be allowed to say "this extension is part of the buildings theme" where the only difference is w the 3P extension, it's not `overture` since that official tag is reserved?

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+endorsement lives in a separate tag (`approved`).
+
+**Multi-ecosystem support requires disambiguation.** When multiple
+ecosystems share the discovery mechanism, prefixes prevent collision:
+`overture:theme=buildings` and `acme:theme=industrial` coexist cleanly.
+Flat `buildings` and `industrial` need external context to distinguish
```
vcschapp, line 184: The difference would be the `overture` tag is on the first but not second.

docs/designs/tags.md
```diff
@@ -0,0 +1,667 @@
...
+their purpose.
+
+**`system` benefits from prefixed tags internally.** `system:extension`
```
vcschapp, line 187: Totally subjective but having to say give me `system:extension` to figure out if something is an extension instead of just `extension` seems clunky to me.

vcschapp:
@sethfitz any objection to closing this as redundant with https://github.com/OvertureMaps/schema/pull/463?

sethfitz:
Yeah, I'll close it. Even with the branch deleted (for purposes of browsing + pulling), I think we're still able to see the changes from that branch here, which is the thing that's useful for posterity. (I also kept a copy of the design locally in case I/we want to revisit it at any point.)

PR #441, opened 2026-02-12
Attach docstrings to NewTypes at runtime

sethfitz:
## Summary

- Convert bare triple-quoted strings after `NewType` assignments into explicit `__doc__` assignments so docstrings are accessible at runtime
- Python treats post-assignment docstrings as expression statements, never attaching them to the `NewType` object (`__doc__` stays `None`)
- Same pattern `DocumentedEnum` uses for enum member docs

## Affected packages

- `overture-schema-system` (primitives, strings, ref/id)
- `overture-schema-core` (names, opening_hours)

### DIFF
--- a/packages/overture-schema-core/src/overture/schema/core/names.py
+++ b/packages/overture-schema-core/src/overture/schema/core/names.py
@@ -110,7 +110,7 @@ CommonNames = NewType(
         Field(json_schema_extra={"additionalProperties": False}),
     ],
 )
-"""A mapping from language to the most commonly used or recognized name in that language."""
+CommonNames.__doc__ = """A mapping from language to the most commonly used or recognized name in that language."""
 
 
 class NameVariant(str, DocumentedEnum):
--- a/packages/overture-schema-core/src/overture/schema/core/scoping/opening_hours.py
+++ b/packages/overture-schema-core/src/overture/schema/core/scoping/opening_hours.py
@@ -15,7 +15,7 @@ OpeningHours = NewType(
         ),
     ],
 )
-"""
+OpeningHours.__doc__ = """
 Time span or time spans during which something is open or active, specified in the OpenStreetMap
 opening hours specification: https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification.
 """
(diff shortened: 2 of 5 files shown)


vcschapp:
Can we talk about this? It breaks the documentation generator (`pdoc` IIRC), which has a special hack mode that recognizes docstrings coming after variables, but it doesn't support reading `__doc__` on those values.

sethfitz:
Interesting... I had to do this to support Markdown/code generation because the docstrings that come after the `NewType` definitions weren't accessible programmatically (the opposite problem). It'd be interesting to see what `proc`'s workaround is in case that's appropriate to incorporate in my work.

RoelBollens-TomTom:
I also ran into the same issue where rather than explicit `__doc__` field assignment I used the `Field(description="...")` instead. This still doesn't help with the `pdoc` documentation generation but wanted to mention it as there are models which currently document their fields in this way.

An other related issue I've come across has been with the Segment alias which is also not know anymore at runtime which I had solved using `Field(title="...")`.

jenningsanderson (approved):
👍 I think optimizing for markdown code generation here is _okay_ for the moment, as this is what's required for our reference docs. We can/should revisit in the future for other documentation generators.

vcschapp:
> 👍 I think optimizing for markdown code generation here is _okay_ for the moment, as this is what's required for our reference docs. We can/should revisit in the future for other documentation generators.

My issue is I sank *hours* into trying to wrangle the docs into shape, apply one consistent docs standard, and to find a Python docs tool that works and doesn't require a Ph.D in Sphinx. I don't want to just throw out all that effort and investigation and analysis and say "hey we're doing this thing now" without carefully thinking through the consequences.

The point I've been making for a while now is there are two different docs audiences

1. Schema enthusiasts (Docusaurus, Markdown).
2. Developers. 

Audience #2 isn't being served well right now. I'd like to find a solution that balances between the two.

vcschapp:
> Interesting... I had to do this to support Markdown/code generation because the docstrings that come after the `NewType` definitions weren't accessible programmatically (the opposite problem). It'd be interesting to see what `proc`'s workaround is in case that's appropriate to incorporate in my work.

This is one of the many annoying aspects of Python. 

- https://peps.python.org/pep-0224/ - *Attribute Docstrings* - Rejected
- https://peps.python.org/pep-0727/ - *Documentation in Annotated Metadata* - Withdrawn

One option we have is good 'ol duplication. Maybe we could build a linter that checks that the duplication is properly duplicated.

vcschapp (approved):
Let's merge this and return to the topic of developer docs/`pdoc` later.

Roel's also suggesting `mkdocs` as a `pdoc` alternative...

PR #442, opened 2026-02-12
Auto-publish Python packages to PyPI on main branch version change

vcschapp:
# Description

This change adds a GitHub Actions workflow that automatically publishes Python packages to PyPI. For now they go to the Overture internal PyPI repository on CodeArtifact, but eventually the workflows will be changed to publish to public PyPI. This is aligned with the design from the 2024-01-30 [Coding Sesh 💻](https://github.com/OvertureMaps/schema-wg/issues/406).

This change also fixes a bunch of bugs in the version check GitHub Action workflow. Essentially, the index URL was always incorrect, which made the workflow result meaningless.

What happens is this:

1. When you make a PR that changes Python package version numbers, the existing version check workflow (now fixed) runs to verify that the version numbers you picked don't already exist in PyPI.
2. When you merge your PR do main, the new publish workflow re-runs the version checks and, for any package with a changed version, publishes it PyPI.

# Future Work

In the future, the publish workflow should also automatically create tagged releases for each package being released. *e.g.*, if commit `badc0ded` contains version changes for `overture-schema-core` and `overture-schema-buildings-theme`, then it would tag it twice, like `core-v1.2.3` and `buildings-v2.3.4`.

To make this work, we'll need to invent a system to include per-version release notes in the commit. The version check workflow will need to be augmented check for valid release notes associated with the new version number, and the publish workflow will need to create the tag and release.

# Testing

Literally dozens of testing iterations pushing commits onto a branch, [pypi-auto-publish-gh-action](https://github.com/OvertureMaps/schema/tree/pypi-auto-publish-gh-action). Click the link to see the gory, awful, testing truth.


sethfitz (approved):
Impressive levels of (necessary) exploration!

sasastanojkov (approved):

PR #437, opened 2026-02-06
[WIP] Add PyArrow Parquet validation to CLI

Rachmanin0xFF:
# Description

Adds two commands to the CLI:
* `parquet-schema`: Generates an emptpy Parquet file with a specific type's schema.
* `validate-schema`: Validates the schema of a Parquet file (or hive-partitioned dataset) by comparing its schema to a PyArrow schema generated from Pydantic.

Adds `--theme` and `--simple` options to the `list-types` command (this makes it easier to programmatically interpret its output).

Additionally, changes all `int8` and `int16` types to `int32`. Parquet's compression is good enough that there are no differences in physical (Parquet) size between the types, and `int32`s are more universally digestible. 

This command will be used for schema validation in our release process; you can see how [here](https://github.com/OvertureMaps/tf-data-platform/pull/2767/changes#diff-e9144bab39041b4936f9e5a20e6d727eb6495d5d630f165e82793e18e7a611a7R7).

### Round-trip generate + validate
```bash
$ overture-schema parquet-schema --type building -o buildings.parquet && overture-schema validate-schema --theme buildings --type building buildings.parquet
Wrote Parquet schema to buildings.parquet
SUCCESS, schema of 'buildings.parquet' matches type 'building' (subset check)
$ echo $?
0
```

### Validation on release data
```bash
$ overture-schema validate-schema --theme addresses --type address s3://overturemaps-us-west-2/release/2026-01-21.0/theme=addresses/type=address/
FAILURE, schema mismatch: 
's3://overturemaps-us-west-2/release/2026-01-21.0/theme=addresses/type=address/' vs type 'address'
FAILURE, schema mismatch: 
's3://overturemaps-us-west-2/release/2026-01-21.0/theme=addresses/type=address/' vs type 'address'

Missing fields (2):
  - theme  (expected: string)
  - type  (expected: string)

Type mismatches (1):
  ~ sources.item.confidence
      expected: float
      actual:   double

$ echo $?
1
```

### Arrow schema text output (uses Arrow schema's `.to_string()`)
```bash
$ overture-schema parquet-schema --format text --type connector
id: string not null
  -- field metadata --
  description: 'A feature ID. This may be an ID associated with the Globa' + 104
bbox: struct<xmin: float, ymin: float, xmax: float, ymax: float>
  child 0, xmin: float
  child 1, ymin: float
  child 2, xmax: float
  child 3, ymax: float
  -- field metadata --
  description: 'An optional bounding box for the feature'
geometry: binary not null
[...]

PR #445, opened 2026-02-17
Widen transportation, cartography, confidence types to match data

sethfitz:
## Summary

- Vehicle dimension selectors (height, length, weight, width) widened from `float32` to `float64` to match double-precision values in the data platform
- `Level` type widened from `int16` to `int32` for the same reason
- `VehicleAxleCountSelector` stays `uint8` (discrete count)

Closes OvertureMaps/tf-data-platform#2829

## Test plan

- [x] `make check` passes (1612 tests, mypy, ruff)
- [x] Verify generated Parquet schema matches updated types


Rachmanin0xFF (approved):
The type expansions are at parity with the schema changes in https://github.com/OvertureMaps/schema/pull/437, which passes validation.

vcschapp (approved):

PR #449, opened 2026-02-23
Load CLI subcommands from entry points

sethfitz:
## Summary

Refactor the CLI so installed packages contribute subcommands via setuptools entry points, then extract `validate` into a standalone `overture-schema-validation` package to prove the mechanism.

- The `cli()` group discovers subcommands from the `overture.schema.cli` entry point group at import time
- `overture-schema-validation` registers the `validate` subcommand through an entry point -- no CLI code changes needed
- Shared model-filtering and union-construction logic moves to `overture-schema-core` to avoid cross-dependencies

## Motivation

The CLI package was accumulating domain-specific command implementations as direct dependencies. Entry-point-based plugin loading inverts this: command packages depend on core or system, and the CLI discovers them at runtime. Adding a new subcommand (e.g. `generate`) means adding a package with an entry point -- no changes to the CLI.

### Entry point registration

Packages register subcommands by declaring an entry point in `pyproject.toml`:

```toml
[project.entry-points."overture.schema.cli"]
validate = "overture.schema.validation.commands:validate"
```

The value points to a Click command object. Built-in commands (`json-schema`, `list-types`) take precedence on name collision; broken plugins warn to stderr without crashing.

## Squash commit message

```
Load CLI subcommands from entry points

The cli() group discovers subcommands from the
"overture.schema.cli" entry point group at import time.
Built-in commands take precedence on name collision;
broken plugins warn to stderr without crashing.

Extract validate into overture-schema-validation, which
registers via entry point. Move filter_models(),
resolve_types(), and union-building helpers to
overture-schema-core so both packages share them
without cross-dependency.
```

## Test plan

- [x] `make check` passes (pytest + ruff + mypy across workspace)
- [x] `overture-schema --help` lists `validate` when validation package is installed
- [x] `overture-schema validate` works end-to-end
- [x] Built-in commands (`json-schema`, `list-types`) work independently of validation package
- [x] Plugin loading integration tests cover discovery, broken-plugin resilience, and name collision


packages/overture-schema-validation/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+    "pydantic>=2.0",
+    "pyyaml>=6.0.2",
+    "click>=8.0",
+    "rich>=13.0",
+    "yamlcore>=0.0.4",
+]
+description = "Validation command for Overture Maps schema CLI"
+dynamic = ["version"]
+license = "MIT"
+name = "overture-schema-validation"
```
RoelBollens-TomTom, line 13: Can we have a slight restructuring of the package names?

Current:

- overture-schema-addresses-theme
- overture-schema-addresses-theme
- overture-schema-base-theme
- overture-schema-cli
- overture-schema-validation
- ...

Proposed:

- overture-schema-theme-addresses
- overture-schema-theme-base
- overture-schema-cli
- overture-schema-cli-validation
- ...

This will keep the related packages visually grouped

sethfitz: I intentionally named this `-validation`; code generation has a minor validation requirement (we want to ensure that the TOML examples are valid) and I can imagine pulling in the Parquet file validation from OvertureMaps/schema#437. I also intend this to be a library for validation, not just the CLI.

For the theme packages, let's stick that on the WG agenda for broader discussion.

PR #450, opened 2026-02-25
Pydantic: update divisions to match data & fix constraints

Rachmanin0xFF:
A couple small updates to Pydantic that are in my [Parquet validation branch](https://github.com/OvertureMaps/schema/pull/437)

# Description

* Widen `AdminLevel` from `uint8` to `int32` to match actual division data
* Fix `ModelConstraint.decorate` validator dispatch: replace bound method with a closure to prevent Pydantic from rebinding self to the model instance


sethfitz (dismissed):

packages/overture-schema-divisions-theme/tests/division_baseline_schema.json
```diff
@@ -524,8 +524,8 @@
       "properties": {
         "admin_level": {
           "description": "Integer representing the division's position in its country's administrative hierarchy, where lower numbers correspond to higher level administrative units.",
-          "maximum": 255,
-          "minimum": 0,
```
jwass, line 528: I'm not sure these are right. I get that changing the min/max makes the integer size type right, but we still want the values constrained by what was there, no? admin_level can't really be negative let alone -2147483648

sethfitz: These are inherited from the data type. They can (and should?) be overridden locally, on `admin_level`.

jwass: Oh I see... they're not the original json files.

jenningsanderson (dismissed):
LGTM. 

Just to confirm — the `json` files in `tests` are autogenerated, correct? 

Should we be excluding these generally and generating them within actions or something else?

sethfitz:
> Just to confirm — the `json` files in `tests` are autogenerated, correct?

Yes, but only on demand (typically in response to changes made to the schema and compared to the previous version tracked by git to ensure that the differences are expected).

> Should we be excluding these generally and generating them within actions or something else?

IMO, in their current form they should be checked in. Whether this mechanism should exist at all is a different matter; rather than testing the JSON Schema, we may want to consider a different mechanism to check the behavior of individual Pydantic models (I don't know what this would look like).

packages/overture-schema-divisions-theme/tests/division_area_baseline_schema.json
```diff
@@ -457,8 +457,8 @@
       "properties": {
         "admin_level": {
           "description": "Integer representing the division's position in its country's administrative hierarchy, where lower numbers correspond to higher level administrative units.",
-          "maximum": 255,
-          "minimum": 0,
+          "maximum": 2147483647,
+          "minimum": -2147483648,
```
sasastanojkov: negative admin_level seems a bit off. Can we still keep it between 0 and 255?

sasastanojkov: Same for other types

sethfitz: @Rachmanin0xFF can you add a [field-level constraint](https://docs.pydantic.dev/latest/api/standard_library_types/#integers) to narrow the valid values (because of the confusion its absence has caused here, but also because it's a good idea?)

RoelBollens-TomTom: I assume a uint16 could also be picked as an alternative to the field-level constraint, as that type gets promoted to a int32 in the arrow schema generation.

Rachmanin0xFF: Done, I opted for the field-level constraint just to be readable / explicit.

FWIW, I started adding constraints to other models, but the list is pretty long. I think we should open a separate PR for hammering down field constraints. As a starting point, here are the actual ranges (in our data) of some other fields that are bottoming out to -2^31 in JSON:

Field | Min | Max
-- | -- | --
level | -9999 | 1993
cartography.sort_key | 1 | 17
num_floors | 1 | 163
num_floors_underground | 1 | 98
min_floor | 1 | 105

sethfitz (approved):

jenningsanderson (approved):

vcschapp:
@Rachmanin0xFF could you please investigate why the docs publish-to-staging workflow keeps failing on this PR and make any changes needed to make it pass?

Rachmanin0xFF:
> Adam Lastowka (@Rachmanin0xFF) could you please investigate why the docs publish-to-staging workflow keeps failing on this PR and make any changes needed to make it pass?

This was resolved here: https://github.com/OvertureMaps/schema/pull/455, I just had to update this branch. We're good now.

vcschapp (approved):

vcschapp:
Mentioned this PR during today's [2026-03-11 / CODING SESH 💻](https://github.com/OvertureMaps/schema-wg/issues/418).

> - Agreed, we should have pushed back and asked Adam to push for a Pipeline-level fix.
> - Agreed, we shouldn't merge anymore such PRs and should push back.
> - Agreed, Vic will raise this with Adam.

PR #451, opened 2026-02-27
[FEATURE](Markdown) code generation

sethfitz:
## Summary

Add `overture-schema-codegen`, a code generator that produces documentation from
Pydantic schema models.

Pydantic's `model_json_schema()` flattens the schema's domain vocabulary into JSON
Schema primitives. NewType names, constraint provenance, and custom constraint classes
disappear. Navigating Python's type annotation machinery -- NewType chains, nested
`Annotated` wrappers, union filtering, generic resolution -- is complex. The codegen
does it once. `analyze_type()` unwraps annotations into `TypeInfo`, a flat
target-independent representation that renderers consume without re-entering the type
system.

## Architecture

Four layers with strict downward imports. The package layout mirrors the
architecture -- each layer is a sub-package:

```text
markdown/    Rendering       ← Output formatting, all presentation decisions
layout/      Output Layout   ← What to generate, where it goes
extraction/  Extraction      ← TypeInfo, FieldSpec, ModelSpec, UnionSpec
             Discovery       ← discover_models() from overture-schema-core
```

Discovery lives in `overture-schema-core`, not in the codegen package.
`cli.py` sits at the package root and imports from all three sub-packages.

`analyze_type()` is the central function. A single iterative loop peels NewType,
Annotated, Union, and container wrappers in fixed order, accumulating constraints tagged
with the NewType that contributed them. The result is a `TypeInfo` dataclass that
downstream modules consume without re-entering the type system.

Both concrete `BaseModel` subclasses and discriminated union type aliases (like `Segment
= Annotated[Union[RoadSegment, ...], ...]`) satisfy the `FeatureSpec` protocol and flow
through the same pipeline. Union extraction finds the common base class, partitions
fields into shared and variant-specific, and extracts the discriminator mapping.

`markdown/pipeline.py` orchestrates the full pipeline without I/O: tree expansion,
supplementary type collection, path assignment, reverse references, and rendering.
Returns `list[RenderedPage]`. The CLI writes files to disk with Docusaurus frontmatter.

Design doc: `packages/overture-schema-codegen/docs/design.md`

## Changes outside the codegen package

Preparatory fixes and refactors in core/system/CLI packages:

- Rename `ModelKey.class_name` to `entry_point` (carries module:Class path, not just the
  class name)
- Attach docstrings to NewTypes at runtime (so the codegen can extract them)
- Add `resolve_discriminator_field_name()` to system feature module
- Fix relative imports and f-string prefixes in core
- Use `dict` instead of `Mapping` in system test util type hints

Example data added to theme `pyproject.toml` files (addresses, base, buildings,
divisions, places) under `[examples.ModelName]` sections.

## What's in the package

**Source** (33 files, ~3,800 lines):

| Module | Purpose |
| --- | --- |
| `extraction/type_analyzer.py` | Iterative type unwrapping into `TypeInfo` |
| `extraction/specs.py` | Data structures shared between extraction and rendering |
| `extraction/type_registry.py` | Type name → per-target display string mapping |
| `extraction/model_extraction.py` | Pydantic model → `ModelSpec`, tree expansion |
| `extraction/union_extraction.py` | Union alias → `UnionSpec`, discriminator mapping |
| `extraction/enum_extraction.py` | Enum → `EnumSpec` |
| `extraction/newtype_extraction.py` | NewType → `NewTypeSpec` |
| `extraction/numeric_extraction.py` | Numeric type extraction (`NumericSpec`) |
| `extraction/field_constraints.py` | Constraint objects → display text |
| `extraction/model_constraints.py` | Model-level constraints → prose |
| `layout/module_layout.py` | Python module paths → output directories |
| `layout/type_collection.py` | Supplementary type discovery from field trees |
| `markdown/path_assignment.py` | Type names → output file paths |
| `markdown/link_computation.py` | Relative links between output pages |
| `markdown/reverse_references.py` | "Used By" reference computation |
| `markdown/type_format.py` | `TypeInfo` → markdown type strings with links |
| `markdown/renderer.py` | Jinja2 template driver for all page types |
| `extraction/examples.py` | TOML example loading, validation, flattening |
| `markdown/pipeline.py` | Pipeline orchestration (no I/O), system type partitioning |
| `cli.py` | Click CLI: `generate` and `list` commands |
| `extraction/case_conversion.py` | PascalCase → snake_case |
| `extraction/docstring.py` | Custom vs. auto-generated docstring detection |

**Tests** (34 files, ~6,600 lines): unit tests per module, golden file tests for
rendered markdown, integration tests against real schema models.

## Design decisions worth reviewing

**`analyze_type` is iterative, not recursive.** The `while True` loop handles arbitrary
nesting depth (NewType wrapping Annotated wrapping NewType wrapping Annotated...)
without stack growth. Dict key/value types are the one exception where it recurses.

**Cache insertion before recursion in `expand_model_tree`.** The sub-model's `ModelSpec`
enters the cache before its fields are expanded. A back-edge encounter finds the cached
entry and marks `starts_cycle=True` rather than infinite-looping.

**`FeatureSpec` is a Protocol, not a base class.** `ModelSpec` and `UnionSpec` have
different field structures (flat list vs. annotated-field list with variant provenance).
A protocol lets them share a pipeline interface without forcing inheritance.

**Schema root computed from all entry points, before theme filtering.** Output directory
structure must remain stable regardless of which themes are selected. Computing the root
from filtered paths would shift directories when themes change.

**Constraint provenance via `ConstraintSource`.** Each constraint records which NewType
contributed it. Field-level constraints with `source=None` render on the field;
constraints with a named source render on the NewType's own page. This prevents
duplication.

## Test plan

- [ ] `make check` passes (pytest + doctests + ruff + mypy): 2,111 tests
- [ ] `make install && python -m overture.schema.codegen generate --format markdown
  --output-dir /tmp/schema-docs` produces output
- [ ] Spot-check generated markdown for a union feature (e.g., Segment) and a model
  feature (e.g., Building) -- field tables, links, constraint descriptions, examples
- [ ] Verify cross-page links resolve correctly (supplementary types link back to
  features, features link to shared types)


packages/overture-schema-codegen/src/overture/schema/codegen/extraction/type_analyzer.py
```diff
@@ -0,0 +1,327 @@
...
+            continue
+
+        # Handle list types (list[X])
+        if origin is list:
+            args = get_args(annotation)
+            if not args:
+                raise TypeError("Bare list without type argument is not supported")
+            state.is_list = True
+            annotation = args[0]
+            continue
```
Rachmanin0xFF, line 260: Haven't reviewed everything yet, but I found an issue in here while testing the parquet generator: this does not properly unpack nested lists. I don't think it's a problem for the markdown, but it surfaces in Divisions where we have `list[NewType("Hierarchy", list[HierarchyItem])]`. You can see the diff in the resulting arrow schemas:

Generated:                                                                                                          
`list<element: struct<division_id: string not null, subtype: string not null, name: string not null>>`                
                                                                                                                      
Release Data (2026-02-18.0):                                                                                                  
`list<element: list<element: struct<division_id: string, subtype: string, name: string>>>` 

This should instead use something like a `list_depth` or a recursive unwrap.

sethfitz: Good find! I’ve got a fix for this that I need to split across this PR and the Arrow one.

sethfitz: Fixes pushed (fed46cf2938daac558aa82c0dfbc85da8043c9c3 and 8a15703d501bca9d96b78cb68baa8584b08db5fe)

RoelBollens-TomTom:
Some observations without having looked into the code:

There are two different representations for a list in the markdown generation e.g:

| Name | Type | Description |
| -----: | :----: | ------------- |
| `emails` | `list<EmailStr>` (optional) | The email addresses of the place.<br/>*`minimum length: 1`*<br/>*Ensures all items in a collection are unique. (`UniqueItemsConstraint`)* |
| `phones` | [`PhoneNumber`](../system/phone_number.md) (list, optional) | The phone numbers of the place.<br/>*`minimum length: 1`*<br/>*Ensures all items in a collection are unique. (`UniqueItemsConstraint`)* |

In places the referenced Address type points to the Address type of the addresses theme `(../addresses/address.md)` which is incorrect:

| Name | Type | Description |
| -----: | :----: | ------------- |
| `addresses[]` | `list<`[`Address`](../addresses/address.md)`>` (optional) | The address or addresses of the place<br/>*`minimum length: 1`* |

And something very minor; when Pydantic types are used such as EmailStr or HttpUrl there wont be a reference:

| Name | Type | Description |
| -----: | :----: | ------------- |
| `emails` | `list<EmailStr>` (optional) | The email addresses of the place.<br/>*`minimum length: 1`*<br/>*Ensures all items in a collection are unique. (`UniqueItemsConstraint`)* |
| `websites` | `list<HttpUrl>` (optional) | The websites of the place.<br/>*`minimum length: 1`*<br/>*Ensures all items in a collection are unique. (`UniqueItemsConstraint`)* |

packages/overture-schema-codegen/src/overture/schema/codegen/example_loader.py
```diff
@@ -0,0 +1,315 @@
...
+            model_fields = {}
+
+    known_keys = _known_field_keys(model_fields)
+    preprocessed = _inject_literal_fields(model_fields, denulled)
+    preprocessed = _strip_null_unknown_fields(preprocessed, known_keys)
+    TypeAdapter(validation_type).validate_python(preprocessed)
+    return denulled
+
+
+_DEFAULT_SKIP_KEYS: frozenset[str] = frozenset({"bbox"})
```
Rachmanin0xFF: Why are we omitting bbox fields from the markdown examples?

packages/overture-schema-codegen/src/overture/schema/codegen/markdown_renderer.py
```diff
@@ -0,0 +1,606 @@
...
+    text = text.strip()
+    text = _unwrap_paragraphs(text)
+    text = text.replace("\n\n", "<br/><br/>")
+    return text.replace("|", "\\|")
+
+
+def _truncate(text: str) -> str:
+    """Truncate text to ``_EXAMPLE_TRUNCATION_LIMIT`` chars, adding ellipsis."""
+    if len(text) > _EXAMPLE_TRUNCATION_LIMIT:
+        return text[:_EXAMPLE_TRUNCATION_LIMIT] + "..."
```
Rachmanin0xFF: Should be `text[:_EXAMPLE_TRUNCATION_LIMIT - 3] + "..."`, I think?

Rachmanin0xFF: ...or `…` and `-1`...

packages/overture-schema-codegen/src/overture/schema/codegen/markdown/renderer.py
```diff
@@ -0,0 +1,606 @@
...
+            return ""
+        return f"`{_truncate(value)}`"
+
+    if isinstance(value, list):
+        items = ", ".join(str(item) for item in value)
+        return f"`{_truncate(f'[{items}]')}`"
+
+    if isinstance(value, dict):
+        pairs = ", ".join(f"{k}: {v}" for k, v in value.items())
+        return f"`{_truncate(f'{{{pairs}}}')}`"
```
Rachmanin0xFF, line 194: I think this is dead code? I don't know how a dict would end up here...

packages/overture-schema-codegen/src/overture/schema/codegen/markdown_renderer.py
```diff
@@ -0,0 +1,606 @@
...
+    if isinstance(value, bool):
+        return "`true`" if value else "`false`"
+
+    if isinstance(value, str):
+        if value == "":
+            return ""
+        return f"`{_truncate(value)}`"
+
+    if isinstance(value, list):
+        items = ", ".join(str(item) for item in value)
```
Rachmanin0xFF: Nit: maybe `repr()` instead of `str()` so strings get quotes?

packages/overture-schema-codegen/src/overture/schema/codegen/layout/type_collection.py
```diff
@@ -0,0 +1,119 @@
...
+        and dict recursion apply regardless of kind.
+        """
+
+        def _visit(node: TypeInfo) -> None:
+            if node.kind == TypeKind.UNION and node.union_members:
+                # Walk each member's fields for supplementary types.
+                # Members that are also top-level feature specs are skipped
+                # by the feature_names guard in _collect_from_model.
+                for member_cls in node.union_members:
+                    member_spec = extract_model(member_cls)
```
Rachmanin0xFF, line 104: Claude identified an issue here:

Union members in collect_all_supplementary_types are re-extracted via `extract_model(member_cls)` instead of using the already-expanded specs from expand_model_tree. The fresh `ModelSpec` has `field_spec.model=None` on all MODEL-kind fields, so if any union member contains a sub-model field, `_collect_from_fields` hits the RuntimeError guard at line 107-112. Currently safe because existing union members don't have nested sub-models, but it'll crash as soon as one does.

Rachmanin0xFF (dismissed):
I'm probably missing quite a bit (there is a lot of code here), but I like the structure + overall design and the generated markdown looks solid, so I'm approving.

Commented on a few small issues in addition to the aforementioned list representation confusion.

sethfitz:
@RoelBollens-TomTom good finds. I'm working on fixes for the incorrect `Address` reference (which results from name collisions) and creating pages + links for the Pydantic types.

The reason for the 2 different list representations is NewTypes that wrap `list` (and annotate with constraints) vs. vanilla lists. If we were to display `PhoneNumber` as `list<T>`, we'd lose the ability to link to `PhoneNumber` (and display its docstring, constraints, and references in more detail). We _could_ render `list<EmailStr>` as "`EmailStr` (list, optional)", but then we'd be treating `PhoneNumber` and `EmailStr` as the same thing even though one is a list, the other is scalar.

@RoelBollens-TomTom and @Rachmanin0xFF suggestions for making this less confusing?

Rachmanin0xFF:
> Roel Bollens (@RoelBollens-TomTom) and Adam Lastowka (@Rachmanin0xFF) suggestions for making this less confusing?

Maybe just a comment at https://github.com/OvertureMaps/schema/pull/451/changes#diff-d3543f3c56213c5ae4cf72e240b850d5cf763f9ceb7d2a9f9cf78c7602075739R110 would be fine.

packages/overture-schema-codegen/src/overture/schema/codegen/example_loader.py
sethfitz: Because the previous version (from last fall) omitted bbox fields. No good reason.

Fixed in 6a1c70c320a4188b1756a262f9c830207eb2b203

sethfitz:
I didn't understand the `list` rendering issue that both Roel and Adam flagged (because I was blind to it in the Markdown output), but after seeing it this morning, I fixed it in 48df0ab9ef23da91058a9a1ce34e5900e28caf87 (part of this PR).

packages/overture-schema-codegen/src/overture/schema/codegen/markdown_renderer.py
sethfitz: If we care that the result isn't 103 characters, sure ;-) (100 was arbitrary anyway.)

Fixed: 14e6ac7815da400b4cdbfab88e33e6d7f948535e

packages/overture-schema-codegen/src/overture/schema/codegen/markdown_renderer.py
sethfitz: Good thought: 3f4bb7238ecc80fcddd050d8911528cf1e88c952

packages/overture-schema-codegen/src/overture/schema/codegen/layout/type_collection.py
sethfitz: I knew about this and was going to fix it later (because we don't have anything that triggers it), but... 531621bf9d858899f98da3290b49df582e426b68

packages/overture-schema-codegen/src/overture/schema/codegen/markdown/renderer.py
sethfitz: Kind of... but this made me realize that the assumptions about example rendering were wrong. Right now it treats dicts (Python)/maps (Parquet) in the example data as if they lined up with dicts/structs in the schema, but that's not valid. Example data for Parquet map types should be rendered as stringified dicts, not broken out into separate rows.

I'm reworking the example rendering to reflect this change.

vcschapp:
One comment about the *Description* - I haven't got to the code yet.

The first five lines of the *Architecture* section with its four layers is a super nice compression of info. "Context compaction" to coin a phrase. It primed me to find an explicit hierarchical organization into those 4 layers. Then when I get to *What's in the package*, it's a big 'ol viewport-filling 22-row table of flag filenames, which is also in the diff.

Would it make sense to group the files into directories/modules based on the layer they belong to? It'd definitely help with the job of climbing up and down the abstraction ladder.

packages/overture-schema-codegen/src/overture/schema/codegen/markdown/renderer.py
sethfitz: Fixed in 1132e483cf688bcfde00c27d3b22ddc96d06dc69.

See `source_tags` in https://github.com/OvertureMaps/docs/pull/282/files#diff-438d0796f6f4052fff389ad396fa8277c26f4f3f747a98529f6943d10acf9e9bR85

sethfitz:
I'd sketched a reorganization that also included extracting the Markdown generator (and using, guess what, entry points to register codegen targets) into separate packages and was planning on discussing that later. I pulled the split by layer into 1132e483cf688bcfde00c27d3b22ddc96d06dc69.

packages/overture-schema-addresses-theme/pyproject.toml
```diff
@@ -38,3 +38,36 @@ testpaths = ["tests"]
 
 [project.entry-points."overture.models"]
 "overture:addresses:address" = "overture.schema.addresses:Address"
+
+[[examples.Address]]
```
vcschapp, line 42: ❤️

packages/overture-schema-addresses-theme/pyproject.toml
```diff
@@ -38,3 +38,36 @@ testpaths = ["tests"]
...
+id = "416ab01c-d836-4c4f-aedc-2f30941ce94d"
+geometry = "POINT (-176.5637854 -43.9471955)"
+country = "NZ"
+postcode = "null"
```
vcschapp: Couldn't we just omit these values?

Seems like that's "the TOML design choice"  anyway so it'd be consistent with Tom's Opinionated Opinion to leave them out, and AFAIK it shouldn't affect either the example validation or the example display in any way...

---

Other benefits of leaving them out: less code, less need to explain, and no risk of a "null conflation", where an example that's trying to put in the explicit string `"null"` has it replaced with `None`.

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/type_analyzer.py
```diff
@@ -0,0 +1,344 @@
+"""Iterative type unwrapping for Pydantic model annotations."""
```
vcschapp, line 1: Very minor suggestion: given the central importance of `TypeInfo` in the docs, I kept looking for the file that contained it, and couldn't find so had to guess a bit. Would calling this file `type_info.py` be an improvement overall, given it seems to be the central star around which the other exports orbit?

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
```diff
@@ -0,0 +1,260 @@
...
+    name: str
+    type_info: TypeInfo
+    description: str | None
+    is_required: bool
+    model: ModelSpec | None = None
+    starts_cycle: bool = False
+
+
+@runtime_checkable
+class FeatureSpec(Protocol):
```
vcschapp, line 120: What does `Feature` signify in this context?

Does it mean `Feature` in the same sense as used in system/core packages, or is it another sense?

---

Is there some other name that'd fit and not conflict with `Feature`?  Something like `Struct`?

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
```diff
@@ -0,0 +1,260 @@
...
+class NewTypeSpec(_SourceTypeIdentityMixin):
+    """Specification for a NewType."""
+
+    name: str
+    description: str | None
+    type_info: TypeInfo
+    source_type: object | None = None
+
+
+@dataclass
```
vcschapp, line 194: Is this just a *numeric* primitive?

If so the usage seems to conflict with other usages of primitive under the same directory, *e.g.*, `primitive_extraction.py` also includes geometry. IMO we should aim for consistent name usage.

---

Aside, in my mental model, a `string` is also a primitive although the system package doesn't really capture that inclusion.

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@cli.command()
+@click.option(
+    "--format",
+    "output_format",
+    required=True,
+    type=click.Choice(_OUTPUT_FORMATS),
+    help="Output format",
+)
```
vcschapp, line 77: Does `click` support short-forms? I'd say `--format markdown` is the correct canonical argument string, but `-f md` is going to be a lot more usable. Would be nice to have both.

packages/overture-schema-codegen/src/overture/schema/codegen/markdown/path_assignment.py
```diff
@@ -0,0 +1,115 @@
...
+__all__ = [
+    "GEOMETRY_PAGE",
+    "PRIMITIVES_PAGE",
+    "build_placement_registry",
+    "resolve_output_path",
+]
+
+# Aggregate page paths.
+PRIMITIVES_PAGE = PurePosixPath("system/primitive/primitives.md")
+GEOMETRY_PAGE = PurePosixPath("system/primitive/geometry.md")
```
vcschapp, line 28: Another place where "primitives" and "geometry" are treated as being separately.

---

Let's figure out ASAP what we need to do to bring everyone's conceptual model into alignment before it balkanizes into a bunch of semi-overlapping usages.

---

My internal definition has been something like:

> A primitive is a fundamental scalar type supported by the platform. Fundamental means that conceptually it is a whole, and not composed of other smaller pieces that can be used independently.

In my mind,the following meet the definition: numeric types, bool, string types, and geometry types; while the following do not: collections; and classes (including enums) apart from the classes that model numbers, strings, geometry.

We could think about alternative terms. For example **builtin** is often used for this kind of thing (but **primitive** does better in my mind at capturing the "it's not a struct item").

We can also look at moving things around in `system` if it helps produce an aligned model.

I think calling just the numbers **primitives** seems wrong. At that point it's more **numerics** isn't it?

packages/overture-schema-addresses-theme/pyproject.toml
sethfitz: Done in 952b59b2ca25f74277c318beac9a99ac591c05e5 as part of refactoring the example renderer to read from the Pydantic model version (vs. the dict loaded from the TOML).

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
sethfitz: I agree. I spent approximately no time on the CLI ergonomics (it doesn't even support `--type`!) with the intent to build a proper subcommand after #449 (and once we know what arguments we actually need).

packages/overture-schema-codegen/src/overture/schema/codegen/markdown/path_assignment.py
sethfitz: I didn't think too hard about this, expecting that we'd do a review of how the types manifest in the [schema reference](https://docs.overturemaps.org/schema/) from the perspective of the user journey and not only refactor the "list" pages ("primitives" + "geometry", which are admittedly arbitrary groupings) but also the placement and labeling of types that come from `core` and `system`. Pydantic types in use (`HttpUrl`, `EmailStr`) get their own pages as well, which is arguably weird.

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
sethfitz: It is [just a numeric primitive]. Good call, I'll adjust the nomenclature accordingly.

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
sethfitz: It's a thing with a setuptools entry point that's either a (`system`/`core`) `Feature` or a union of them (`ModelSpec` or `UnionSpec`). It's a top-level concept that gets its own page with special treatment (vs. `BaseModel`s used as fields, which are nested in the navigation hierarchy because they don't deserve to be _as_ discoverable).

`Struct` erases the distinction between top-level specs and field specs.

`EntryPointSpec` is another option, but lines up with our specific discovery implementation. I'm inclined to keep `FeatureSpec` in the sense that it references what we (broadly) consider a "Feature" (in Overture, but also `system`).

vcschapp (approved):

RoelBollens-TomTom (approved):

vcschapp:
image

Failing because of:

image

> ```
> Run uv run python ./.github/workflows/scripts/package-versions.py compare \
> Traceback (most recent call last):
>   File "/home/runner/work/schema/schema/./.github/workflows/scripts/package-versions.py", line 154, in <module>
>     main()
>   File "/home/runner/work/schema/schema/./.github/workflows/scripts/package-versions.py", line 147, in main
>     compare(sys.argv[2], sys.argv[3])
>   File "/home/runner/work/schema/schema/./.github/workflows/scripts/package-versions.py", line 71, in compare
>     combined_keys = sorted(list(set(before_dict.keys()) | set(after_dict.keys())), key=level)
>   File "/home/runner/work/schema/schema/./.github/workflows/scripts/package-versions.py", line 69, in level
>     raise ValueError(f"Unknown package for level computation: {package}")
> ValueError: Unknown package for level computation: overture-schema-codegen
> Error: Process completed with exit code 1.
> ```

Looks like we just need to add the codegen package into a manifest.

PR #457, opened 2026-03-03
Treat None as absent in model constraint validation

### LINKED ISSUE #456: `require_any_of` and related model constraints accept explicitly-set `None` as satisfying the requirement
## Behavior

`require_any_of('max_speed', 'min_speed')` accepts `max_speed=None` as satisfying the constraint. The same applies to `require_if` and `forbid_if` — all three treat "field is in `model_fields_set`" as sufficient, regardless of value.

```python
@require_any_of("max_speed", "min_speed")
class SpeedLimit(BaseModel):
    max_speed: int | None = None
    min_speed: int | None = None

SpeedLimit(max_speed=None)  # passes validation — should it?
```

## Why it works this way

The docstrings document this as intentional, designed for JSON Schema parity (appropriate given the nature of the port). In JSON Schema, `{"required": ["foo"]}` checks key presence — `{"foo": null}` satisfies it. The Python validators align to that semantic: `model_fields_set` membership equals satisfaction.

@vcschapp and I discussed this: the implementation focused on the presence/omitted distinction rather than Overture's domain semantics.

## Proposed fix

These constraints encode domain invariants. `require_any_of('max_speed', 'min_speed')` means "a speed limit needs at least one bound." `None` is not a bound.

Update `validate_instance` in all three constraints to require `model_fields_set` membership AND a non-None value. Update JSON Schema generation to match:

```json
{
  "anyOf": [
    {"required": ["max_speed"], "properties": {"max_speed": {"not": {"type": "null"}}}},
    {"required": ["min_speed"], "properties": {"min_speed": {"not": {"type": "null"}}}}
  ]
}
```

## Affected constraints

- `require_any_of` — at least one field must be set and non-null
- `require_if` — conditionally required fields must be set and non-null
- `forbid_if` — only non-null values violate the prohibition

sethfitz:
## Summary

- `require_any_of`, `require_if`, and `forbid_if` now treat `None` as "not present" for constraint purposes — fields must be both in `model_fields_set` and non-null to satisfy (or violate) a constraint
- JSON Schema generation emits `{"not": {"type": "null"}}` property constraints alongside `"required"` assertions, keeping Python and JSON Schema validation aligned
- Shared `required_non_null` helper in `_json_schema.py` and `_field_has_non_null_value` predicate on `OptionalFieldGroupConstraint` so the logic lives in one place

Fixes #456

## Test plan

- [x] `make check` passes
- [x] `require_any_of`: `TestModel(foo=None, bar=None)` raises `ValidationError`; `TestModel(foo=42, bar=None)` still passes
- [x] `require_if`: fields explicitly set to `None` are treated as missing when condition is true
- [x] `forbid_if`: fields explicitly set to `None` do not violate the prohibition
- [x] JSON Schema output includes `{"not": {"type": "null"}}` in `anyOf`, `then`, and `then.not` clauses
- [x] Baseline schemas for transportation and divisions themes regenerated

### DIFF
--- a/packages/overture-schema-system/src/overture/schema/system/_json_schema.py
+++ b/packages/overture-schema-system/src/overture/schema/system/_json_schema.py
@@ -333,20 +333,20 @@ def try_move(key: str, src: JsonSchemaValue, dst: JsonSchemaValue) -> None:
 
 def required_non_null(aliases: list[str]) -> JsonSchemaValue:
     """
-    Build a JSON Schema requiring listed fields to be present and non-null.
+    Build a JSON Schema requiring listed properties to be present and non-null.
 
-    Combines ``"required"`` (field key must exist) with a per-field property
-    constraint ``{"not": {"type": "null"}}`` (value must not be null).
+    Combines `"required"` (property must exist) with a per-property
+    constraint `{"not": {"type": "null"}}` (value must not be null).
 
     Parameters
     ----------
     aliases : list[str]
-        Non-empty list of field aliases (JSON property names) to constrain
+        Non-empty list of JSON Schema property names to constrain
 
     Returns
     -------
     JsonSchemaValue
-        Schema requiring each alias to be present and non-null
+        Schema requiring each property to be present and non-null
     """
     _verify_operands_not_empty(str, aliases)
     return {
--- a/packages/overture-schema-system/src/overture/schema/system/_json_schema.py
+++ b/packages/overture-schema-system/src/overture/schema/system/_json_schema.py
@@ -331,6 +331,30 @@ def try_move(key: str, src: JsonSchemaValue, dst: JsonSchemaValue) -> None:
         pass
 
 
+def required_non_null(aliases: list[str]) -> JsonSchemaValue:
+    """
+    Build a JSON Schema requiring listed fields to be present and non-null.
+
+    Combines ``"required"`` (field key must exist) with a per-field property
+    constraint ``{"not": {"type": "null"}}`` (value must not be null).
+
+    Parameters
+    ----------
+    aliases : list[str]
+        Non-empty list of field aliases (JSON property names) to constrain
+
+    Returns
+    -------
+    JsonSchemaValue
+        Schema requiring each alias to be present and non-null
+    """
+    _verify_operands_not_empty(str, aliases)
+    return {
+        "required": aliases,
+        "properties": {a: {"not": {"type": "null"}} for a in aliases},
+    }
+
+
 T = TypeVar("T", JsonSchemaValue, str)
 
 
--- a/packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
+++ b/packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
@@ -1,6 +1,6 @@
 """
-Prohibit every field in a group of fields from having a non-null value, but only if a condition
-is true.
+Prohibit every field in a group of fields from having a non-`None` value, but only if a
+condition is true.
 """
 
 from collections.abc import Callable
@@ -22,13 +22,11 @@ def forbid_if(
 ) -> Callable[[type[BaseModel]], type[BaseModel]]:
     """
     Decorate a Pydantic model class with a constraint forbidding any of the named fields from
-    holding a non-null value, but only if a field value condition is true.
+    holding a non-`None` value, but only if a field value condition is true.
 
     To ensure parity between Python and JSON Schema validation, a field's value must be explicitly
-    set to a non-null value to violate the constraint. This means in particular that fields whose
-    value was set by Pydantic using a default value do not count as violating the prohibition, and
-    fields containing the value `None`, even if explicitly set, do not count as violating the
-    prohibition.
+    set to a non-`None` value to violate the constraint. Fields containing the value `None`,
+    whether set explicitly or by default, are always compliant.
 
     Parameters
     ----------
@@ -57,13 +55,13 @@ def forbid_if(
     MyModel(foo='something', bar=42, baz='qux')
     >>> MyModel(foo='special value')                    # validates OK because bar/baz are omitted
     MyModel(foo='special value', bar=None, baz=None)
-    >>> MyModel(foo='special value', bar=None)          # validates OK because None doesn't violate
+    >>> MyModel(foo='special value', bar=None)          # validates OK because None is compliant
     MyModel(foo='special value', bar=None, baz=None)
     >>>
     >>> try:
     ...     MyModel(foo='special value', bar=42)
     ... except ValidationError as e:
-    ...     assert 'at least one field has a non-null value when it should not: bar' in str(e)
+    ...     assert 'at least one field is set to a value other than None when it must not be: bar' in str(e)
     ...     print('Validation failed')
     Validation failed
     """
@@ -128,13 +126,13 @@ class ForbidIfConstraint(OptionalFieldGroupConstraint):
         present_fields = [
             f
             for f in self.field_names
-            if self._field_has_non_null_value(model_instance, f)
+            if self._field_has_non_none_value(model_instance, f)
         ]
 
         if present_fields:
             raise ValueError(
-                f"at least one field has a non-null value when it should not: {', '.join(present_fields)} - "
-                f"these field value(s) are forbidden because {self.__condition} is true "
+                f"at least one field is set to a value other than None when it must not be: {', '.join(present_fields)} - "
+                f"these field(s) are forbidden because {self.__condition} is true "
                 f"(`{self.name}`)"
             )
 
--- a/packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
+++ b/packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
@@ -1,6 +1,6 @@
 """
-Prohibit every field in a group of fields from having a value explicitly set, but only if a
-condition is true.
+Prohibit every field in a group of fields from having a non-null value, but only if a condition
+is true.
 """
 
 from collections.abc import Callable
@@ -8,7 +8,7 @@ from collections.abc import Callable
 from pydantic import BaseModel, ConfigDict
 from typing_extensions import override
 
-from .._json_schema import get_static_json_schema_extra, put_if
+from .._json_schema import get_static_json_schema_extra, put_if, required_non_null
 from .model_constraint import (
     Condition,
     OptionalFieldGroupConstraint,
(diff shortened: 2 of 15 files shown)


packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
```diff
@@ -22,12 +22,13 @@ def forbid_if(
 ) -> Callable[[type[BaseModel]], type[BaseModel]]:
     """
     Decorate a Pydantic model class with a constraint forbidding any of the named fields from
-    holding an explicitly-assigned value, but only if a field value condition is true.
+    holding a non-null value, but only if a field value condition is true.
```
vcschapp: Can we use Python language and say non-`None` value and just generally refer to `None` rather than "null" throughout, both in the docs and in the code also?

packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
```diff
@@ -22,12 +22,13 @@ def forbid_if(
...
 
     To ensure parity between Python and JSON Schema validation, a field's value must be explicitly
-    set to violate the constraint. This means in particular that fields whose value was set by
-    Pydantic using a default value do not count as having a set value, and fields containing the
-    value `None`, if this value was explicitly set rather than being inherited by default, do count.
+    set to a non-null value to violate the constraint. This means in particular that fields whose
+    value was set by Pydantic using a default value do not count as violating the prohibition, and
+    fields containing the value `None`, even if explicitly set, do not count as violating the
+    prohibition.
```
vcschapp: This is breaking my brain right now. I think the *even if explicitly set* feels *non sequitur*-ish to me, or it's somehow subverting my expectations.

What if we finished it off from the "and," as:

> fields containing the value `None`, whether set explicitly or by default, are always compliant.

---

über-nit, is it better to stick with the term "constraint" throughout rather than adding the synonym "prohibition"?

packages/overture-schema-system/src/overture/schema/system/model_constraint/forbid_if.py
```diff
@@ -56,11 +57,13 @@ def forbid_if(
...
     >>> MyModel(foo='special value')                    # validates OK because bar/baz are omitted
     MyModel(foo='special value', bar=None, baz=None)
+    >>> MyModel(foo='special value', bar=None)          # validates OK because None doesn't violate
+    MyModel(foo='special value', bar=None, baz=None)
     >>>
     >>> try:
     ...     MyModel(foo='special value', bar=42)
     ... except ValidationError as e:
-    ...     assert 'at least one field has an explicit value when it should not: bar' in str(e)
+    ...     assert 'at least one field has a non-null value when it should not: bar' in str(e)
```
vcschapp: Can we do non-`None`, throughout?

packages/overture-schema-system/src/overture/schema/system/model_constraint/model_constraint.py
```diff
@@ -334,6 +334,13 @@ class OptionalFieldGroupConstraint(FieldGroupConstraint):
     def __init__(self, name: str | None, field_names: tuple[str, ...]):
         super().__init__(name, field_names)
 
+    @staticmethod
+    def _field_has_non_null_value(model_instance: BaseModel, field_name: str) -> bool:
```
vcschapp: Can we do `_field_has_non_none_value` and *mutatis mutandis* throughout, just to keep ourselves firmly in the Python dialect?

packages/overture-schema-system/src/overture/schema/system/model_constraint/require_any_of.py
```diff
@@ -1,28 +1,27 @@
...
 
     This function is the decorator version of the `RequireAnyOfConstraint` class.
 
     To ensure parity between Python and JSON Schema validation, a field's value must be explicitly
-    set to satisfy the constraint. This means in particular that fields whose value was set by
-    Pydantic using a default value do not count as having a set value, and fields containing the
-    value `None`, if this value was explicitly set rather than being inherited by default, do count.
+    set to a non-null value to satisfy the constraint. Fields whose value was set by Pydantic using
+    a default value do not count, and fields explicitly set to `None` do not count as satisfying
+    the constraint.
```
vcschapp: I don't know why I keep magnetically tweaking to these docs. Maybe the inherent double-negativeness of what we're working on is scrambling my already scrambled mind.

Anyway, this feels easier to grok to me:

> Fields whose value was set by Pydantic using a default value violate the constraint, as do fields explicitly set to the value `None`.

packages/overture-schema-system/src/overture/schema/system/_json_schema.py
```diff
@@ -331,6 +331,30 @@ def try_move(key: str, src: JsonSchemaValue, dst: JsonSchemaValue) -> None:
         pass
 
 
+def required_non_null(aliases: list[str]) -> JsonSchemaValue:
+    """
+    Build a JSON Schema requiring listed fields to be present and non-null.
+
+    Combines ``"required"`` (field key must exist) with a per-field property
+    constraint ``{"not": {"type": "null"}}`` (value must not be null).
```
vcschapp: I think the double backticks are a Sphinx/reST thing? The project is now mostly standardized on numpy-style and `pdoc` will render one backtick just fine. Might struggle with the doubles.

```
$ uv run pdoc -d numpy overture.schema.system
```

packages/overture-schema-system/src/overture/schema/system/_json_schema.py
```diff
@@ -331,6 +331,30 @@ def try_move(key: str, src: JsonSchemaValue, dst: JsonSchemaValue) -> None:
         pass
 
 
+def required_non_null(aliases: list[str]) -> JsonSchemaValue:
+    """
+    Build a JSON Schema requiring listed fields to be present and non-null.
```
vcschapp: Can we `s/fields/properties` and `s/aliases/properties`?

Philosophically this is just a module for manipulating JSON Schemas so doesn't really care about aliases and its terminology is "properties" rather than "fields".

packages/overture-schema-system/src/overture/schema/system/model_constraint/require_if.py
```diff
@@ -22,12 +21,12 @@ def require_if(
...
-    value explicitly set, but only if a condition is true.
+    non-null value, but only if a condition is true.
 
     To ensure parity between Python and JSON Schema validation, a field's value must be explicitly
-    set to satisfy the constraint. This means in particular that fields whose value was set by
-    Pydantic using a default value do not count as having a set value, and fields containing the
-    value `None`, if this value was explicitly set rather than being inherited by default, do count.
+    set to a non-null value to satisfy the constraint. This means in particular that fields whose
+    value was set by Pydantic using a default value do not count as having a set value, and fields
+    containing the value `None`, even if explicitly set, do not count as satisfying the constraint.
```
vcschapp: I think part of the *non sequitur*-ish feel is from the combination of the paragraphs:

> ... constraint requiring all of the named fields to have a non-`None` value ... and fields containing the value `None`, even if explicitly set, do not count as satisfying the constraint.

My mind says something feels confusing because the first part says you can't set `None` and then it goes on to say that *even if you set `None` explicitly, you still can't set `None`*... Like the second part is stating an obvious corollary in a way that makes it sound like a special case, so I end up going back and reading it over and over trying to figure out what I'm not getting. 🤯

Maybe if we ended the second paragraph something more like:

> This means in particular that fields whose value was set by Pydantic using a default still count as `None` and violate the constraint.

And just end it there?

packages/overture-schema-system/tests/model_constraint/test_forbid_if.py
```diff
@@ -105,6 +106,19 @@ class TestModel(BaseModel):
...
+def test_valid_model_instance_fields_explicitly_none_when_condition_true() -> None:
+    """Setting a forbidden field to None does not violate the prohibition."""
+
+    class TestModel(BaseModel):
+        foo: int | None = None
+        bar: int | None = None
+        baz: int
+
+    constraint = ForbidIfConstraint(["foo", "bar"], FieldEqCondition("baz", 42))
+    # foo and bar are explicitly set to None, but None doesn't count as a value
```
vcschapp: Superfluous comment?

vcschapp (changes requested):
I commented a bit, mostly about two topics:

- Tweaks to the docs that might make them clearer?
- Consistently using the Python term `None` over the Java/SQL term "null".

sethfitz:
Terminology tweaked and clarified.

vcschapp (approved):
```
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⠀⠤⠴⠶⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⣶⣾⣿⡟⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠂⠉⡇⠀⠀⠀⢰⣿⣿⣿⣿⣧⠀⠀⢀⣄⣀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⢠⣶⣶⣷⠀⠀⠀⠸⠟⠁⠀⡇⠀⠀⠀⠀⠀⢹⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠘⠟⢹⣋⣀⡀⢀⣤⣶⣿⣿⣿⣿⣿⡿⠛⣠⣼⣿⡟⠀⠀⠀⠀
⠀⠀⠀⠀⠀⣴⣾⣿⣿⣿⣿⢁⣾⣿⣿⣿⣿⣿⣿⡿⢁⣾⣿⣿⣿⠁⠀⠀⠀⠀
⠀⠀⠀⠀⠸⣿⣿⣿⣿⣿⣿⢸⣿⣿⣿⣿⣿⣿⣿⡇⢸⣿⣿⣿⠿⠇⠀⠀⠀⠀
⠀⠀⠀⠳⣤⣙⠟⠛⢻⠿⣿⠸⣿⣿⣿⣿⣿⣿⣿⣇⠘⠉⠀⢸⠀⢀⣠⠀⠀⠀
⠀⠀⠀⠀⠈⠻⣷⣦⣼⠀⠀⠀⢻⣿⣿⠿⢿⡿⠿⣿⡄⠀⠀⣼⣷⣿⣿⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠈⣿⣿⣿⣶⣄⡈⠉⠀⠀⢸⡇⠀⠀⠉⠂⠀⣿⣿⣿⣧⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠘⣿⣿⣿⣿⣿⣷⣤⣀⣸⣧⣠⣤⣴⣶⣾⣿⣿⣿⡿⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠇⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠘⢿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠿⠟⠛⠉⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠉⠉⠉⠉⠉⠉⠉⠉⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```

PR #458, opened 2026-03-04
[REFACTOR](system) Reparent string constraints under PatternConstraint + test improvements

sethfitz:
## Summary

- Reparent 8 `StringConstraint` subclasses (`CountryCodeAlpha2`, `HexColor`, `LanguageTag`, `NoWhitespace`, `SnakeCase`, `PhoneNumber`, `RegionCode`, `WikidataId`) under `PatternConstraint` as thin `__init__`-only wrappers
- Enrich `PatternConstraint` with optional `description`, `min_length`, `max_length` keyword parameters for JSON Schema annotations
- Extract `_raise_validation_error()` helper on `StringConstraint` to deduplicate error construction across `PatternConstraint`, `JsonPointerConstraint`, and `StrippedConstraint`
- Fix `StrippedConstraint` JSON schema pattern to accept empty string (matching Python validator behavior)
- Strengthen weak test assertions to check error messages, not just error presence
- Parametrize PatternConstraint subclass tests: 16 individual methods → two data-driven tests covering all 8 subclasses

-285 net lines across constraints and tests. Enables `isinstance`-based dispatch for pattern constraints in downstream codegen.

`JsonPointerConstraint` and `StrippedConstraint` remain as direct `StringConstraint` subclasses (non-regex validation logic).

## Test plan

- [x] Existing validation and JSON schema tests pass (behavioral safety net)
- [x] `isinstance` parametrized test confirms all 8 subclasses are `PatternConstraint` instances
- [x] Tests verify `PatternConstraint` kwargs emit/omit JSON Schema annotations correctly
- [x] `StrippedConstraint` empty-string acceptance tested against both validator and JSON schema pattern
- [x] `make check` passes

### DIFF
--- a/packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py
+++ b/packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py
@@ -3,7 +3,7 @@ Constraints on fields with string values.
 """
 
 import re
-from typing import Any
+from typing import Any, NoReturn
 
 from pydantic import (
     GetCoreSchemaHandler,
@@ -23,6 +23,23 @@ from .field_constraint import FieldConstraint
 class StringConstraint(FieldConstraint):
     """Base class for string-based constraints."""
 
+    def _raise_validation_error(
+        self, value: str, info: ValidationInfo, message: str
+    ) -> NoReturn:
+        context = info.context or {}
+        loc = context.get("loc_prefix", ()) + ("value",)
+        raise ValidationError.from_exception_data(
+            title=self.__class__.__name__,
+            line_errors=[
+                InitErrorDetails(
+                    type="value_error",
+                    loc=loc,
+                    input=value,
+                    ctx={"error": message},
+                )
+            ],
+        )
+
     def __get_pydantic_core_schema__(
         self, source: type[Any], handler: GetCoreSchemaHandler
     ) -> core_schema.CoreSchema:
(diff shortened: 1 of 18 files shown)


packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py
```diff
@@ -38,33 +55,51 @@ def validate_string(value: str, info: ValidationInfo) -> str:
...
+    """Generic pattern-based string constraint.
+
+    Args:
+        pattern: Regular expression to match against.
+        error_message: Error message template. Use ``{value}`` to interpolate
+            the failing value (the only available placeholder).
+        flags: Regex flags passed to ``re.compile``.
+        description: JSON Schema ``description`` annotation.
+        min_length: JSON Schema ``minLength`` annotation.
+        max_length: JSON Schema ``maxLength`` annotation.
```
vcschapp: Don't hate me, but can you write this in numpy style?

---

Ideally `$ make docformat` does not add any new issues, and `$ uv run pdoc -d numpy overture.schema.system` should render the docs in the way you expect.

Part of my unfinished work is to get the `make docformat` errors down to zero and then enable it as part of `make check` - but in the meantime if we could avoid increasing the uphill climb that'd be ideal.

vcschapp (changes requested):
This is fantastic, especially the:

> Net -112 lines.

And also the:

> Enables isinstance-based dispatch for pattern constraints in downstream codegen.

Would you mind humoring me with the numpy-style docs? I was so close to getting it all on numpy.

packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py
sethfitz: Sorry, I've been trying to hew to this (and the codegen PR actually does!). Fixed.

sethfitz:
Now -285 net lines after I found some more things to simplify.

I also fixed a JSON Schema validation bug with `StrippedString`--its behavior differed from Pydantic's `validate()`.

packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py
```diff
@@ -38,33 +55,58 @@ def validate_string(value: str, info: ValidationInfo) -> str:
...
+        Error message template. Use ``{value}`` to interpolate the failing
+        value (the only available placeholder).
+    flags : int
+        Regex flags passed to ``re.compile``.
+    description : str or None
+        JSON Schema ``description`` annotation.
+    min_length : int or None
+        JSON Schema ``minLength`` annotation.
+    max_length : int or None
+        JSON Schema ``maxLength`` annotation.
```
vcschapp: For future reference: single backticks should be fine. AFAIK, double backticks are Sphinx-style reST markdown, which we're not using.

If it renders correctly in `pdoc` and doesn't add any more errors to `make docformat` it should be fine - but can you confirm?

sethfitz: It does render correctly, but I'll update these for consistency.

image

PR #460, opened 2026-03-07
feat(codegen): add wassirman validation IR generation target

sethfitz:
Proof-of-concept: generating the same YAML validation IR that
`schema-validator` produces, but from the codegen extraction layer instead
of re-walking Pydantic internals.

## Why this matters

The codegen extraction layer (`TypeInfo`, `FieldSpec`, `ModelSpec`, tree
expansion) already solves the hard problems — unwrapping `Annotated`,
`NewType`, `Union`, nested lists, collecting constraints with provenance.
The markdown renderer is one consumer, [Arrow/Parquet](#452) is another.
This PR adds a third target that produces validation rules, demonstrating
that the extraction machinery generalizes across output formats.

The walker (`wassirman/walker.py`) is ~160 lines of rule emission logic
that operates on expanded `FeatureSpec` trees. Compare that to
`schema-validator`'s `extract.py`, which re-derives the same information
from raw Pydantic model internals. Same output, different extraction path.

Golden tests confirm the output matches the reference validator for all 16
feature types.

## Usage

```bash
# All feature types to stdout
overture-codegen generate --format wassirman

# Filter by theme
overture-codegen generate --format wassirman --theme places

# Per-dataset files
overture-codegen generate --format wassirman --output-dir ./validation-rules
```

## Discussion points

- The extraction layer captures more than what the current IR needs (descriptions,
  union discriminators, cross-field constraint provenance). Future targets get
  that for free.
- The IR is target-agnostic — it describes *what* to validate, not *how*.
  The same rules feed PySpark and DuckDB backends.
- What should the IR look like long-term? The current shape matches
  `schema-validator`'s output, but the richer extraction data could support
  a more expressive format.


vcschapp:
Closing stale PR.

PR #463, opened 2026-03-10
[FEATURE] Extensible tag classification model discovery through Entry Points

### LINKED ISSUE #512: Extensible tag-based model classification and discovery
Replace the hardcoded `namespace`/`ModelKind` classification with tags --
string labels derived by tag providers registered through Entry Points.

`system` provides generic tag-based grouping without understanding what
any tag means. Any package can register tag providers that classify
models without special support in the discovery layer. Model discovery
moves from `core` into `system` in the process.

Tags serve three roles:

- **CLI filtering** -- select subsets of models for output and codegen
  (`--tag system:feature`, `--tag draft`)
- **Classification and endorsement** -- distinguish features from
  extensions, mark models as vetted or approved
- **Marketplace taxonomy** -- browse and classify models and extensions
  in a future extension catalog

Builds on the design from #440.

Tracked by PR #463.

RoelBollens-TomTom:
# Extensible tag classification model discovery through Entry Points

This replaces the hardcoded model classification system with tag-based classification model discovery through Entry Points. This is based on #440 by Seth and several schema (ad-hoc) coding sessions where Seth, Vic, Dana, Tristan and Roel participated in.

Model discovery moved into `system`, eliminating assumptions about Overture in the process. The hardcoded `namespace` concept (`"overture"`, `"annex"`) and the `ModelKind` classifier is replaced with tags -- string labels derived by tag providers. Tags become the filtering, grouping, and classification mechanism for model discovery, driven by introspection and package metadata rather than central coordination.

`system` provides generic tag-based grouping without understanding what any particular tag means. Any package can register tag providers that classify models without special support in the discovery layer.

## Purpose

Tags serve three roles:

- **CLI filtering**: select subsets of models for output and codegen (`--tag system:feature`, `--tag draft`)
- **Classification and endorsement**: distinguish features from extensions, mark models as vetted or approved by an authority
- **Marketplace taxonomy**: browse and classify models and extensions in a future extension catalog

These roles overlap -- a tag like `overture:theme=buildings` serves both filtering and taxonomy. The design accommodates this overlap through structured tags that encode both ownership and dimension.

## Tag Format

Tags are strings following the pattern `[prefix:]key[=value]`:

- **Plain**: `overture`, `draft`, `feature`
- **Prefixed**: `system:extension` -- `:` separates ownership
- **Prefixed k/v**: `overture:theme=buildings`

`:` signals ownership and enables prefix reservation (see Privileged Packages and Tag Reservation). `=` signals a dimension with a value (groupable via `--group-by`). One level of each -- no nested colons or multiple `=` signs.

## Minimal launch set

| Tag | Meaning |
|-----|---------|
| `feature`  (was: `system:feature`)| This model is a feature type (has geometry, inherits from Feature | OvertureFeature) |
| `overture:theme=<theme>` | Which Overture theme this belongs to (e.g., `buildings`, `transportation`) |
| `overture` (was: `overture:official`) | Placeholder for a lifecycle/endorsement tag — exact name deferred pending Dana and Tristan's work on extension lifecycle |

## Reserved tags

Tags can be reserved either as simple tags or by namespaces. These are the tags and namespaces that are currently reserved:

| Tag | Reserved for use by |
|-----|---------------------|
| `feature`  |  Tag providers from `overture-schema-system` |
| `overture` | Tag providers from  `overture-schema-core` |
| `overture:*` |  Tag providers from  `overture-schema-core` |
| `system:*` |  Tag providers from  `overture-schema-system` |

## Extensions 

Additional extensions and accompanied tags will be introduced in a future PR. Extensions allows to augment existing types with new fields (columns). 

| Tag | Meaning | 
|-----|---------| 
| system:extension | This model is an extension (adds columns/fields to an existing type) |

## CLI

The `list-types` command has been updated to support filtering and grouping by tags. Currently, it no longer displays the description or fully qualified class name. The `json-schema` and `validate` commands from the overture-schema cli and  `generate` command from the overture-codegen cli have been updated to be able to filter on tags instead of filtering by theme and type. Further changes can be introduced in a future update.

### Examples

```console
% overture-schema list-types
address            feature  overture  overture:theme=addresses
bathymetry         feature  overture  overture:theme=base
building           feature  overture  overture:theme=buildings
building_part      feature  overture  overture:theme=buildings
connector          feature  overture  overture:theme=transportation
division           feature  overture  overture:theme=divisions
division_area      feature  overture  overture:theme=divisions
division_boundary  feature  overture  overture:theme=divisions
infrastructure     feature  overture  overture:theme=base
land               feature  overture  overture:theme=base
land_cover         feature  overture  overture:theme=base
land_use           feature  overture  overture:theme=base
place              feature  overture  overture:theme=places
segment            feature  overture  overture:theme=transportation
sources            overture
water              feature  overture  overture:theme=base
```

```console
% overture-schema list-types --group-by overture:theme
overture:theme=addresses (1)
→ address            feature  overture  overture:theme=addresses

overture:theme=base (6)
→ bathymetry         feature  overture  overture:theme=base
→ infrastructure     feature  overture  overture:theme=base
→ land               feature  overture  overture:theme=base
→ land_cover         feature  overture  overture:theme=base
→ land_use           feature  overture  overture:theme=base
→ water              feature  overture  overture:theme=base

overture:theme=buildings (2)
→ building           feature  overture  overture:theme=buildings
→ building_part      feature  overture  overture:theme=buildings

overture:theme=divisions (3)
→ division           feature  overture  overture:theme=divisions
→ division_area      feature  overture  overture:theme=divisions
→ division_boundary  feature  overture  overture:theme=divisions

overture:theme=places (1)
→ place              feature  overture  overture:theme=places

overture:theme=transportation (2)
→ connector          feature  overture  overture:theme=transportation
→ segment            feature  overture  overture:theme=transportation
```

```console
% overture-schema list-types --tag overture --exclude-tag overture:theme=base 
address            feature  overture  overture:theme=addresses
building           feature  overture  overture:theme=buildings
building_part      feature  overture  overture:theme=buildings
connector          feature  overture  overture:theme=transportation
division           feature  overture  overture:theme=divisions
division_area      feature  overture  overture:theme=divisions
division_boundary  feature  overture  overture:theme=divisions
place              feature  overture  overture:theme=places
segment            feature  overture  overture:theme=transportation
sources            overture
```

## Deviations

- Tag providers are additive only and can't remove existing tags.
- The execution order of tag providers is non-deterministic.
- There is currently no warning on a tag amount limit 
- Agreed minimal tag set deviates from coding session outcome to make them less clunky by dropping namespace

Closes #512


packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -207,34 +206,40 @@ def resolve_types(
...
     # Filter models based on CLI options
     filtered_models: ModelDict = {}
 
+    if namespace and namespace != "overture":
+        filtered_models = {
+            key: model_class
+            for key, model_class in all_models.items()
+            if namespace in key.tags
+        }
+
```
sethfitz: I doubt that anyone's relying on `--namespace` right now, so I'd drop this in favor of the `--tag <namespace>` support that you've added.

README.pydantic.md
```diff
@@ -149,26 +149,26 @@ Registration is done in the `[project.entry-points."overture.models"]` section:
...
+from overture.schema.system.discovery import discover_models, get_registered_model
 
 # Discover all registered models
 all_models = discover_models()
 # Returns:
 # {
-#   ("buildings", "building"): BuildingModel,
-#   ("places", "place"): PlaceModel,
+#   ("building", "acme:Building", {"building_tag"}): BuildingModel,
+#   ("place", "acme:Place", {"place_tag"}): PlaceModel,
```
sethfitz: `(<entry point name>, ???, set(<tags>)): <model>`, right?

What is `???` and where does it come from?

packages/overture-schema-cli/tests/test_resolve_types.py
```diff
@@ -124,10 +125,10 @@ def test_resolve_types_returns_expected_themes(
         expected_themes: set[str],
     ) -> None:
         """Test that resolve_types returns models from expected themes."""
-        from overture.schema.core.discovery import discover_models
+        from overture.schema.system.discovery import discover_models
 
-        models = discover_models(namespace=namespace)
-        actual_themes = {key.theme for key in models.keys()}
+        models = discover_models()
+        actual_themes = {next(iter(tags_by_key(key.tags, "overture:theme")),None) for key in models.keys()}
```
sethfitz: Extract `next(iter(tags_by_key(key.tags, "overture:theme")), None)` into a helper?

Ah, but where (since `theme` isn't a system concept)? Maybe `core`, but the intent is for `codegen` to drop its dependency on `core` in favor of `system` after this PR.

packages/overture-schema-cli/tests/test_resolve_types.py
```diff
@@ -124,10 +125,10 @@ def test_resolve_types_returns_expected_themes(
         expected_themes: set[str],
     ) -> None:
         """Test that resolve_types returns models from expected themes."""
-        from overture.schema.core.discovery import discover_models
+        from overture.schema.system.discovery import discover_models
```
sethfitz: My bad from before; we should aim to eliminate non-top-level imports unless they're absolutely necessary. I'd like to add a linting rule to catch this at some point (my local linter config adds this now, so `codegen` should be clean).

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
```diff
@@ -99,7 +104,7 @@ def generate(
     schema_root = compute_schema_root(module_paths)
 
     models = (
-        {k: v for k, v in all_models.items() if k.theme in theme}
+        {k: v for k, v in all_models.items() if _find_theme(k.tags) in theme}
```
sethfitz: We should replace this with `--tag` support. (As I said this before, I didn't spend much time on the CLI because I'd like to refactor it into a contributed sub-command that can re-use common CLI functionality more easily.)

packages/overture-schema-core/src/overture/schema/core/tag_providers.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if origin is Literal:
+            for val in get_args(t):
+                result.add(type(val))
+            return
+
+        result.add(t)
+
+    visit(tp)
+    return result
```
sethfitz: There are 2 other implementations that do similar things across the project. The CLI package has one (which is likely incomplete, but good enough for its purposes); `codegen` has the most complete version. I have a local TODO (for after you land this) to reconcile type analysis and move the relevant parts into `system` so that CLI can drop its implementation and `codegen` can re-use it.

It might make sense to pull the `codegen` implementation into scope so you can get one that's battle tested for all of our edge cases.

See `analyze_type` (`codegen/extraction/type_analyzer.py:167`); it'd be something like this:

```python
def extract_types(tp: object) -> set[type]:
    info = analyze_type(tp)
    if info.kind == TypeKind.MODEL and info.source_type is not None:
        return {info.source_type}
    if info.kind == TypeKind.UNION and info.union_members:
        return set(info.union_members)
    return set()
```

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+    "feature": {"overture-schema-system"},
+}
+TAG = r"[a-z0-9][a-z0-9_-]*"
+NAMESPACE_TAG = r"[a-z0-9]+:[a-z0-9]+(?:=[a-z0-9_.-]+)?"
+TAG_RE = re.compile(rf"^(?:{TAG}|{NAMESPACE_TAG})$")
+
+
+@dataclass(frozen=True, slots=True)
+class ModelKey:
+    """Key identifying a registered model by namespace, theme, and type.
```
sethfitz: Stale; we dropped `namespace` and `theme`.

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+from overture.schema.system.feature import Feature
+
+logger = logging.getLogger(__name__)
+
+
+RESERVED_TAGS: dict[str, set[str]] = {
+    "overture": {"overture-schema-core"},
+    "feature": {"overture-schema-system"},
+}
```
sethfitz: `feature_provider` emits `system:feature` and `overture_provider` emits `overture:feature` — both prefixed. These reserve the plain forms (`feature`, `overture`), which no provider currently produces.

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+def feature_provider(
+    model_class: type[BaseModel], key: ModelKey, tags: set[str]
+) -> set[str]:
+    if any(issubclass(tp, Feature) for tp in _extract_types(model_class)):
+        tags.add("system:feature")
+    return tags
+
+
+def _extract_types(tp: Any) -> set[type]:
```
sethfitz: As above.

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+        f"{namespace}:"
+        for namespace, reserved_packages in RESERVED_TAGS.items()
+        if package not in reserved_packages
+    )
+
+    return {
+        tag
+        for tag in tags
+        if TAG_RE.match(tag) and not tag.startswith(reserved_namespaces)
+    }
```
sethfitz: `_filter_tags` enforces prefix reservation but not plain tag reservation. A third-party package producing a plain `overture` tag would pass through unchecked — is that intentional, or should the plain entries in `RESERVED_TAGS` be enforced too?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+            except Exception as e:
+                # Log warning but don't fail for individual tag providers
+                logger.warning(
+                    "Could not load tag provider %s: %s", tag_provider.name, e
+                )
+    except Exception as e:
+        logger.warning("Could not discover entry points: %s", e)
+
+    return tag_providers
```
sethfitz: Providers see accumulated tags from prior providers, so execution order matters (e.g., the feature provider needs to run before anything that checks for `system:feature`). `discover_tag_providers` doesn't sort — what determines the ordering?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+                        if model.dist
+                        else frozenset()
+                    ),
+                )
+
+                try:
+                    key = replace(
+                        key,
+                        tags=frozenset(generate_tags(model_class, key, tag_providers)),
+                    )
```
sethfitz: `generate_tags` starts from an empty set, so the `_distribution_tags` seeded on lines 157-163 get replaced here. Should `generate_tags` start from the existing tag set instead?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+    reserved_namespaces = tuple(
+        f"{namespace}:"
+        for namespace, reserved_packages in RESERVED_TAGS.items()
+        if package not in reserved_packages
+    )
+
+    return {
+        tag
+        for tag in tags
+        if TAG_RE.match(tag) and not tag.startswith(reserved_namespaces)
```
sethfitz: `_filter_tags` builds prefix checks from `RESERVED_TAGS`, so `"overture"` here implicitly reserves both the plain tag and the `overture:` prefix. What happens when you need to reserve a plain tag that shouldn't also claim a prefix namespace?

packages/overture-schema-core/src/overture/schema/core/tag_providers.py
```diff
@@ -0,0 +1,88 @@
...
+            tags.add(
+                "overture:theme=" + get_args(tp.model_fields["theme"].annotation)[0]
+            )
+    return tags
+
+
+def _matches_manifest(key: ModelKey) -> bool:
+    if key.entry_point in APPROVED:
+        return True
+    return False
```
sethfitz:
```suggestion
    return key.entry_point in APPROVED:
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -822,35 +806,51 @@ def list_types() -> None:
     """
     try:
         models = discover_models()
+        filters = []
+
+        if tags:
+            filters.append(lambda key: all(tag in key.tags for tag in tags))
```
sethfitz: `all()` here means `--tag foo --tag bar` requires BOTH tags (AND). Is that the intended semantics, or should repeated `--tag` flags use OR?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+logger = logging.getLogger(__name__)
+
+
+RESERVED_TAGS: dict[str, set[str]] = {
+    "overture": {"overture-schema-core"},
+    "feature": {"overture-schema-system"},
+}
+TAG = r"[a-z0-9][a-z0-9_-]*"
+NAMESPACE_TAG = r"[a-z0-9]+:[a-z0-9]+(?:=[a-z0-9_.-]+)?"
+TAG_RE = re.compile(rf"^(?:{TAG}|{NAMESPACE_TAG})$")
```
sethfitz: `TAG` allows hyphens and underscores (`[a-z0-9_-]`), but `NAMESPACE_TAG`'s key segment doesn't — `acme:my-ext` would fail validation even though `my-ext` is a valid plain tag. Intentional?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+    for provider_key, provider in providers.items():
+        try:
+            added_tags = provider(model_class, key, tags.copy()).difference(tags)
+            filtered_tags = _filter_tags(added_tags, provider_key.package_name)
+            tags.update(filtered_tags)
+        except Exception as e:
+            print(
+                f"Error in tag provider {provider.__name__} for model {key.name}: {e}"
+            )
```
sethfitz:
```suggestion
            logger.warning(
                f"Error in tag provider {provider.__name__} for model {key.name}: {e}"
            )
```

sethfitz (changes requested):

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: This was intentional, but I guess this should still be formalized so I'll leave this open for discussion. 

from the coding session minutes:
> **--tag semantics are AND**
The mockups assume `--tag overture --tag system:feature` means "must have both tags" (_AND_). This contradicts the design doc's stated _OR_ semantics. The group implicitly operated with _AND_ throughout the exercise and no one objected. This should be formalized.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: The tag definition/validation needs some more love. I added a regex to have some validation but very little thought went into it. Same where TAG doesn't allow for decimal points/periods were a value in a namespace tag would (`3.14` would fail but `acme:pi=3.14` will pass).

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: I had left the `json-schema` and `validate` commands to kept their existing interface and modified them to maintain backward compatibility, I still plan update then (and codegen cli's commands) but left it a mystery as if it would be done as part of this PR or an other one.

README.pydantic.md
RoelBollens-TomTom: `(<entry point key>, <entry point value>, set(<tags>)): <model>`

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
RoelBollens-TomTom: Same as above. I'll either update the CLI's in this PR or I'll create a new one.

packages/overture-schema-core/src/overture/schema/core/tag_providers.py
RoelBollens-TomTom: I mentioned this during yesterdays code session as well. I intentionally left this un-extracted into the code where it's needed (and duplicated it for `system`'s tag providers where I needed it). I prefer not to reuse anything from `codegen` yet to avoid circular dependancies. 

Reconciling relevant parts from `codegen` into `system` is a good idea but we might want to hold off until extensions land as well, where I have more specialized needs and I want to make sure we create the right abstractions for it.

```python
def create_extended_model(model, extensions):

    origin = get_origin(model)

    if origin is Annotated:
        tp, *metadata = get_args(model)
        return Annotated.__class_getitem__(
            (create_extended_model(tp, extensions), *metadata)
        )

    if origin is Union:
        return Union[
            tuple(create_extended_model(tp, extensions) for tp in get_args(model))
        ]

    if hasattr(model, "__supertype__"):
        return NewType(
            model.__name__, create_extended_model(model.__supertype__, extensions)
        )

    extensions = {
        type: (cls | None, None)
        for type, cls in extensions.items()
        if any(issubclass(model, _unwrap_types(base)) for base in extends_classes(cls))
    }

    if extensions:

        return create_model(
            model.__name__,
            __base__=model,
            __doc__=model.__doc__,
            __module__=model.__module__,
            **extensions,
        )

    return model
```

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: Ahh my bad, I'll have to refactor this a bit, this is caused by a misunderstanding where the implementation is reserving namespaces rather than tags.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: As above.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: As above, I indeed reserved namespaces and not tags, this will be refactored.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: As above.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: I populated the initial key with a narrow set of distribution metadata so that tag providers can use of this in a (arguably) slightly cleaner way. I didn't want to `generate_tags` start from this information as it isn't necessarily what is desired as tags and any tag provider would be able to add information from it to the set but wouldn't really know which tags from distribution metadata is undesirable (as well as tag providers being able to add tags, but not remove).  I added this because the design document mentioned that reserved tags could also be reserved from specific package versions rather than just the distribution name but decided to leave this of implementation to keep it simple. While unused at this moment, I left the `_distribution_tags` in as a discussion point.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: The ordering is not deterministic. I do have a proposal for adding determinism, different from the initial design document but won't add it as part of this PR. 

Basically rather than setting a priority which I believe will be difficult to maintain as it's scattered and adds semantics to the entry_point key. I'd like to build a dependency graph instead.

Something like:
```python
def depends_on(tag_provider: EntryPointKey | EntryPointValue):
    def decorator(func):
        func._depends_on = tag_provider   
        return func
    return decorator


@depends_on("system") # optional
def some_tag_provider(model_class: type[BaseModel], key: ModelKey, tags: set[str]) -> set[str]:
    pass
```

README.pydantic.md
sethfitz: :+1: Makes sense, especially since the value may be an alias that links to the type that's loaded.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
sethfitz: Makes sense. My point was to not worry about backward-compatibility.

packages/overture-schema-system/src/overture/schema/system/discovery.py
sethfitz: I like the idea of passing metadata as part of `ModelKey` so that tag providers can use it. I'm not sure it makes sense to turn into tags (although I think I also considered that in my design).

I misread the implementation; I was looking for `tags` in here (as the thing passed in, probably a result of (proper) Python formatting in the `key` assignment above and missed that the `dist:` tags come in on `key` instead.

I guess my inadvertent point is that the current mechanism is hard to track because of the indirection, at least for me.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
sethfitz: Ah, great. I'm good with that; it didn't track with how I'd been thinking about it earlier is all.

packages/overture-schema-system/src/overture/schema/system/discovery.py
sethfitz: That handling of numbers makes sense to me. Raw tags with decimals feels somewhat confusing, so while it's restrictive, I could justify it.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: To be fair, I didn't care about making it backwards compatible, it was just the shorter pain. But I'll refactor the CLI commands in this PR now that `codegen` cli is also in, that will also resolve some of the other issues.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: I'll still leave this open. Maybe after refactoring the other CLI commands opinions might change.

packages/overture-schema-core/src/overture/schema/core/tag_providers.py
sethfitz: > I prefer not to reuse anything from `codegen` yet to avoid circular dependancies.

Absolutely. I'm proposing moving the necessary methods into `system` since we may be at a point where we need them elsewhere. I'm also fine holding off until you are clearer on what extensions need--keep me posted!

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: I'll remove this for now as I do agree with the hidden functionality of it. We can add the metadata as part of the `ModelKey` when we have a need for it and an agreement on which values of distribution metadata that should be a part of it. 

As an additional remark, I was initially contemplating on the idea of changing the signature of the `tags` in  to `tuple[str]` and simplify the signature of the `tag_providers` to tag_provider(model_class: type[BaseModel], key: ModelKey) -> tuple[str]` and update the tags in `ModelKey` after each provider. Any opinion on this?

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: CLI commands are refactored to take tags instead of namespaces and themes

packages/overture-schema-cli/tests/test_resolve_types.py
RoelBollens-TomTom: This is now also resolved after refactoring of the CLI commands

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: Well a standalone number might be odd, but i support it could also be something else like overturemaps.org as a tag, which it currently doesn't allow.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
vcschapp: What we discussed this morning in the coding session: AND makes the most sense for CLI list, but there are other use cases where OR fits better.

---

I did some very cursory LLM-driven research about common CLI paradigms that support both AND and OR in filtering. The best example I could come up with were `kubectl` label selectors with `-l` and `docker` filters, but none of them fit our exact use case.

I thought about it a bit and came up with some priorities:

1. CLI should support both AND and OR.
2. Should not use any special shell characters, especially characters like `!` that can get expanded even within double quotes.
3. The most common use cases should be simple.
4. Should not require a lot of typing.

The best idea I can come up with is to allow `--tag` to express an "OR of ANDs" by supporting multiple `--tags` values.

1. `--tag` can be repeated as many times as you want.
6. Each `--tag` supports as plus-sign-separated list. (Alternative: comma.)
7. Within a `--tag` the plus-separated values form an AND expression.
8. The multiple `--tag` arguments are unioned together as a big OR expression.
9. This allows you to express an "OR of ANDs" which is probably powerful enough for most CLI use cases.
10. I would also suggest allowing `*` wildcards to further strengthen the CLI power.

# Examples

Get all models matching single tag, foo.

```
--tag foo
```

Get all models that are both foo AND bar.

```
--tag foo+bar
```

Get all models that are in an Overture theme AND also bar.

```
--tag overture:theme=*+bar
```

Get all models that are either foo OR bar.

```
--tag foo --tags bar
```

Get all models that are either Overture transportation features or Overture places features.

```
--tag feature+overture:theme=transportation --tag feature+overture:theme=places
```

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,340 @@
...
+from typing import Annotated, Any, Literal, TypeAlias, Union, get_args, get_origin
+
+from pydantic import BaseModel
+
+from overture.schema.system.feature import Feature
+
+logger = logging.getLogger(__name__)
+
+
+RESERVED_TAGS: dict[str, set[str]] = {
```
vcschapp: Are these various `CONSTANTS` meant to be part of the public interface or are they internal use only?

If internal use only, can we give them the appropriate underscore prefix?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,340 @@
...
+import re
+from collections.abc import Callable
+from dataclasses import dataclass, replace
+from typing import Annotated, Any, Literal, TypeAlias, Union, get_args, get_origin
+
+from pydantic import BaseModel
+
+from overture.schema.system.feature import Feature
+
+logger = logging.getLogger(__name__)
```
vcschapp: This seems private, should i get the appropriate underscore prefix?

packages/overture-schema-system/src/overture/schema/system/discovery.py
vcschapp: I thought `feature_provider` was intentionally emitting `feature` rather than `system:feature` to be less clunky, per the PR description.

I understood from the PR description that you can reserve both plain tags and namespaces.

Personally I like this approach.

packages/overture-schema-system/src/overture/schema/system/discovery.py
vcschapp: > TAG allows hyphens and underscores (`[a-z0-9_-]`), but `NAMESPACE_TAG`'s key segment doesn't — `acme:my-ext` would fail validation even though `my-ext` is a valid plain tag. Intentional?

To @mojodna's point I would suggest the same pattern should be used for both.

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,340 @@
...
+@dataclass(frozen=True, slots=True)
+class TagProviderKey:
+    """Key identifying a registered model by namespace, theme, and type.
+
+    Attributes
+    ----------
+    name : str
+        The friendly name of the model, derived from the entry point key
+    entry_point : str
+        The entry point value in "module:Class" format
```
vcschapp: Copy/haste?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,340 @@
...
+TagProvider: TypeAlias = Callable[[type[BaseModel], ModelKey, set[str]], set[str]]
+
+ModelDict: TypeAlias = dict[ModelKey, type[BaseModel]]
+
+TagProviderDict: TypeAlias = dict[TagProviderKey, TagProvider]
+
+ModelKeyFilter: TypeAlias = Callable[[ModelKey], bool]
+
+
+def generate_tags(
```
vcschapp: `_private`?

packages/overture-schema-system/src/overture/schema/system/discovery.py
```diff
@@ -0,0 +1,340 @@
...
+    package_name: str  # distribution package name
+
+
+TagProvider: TypeAlias = Callable[[type[BaseModel], ModelKey, set[str]], set[str]]
+
+ModelDict: TypeAlias = dict[ModelKey, type[BaseModel]]
+
+TagProviderDict: TypeAlias = dict[TagProviderKey, TagProvider]
+
+ModelKeyFilter: TypeAlias = Callable[[ModelKey], bool]
```
vcschapp: Are any of these intended to be `_Private`, *e.g.* maybe `TagProviderDict`?

packages/overture-schema-system/src/overture/schema/system/discovery.py
vcschapp: Basically switching the return type to `tuple[str]` from `set[str]`?

I have no opinion on this.

You could go the other way and return `collections.abc.Iterable[str]` - that way anything that can produce a sequence of strings via iteration would suffice.

vcschapp (commented):
Let some comments, but I'm generally aligned and would merge once @RoelBollens-TomTom and @mojodna are jointly aligned on merging.

Left some thoughts on the AND/OR issue in the CLI, probably above there somewhere. 👆

packages/overture-schema-cli/src/overture/schema/cli/commands.py
RoelBollens-TomTom: `--tag foo+bar ` I'm less a fan of, because it introduces a query language in the value. I'd rather keep `--tag` as a simple repeatable flag. But I like adding the wildcard for `namespace:*` and `ns:predicate=*`. 

A git log style `--all-match` is something I considered to switch `--tag` values from OR logic to AND. My concern with that is that some commands may want AND by default, and that still doesn't give a clean way to support both AND and OR on the same command, if that's at all desired.

vcschapp: Alternative ideas:

- `--tag` to include, `--filter` to exclude
- `-i/--include` to include, `-e/--exclude` to exclude 
- existing is `--tags` and `--exclude-tags`

packages/overture-schema-system/src/overture/schema/system/discovery.py
vcschapp: Discussed in the coding sesh and ...

- Namespace and predicate have the same pattern.
- Which is: all lowercase, start alphanumeric, then can be alphanumeric or `-` or `_` or `.`.
- Value can include uppercase and need not start alphanumeric.
- Tag matching/combining/filtering/jiggering is case sensitive.

packages/overture-schema-system/tests/test_tags.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+            self.assertFalse(TAG.fullmatch(tag), f"TAG should not match: {tag}")
+            self.assertFalse(
+                is_valid_tag(tag), f"is_valid_tag should return False for: {tag}"
+            )
+
+
+class TestNamespaceTagRegex(unittest.TestCase):
+    def test_valid_namespace_tags(self) -> None:
+        valid_tags = [
```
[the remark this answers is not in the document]
sethfitz: +1

packages/overture-schema-system/src/overture/schema/system/discovery/discovery.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    tags: set[str] = set()
+    for provider_key, provider in providers.items():
+        try:
+            added_tags = provider(model_class, key, tags.copy()).difference(tags)
+            filtered_tags = _filter_tags(added_tags, provider_key)
+            tags.update(filtered_tags)
+        except Exception as e:
+            log.warning(
+                f"Error in tag provider {provider.__name__} for model {key.name}: {e}"
```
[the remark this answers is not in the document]
sethfitz: +1

packages/overture-schema-core/src/overture/schema/core/tag_providers.py
sethfitz, line 1:
```console
make docformat
...
packages/overture-schema-core/src/overture/schema/core/tag_providers.py:1 at module level:
        D100: Missing docstring in public module
packages/overture-schema-core/src/overture/schema/core/tag_providers.py:28 in public function `authority_provider`:
        D103: Missing docstring in public function
packages/overture-schema-core/src/overture/schema/core/tag_providers.py:36 in public function `theme_provider`:
        D103: Missing docstring in public function
```

packages/overture-schema-system/src/overture/schema/system/discovery/discovery.py
sethfitz, line 1:
```console
make docformat
...
packages/overture-schema-system/src/overture/schema/system/discovery.py:78 in public function `generate_tags`:
        D103: Missing docstring in public function
packages/overture-schema-system/src/overture/schema/system/discovery.py:143 in public function `discover_tag_providers`:
        D103: Missing docstring in public function
packages/overture-schema-system/src/overture/schema/system/discovery.py:301 in public function `feature_provider`:
        D103: Missing docstring in public function
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -190,62 +195,34 @@ def validate_features(data: list, model_type: UnionType) -> list[BaseModel]:
...
 ) -> UnionType:
     """Resolve CLI options into a model type suitable for parse_feature.
 
     Args
     ----
-        use_overture_types: Boolean from --overture-types flag
-        namespace: Namespace to filter by (e.g., "overture", "annex")
-        theme_names: List of theme names from --theme option
+        tags: Tags to include (e.g., "feature", "overture:theme=buildings")
+        excluded_tags: Tags to exclude (e.g., "draft")
```
sethfitz: These are tags *to* include/exclude; they're tags to use when filtering models.

```suggestion
        tags: Include models with these tags (e.g., "feature", "overture:theme=buildings")
        excluded_tags: Exclude models with these tags (e.g., "draft")
```

`tags` implies "or" in the wording, but as we've discovered, that's hard to track (and I'd use the docstring to infer which behavior is used here), so let's be explicit.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
sethfitz: Sorry, you inherited some strangely worded advice. Here's an attempt to resolve it.

```suggestion
            "  ⚠ Heterogeneous collection: Data contains multiple feature types. Consider:",
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -527,7 +504,7 @@ def handle_validation_error(
             style="yellow",
         )
         stderr.print(
-            "    • Consider validating each type separately with --theme or --type",
+            "    • Consider validating each type separately with --tag or --type",
```
sethfitz:
```suggestion
            "    • Validating each type separately with --tag or --type",
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -628,18 +605,16 @@ def handle_generic_error(e: Exception, filename: Path, error_type: str) -> None:
...
-    is_flag=True,
-    help="Validate against all official Overture types (excludes extensions)",
-)
-@click.option(
-    "--namespace",
-    help="Namespace to filter by (e.g., overture, annex)",
+    "--tag",
+    "tags",
+    multiple=True,
+    help="Tags to include (e.g., overture:theme=addresses)",
```
sethfitz: Same problem as above.

```suggestion
    help="Include models with these tags (e.g., overture:theme=addresses)",
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -628,18 +605,16 @@ def handle_generic_error(e: Exception, filename: Path, error_type: str) -> None:
...
 )
 @click.option(
-    "--theme",
+    "--exclude-tag",
+    "excluded_tags",
     multiple=True,
-    help="Theme to validate against (shorthand for all types in theme)",
+    help="Tags to exclude (e.g., overture:theme=base)",
```
sethfitz:
```suggestion
    help="Exclude models with these tags (e.g., overture:theme=base)",
```

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -748,71 +719,47 @@ def json_schema_command(
...
-
-
 @cli.command("list-types")
-def list_types() -> None:
-    r"""List all available types grouped by theme with descriptions.
+@click.option(
+    "--tag",
+    "tags",
+    multiple=True,
+    help="Filter types by tag (e.g., overture:theme=addresses)",
```
sethfitz: "Filter" implies "and"; let's make the actual behavior clearer in the help example.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
```diff
@@ -748,71 +719,47 @@ def json_schema_command(
...
+@click.option(
+    "--exclude-tag",
+    "excluded_tags",
+    multiple=True,
+    help="Exclude types by tag (e.g., overture:theme=base)",
+)
+@click.option(
+    "--group-by",
+    help="Group types by tag key (e.g., 'overture:theme')",
+)
```
sethfitz, line 718: Does this require a k/v-style tag? If so, let's elaborate on the help and make that clearer.

packages/overture-schema-cli/tests/test_resolve_types.py
```diff
@@ -1,168 +1,172 @@
...
+    BUILDING_KEY: Building,
+    SEGMENT_KEY: Segment,
+    CONNECTOR_KEY: Connector,
+    PLACE_KEY: Place,
+    SOURCES_KEY: Sources,
+}
+
+
+class TestResolveTypes:
     @pytest.mark.parametrize(
```
sethfitz: We should include some parameters that include more than 1 value for `tags` and `excluded` tags to assert the behavior we intend.

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
```diff
@@ -76,9 +76,16 @@ def list_models() -> None:
     help="Output format",
 )
 @click.option(
-    "--theme",
+    "--tag",
+    "tags",
     multiple=True,
-    help="Filter to specific theme(s); repeatable (e.g., --theme buildings --theme places)",
+    help="Tag(s) to include; repeatable (e.g., --tag feature --tag overture)",
```
sethfitz: "Model" should be the subject, like in the other CLI.

packages/overture-schema-codegen/src/overture/schema/codegen/cli.py
```diff
@@ -76,9 +76,16 @@ def list_models() -> None:
...
+)
+@click.option(
+    "--exclude-tag",
+    "excluded_tags",
+    multiple=True,
+    help="Tag(s) to exclude; repeatable (e.g., --exclude-tag draft --exclude-tag overture:theme=base)",
```
sethfitz: "Model" should be the subject, like in the other CLI.

packages/overture-schema-cli/src/overture/schema/cli/commands.py
sethfitz: I'm still struggling to wrap my head around whether AND or OR is used because I don't have a mental model that leads to expectations. I *think* I expect it to differ by command. However, in the process of writing this out, my current inclination is:

- `--tag` treated as OR to create a preliminary filter / result.
- `--filter` (or `--include` for symmetry with `--exclude`, although the AND/OR inconsistency is trouble) treated as AND, applied to the preliminary result with AND to *positively* filter using large sets.
- `--exclude` treated as OR, applied to the preliminary result with AND to filter out smaller sets.

The idea is that `--tag` does an initial select to reduce "everything" to an inclusive result, and then `--filter` draws a smaller circle and `--exclude` does a targeted job of removing undesired models.

```console
# I want to expand my criteria once I commit to filtering
# OR
overture-schema list-types \
  --tag overture:theme=places \
  --tag overture:theme=buildings

# I want to (positively) filter the preliminary results (from ORed tags) to datasets licensed under CDLA 2.0
# OR, with ANDed filters applied
overture-schema list-types \
  --tag overture:theme=places \
  --tag overture:theme=base \
  --filter license=CDLA-Permissive-2.0

# I want to (negatively) filter out ODbL-licensed datasets
# OR, with ORed exclusions applied as filters
overture-schema list-types \
  --tag overture:theme=base \
  --tag overture:theme=divisions \
  --exclude license=ODbL-1.0


# I want to produce an increasingly broad JSON Schema equivalent
# OR
overture-schema json-schema \
  --tag overture:theme=places \
  --tag overture:theme=buildings


# I want to select a narrow set of models to validate against
# AND, using --filter instead of --tag
overture-schema validate \
  --filter overture:theme=base \
  --filter source_type=raster


# I want to generate docs for an arbitrary set of models
# OR; AND would be too limiting
overture-schema-codegen generate --format markdown --output-dir /tmp/overture \
  --tag overture:theme=places \
  --tag overture:theme=buildings
```

sethfitz: Attempting to clarify my idea further (and use set algebra to describe it):

1. **`--tag` defines the scope.** Without `--tag`, the scope is all models. With one or more `--tag` flags, the scope is the union of models matching any listed tag — `--tag X --tag Y` keeps models tagged X or Y.
2. **`--filter` narrows the scope.** Each `--filter` predicate adds a requirement every model must satisfy — `--filter X --filter Y` keeps only models matching both (**AND**).
3. **`--exclude` removes from the scope.** Models matching any listed exclusion are dropped — `--exclude X --exclude Y` drops models matching X **OR** Y.

So `--tag` is OR, `--filter` is AND, `--exclude` is OR-then-subtract.

Tag-based discovery is the primary entry point. OR is the right default for `--tag` — users reach for it to declare interest in any of several themes, and is what we concluded while working through this together. `--filter` is for the stricter case where every result must satisfy an additional requirement.

Equivalent set algebra: `result = (⋃ tags) ∩ (⋂ filters) \ (⋃ excludes)`, with absent classes
contributing no restriction.

sethfitz: Drafted an implementation in 1ed1ce636ad3

packages/overture-schema-system/src/overture/schema/system/discovery/discovery.py
```diff
@@ -0,0 +1,288 @@
...
+    -------
+    set[str]
+        Tags generated for the model.
+    """
+    tags: set[str] = set()
+    for provider_key, provider in providers.items():
+        try:
+            added_tags = set(provider(model_class, key, tags.copy())) - tags
+            filtered_tags = _filter_tags(added_tags, provider_key)
+            tags.update(filtered_tags)
```
sethfitz: Dropped tags (when the tag provider manipulates the tag set) are silently ignored, so tag providers may only **add** tags. While I think this is desired behavior, the tag provider signature makes it look like they can be removed.

Thoughts?

sethfitz (approved):
Well, *I'm* happy with the state of the PR, but that's because I worked through my reactions to it and pushed a series of follow-up commits. (Not to say that it wasn't great before, but I took liberties with actually responding to my own comments by making changes. Mostly specific to filtering and supporting the nuances of union types like `Segment`.)

It's definitely worth others looking over the `TagSelector` business, particularly how they're handled (and what the CLI UX looks like), in addition to the other follow-ups I made.

packages/overture-schema-system/src/overture/schema/system/discovery/discovery.py
RoelBollens-TomTom: I left the option open for having tag providers that could remove tags as this was mentioned in the initial tag design document (see #440). I never quite saw a good use case for it though, which is why it's left unimplemented.

packages/overture-schema-system/src/overture/schema/system/discovery.py
RoelBollens-TomTom: Fixed.

sethfitz (dismissed):

vcschapp (approved):

sethfitz (approved):

PR #468, opened 2026-03-11
[SECURITY] Remove vulnerabilities from GH workflows

### LINKED ISSUE #467: [Devops] Remediate GitHub workflow security concerns
## `pull_request_target` misuse and credential exposure in Python workflows

Due to the high number of ongoing AI-enabled attacks on public GitHub repositories, we need to address the following identified issues in this repository related to `pull_request_target` and credentials.

#### Privileged workflow execution via `pull_request_target` (high)

`check-python-code.yaml` and `check-python-package-versions.yaml` use `pull_request_target` with no legitimate need. This trigger runs workflows from the base branch with repository secrets and elevated permissions, including an OIDC token capable of assuming AWS IAM roles. 

Any repo contributor with branch access could open a PR from a modified branch to execute arbitrary code in this privileged context - potentially assuming the `GithubActions_Schema_CodeArtifact_Publish` role to push malicious packages to the CodeArtifact repository.

#### Unmasked credentials in workflow logs (medium)

`reusable-check-python-package-versions.yaml` writes the CodeArtifact index_url (which embeds an auth token as `aws:<token>@hostname`) to `$GITHUB_OUTPUT` without masking, then interpolates it directly into a `run:` step. 

Similarly, `publish-python-packages.yaml` writes the CodeArtifact auth token to `$GITHUB_OUTPUT` unmasked before use. Both values will appear in plaintext in workflow logs.

lowlydba:
## Context

See https://github.com/OvertureMaps/schema/issues/467 for findings that we need to remediate.

## Changes

- 🔒 Replace `pull_request_target` with `pull_request` in `check-python-code.yaml` and `check-python-package-versions.yaml` - neither workflow requires the elevated permissions that `pull_request_target` grants, and its use created a path for contributors to execute arbitrary code with AWS IAM credentials
   - An artificial touch to one of the project files triggered these modified jobs as a sanity check
   
      {CF93939C-AB77-4E3D-AD47-FE83325F6D8F}
 
- 🔒 Remove the now-redundant same-repo `if` guards from both workflows
- 🔒 Mask CodeArtifact auth tokens before writing to `$GITHUB_OUTPUT` in `reusable-check-python-package-versions.yaml` and `publish-python-packages.yaml`
- 🔒 Pass sensitive values via `env:` into consuming steps rather than direct `${{ steps.*.outputs.* }}` interpolation in `run:` scripts
- ♻️Updated blessed GHA providers (`github`, `astral`) to current major versions
   - Configure dependabot to keep GHA updated going forward
- 🧑‍🦲 Stubbed out a CODEOWNERS
   - We'll spend more time figuring out more complex ownership paths, but this gets the initial file committed and sets myself & public reviewers for the `.github` directory to ensure smooth, safe changes to CI/CD for this public repository
   - This is **not** a required reviewer at this time, or scheduled for a future date to become one


.github/workflows/test-schema.yaml
```diff
@@ -14,12 +14,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v6
     - name: Set up Go
-      uses: actions/setup-go@v5
+      uses: actions/setup-go@v6
       with:
         go-version: 1.22
```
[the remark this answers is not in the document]
lowlydba: Fixed

vcschapp:
Thanks for this John. I'll get chance to look at it eventually, but help me with the first bullet:

> - 🔒 Replace `pull_request_target` with `pull_request` in `check-python-code.yaml` and `check-python-package-versions.yaml` - neither workflow requires the elevated permissions that `pull_request_target` grants

Let's make sure this is true. The reason `pull_request_target` was chosen is that the workflows are *intended* to use the repo base commit - it's how they determine the base versions of the packages and whether they have changed. If they instead use the `pull_request` base commit, it's not clear to me that they'll be able to identify the right versions.

> and its use created a path for contributors to execute arbitrary code with AWS IAM credentials

The IAM roles that can be assumed are very defensively scoped; one is read-only, and the only thing the other can do is publish to the Python sub-repo of CodeArtifact.

.github/workflows/publish-python-packages.yaml
```diff
@@ -64,16 +65,18 @@ jobs:
...
-          ./.github/workflows/scripts/code-artifact.sh token \
-            505071440022 us-west-2 overture-pypi >> $GITHUB_OUTPUT
-          echo EOF >> $GITHUB_OUTPUT
-          echo 'publish_url<<EOF' >> $GITHUB_OUTPUT
-          ./.github/workflows/scripts/code-artifact.sh publish-url \
-            505071440022 us-west-2 overture-pypi overture >> $GITHUB_OUTPUT
-          echo EOF >> $GITHUB_OUTPUT
+          token=$(./.github/workflows/scripts/code-artifact.sh token \
+            505071440022 us-west-2 overture-pypi)
+          echo "::add-mask::${token}"
```
vcschapp, line 70: Does this relate to:

> - 🔒 Mask CodeArtifact auth tokens before writing to `$GITHUB_OUTPUT` in reusable-check-python-`package-versions.yaml` and `publish-python-packages.yaml`

Wouldn't that just print a masked version of it to the standard output?

lowlydba: It does do that, but the real value is that it is treated thereafter as if it were a secret. Any downstream step, script, etc. that tried to print the token would result in the same masked output, thereby covering us for accidental and malicious attempts to leak (even temporary cred) secrets. 

Longer term, relying more on GitHub Actions instead of shell scripts will reduce the need for these, but that will be future enhancements.

lowlydba:
> > * 🔒 Replace `pull_request_target` with `pull_request` in `check-python-code.yaml` and `check-python-package-versions.yaml` - neither workflow requires the elevated permissions that `pull_request_target` grants
> 
> Let's make sure this is true. The reason `pull_request_target` was chosen is that the workflows are _intended_ to use the repo base commit - it's how they determine the base versions of the packages and whether they have changed. If they instead use the `pull_request` base commit, it's not clear to me that they'll be able to identify the right versions.

Yeah, that can be achieved without using an elevated context that `pull_request_target` gives. The workflow already bifurcates its work across before/after commits:

https://github.com/OvertureMaps/schema/blob/39a24f258d8259d5e5c6866a4031daf9e36da159/.github/workflows/reusable-check-python-package-versions.yaml#L69-L72

https://github.com/OvertureMaps/schema/blob/39a24f258d8259d5e5c6866a4031daf9e36da159/.github/workflows/reusable-check-python-package-versions.yaml#L85-L88

There are a few use cases for `pull_request_target` but (thankfully) it isn't required here!

.github/workflows/publish-python-packages.yaml
vcschapp: Nice.

vcschapp:
> The workflow already bifurcates its work across before/after commits

Right, but the content of the *before* commit is important. It should reflect the state of the current state of the branch in the official repo, not any state that comes with the PR. Can you confirm that's true?

lowlydba:
> > The workflow already bifurcates its work across before/after commits
> 
> Right, but the content of the _before_ commit is important. It should reflect the state of the current state of the branch in the official repo, not any state that comes with the PR. Can you confirm that's true?

Yes, this is the core functionality of GitHub's checkout action and in this case is not affected at all by the change in event trigger:

* Check out code before changes: [7162a0de6ea9317102cd28176ee441c9b3850a8c](https://github.com/OvertureMaps/schema/actions/runs/23005561029/job/66801305086?pr=468#step:4:101) --> https://github.com/OvertureMaps/schema/pull/466 which was the last merged PR as of writing
* Check out code after changes: [5f8baeea507117856eba0b12708a0007c52de3c4](https://github.com/OvertureMaps/schema/actions/runs/23005561029/job/66801305086?pr=468#step:8:111) --> https://github.com/OvertureMaps/schema/pull/468/commits/5f8baeea507117856eba0b12708a0007c52de3c4 Last commit on this PR

Rachmanin0xFF (approved):
LGTM; I'm no workflow wizard but it seems like Vic's concerns were addressed + the workflow updates make sense

ericgodwin (approved):

lowlydba:
@vcschapp I've added some specifics about your concerns - can you take another look? Happy to set up some time to review in person if that is helpful.

PR #485, opened 2026-03-31
[WIP] Add Bbox primitive directly to Feature next to geometry

jenningsanderson:
# Description

Testing the addition of `bbox` directly into the Overture Feature to see how it propagates to the documentation. 

...The `bbox` primitive is different from the GeoParquet `bbox` that we create, so we'll need to re-think this — but great to see how a 1 line addition flows so nicely from schema to schema-docs!

### DIFF
--- a/packages/overture-schema-core/src/overture/schema/core/models.py
+++ b/packages/overture-schema-core/src/overture/schema/core/models.py
@@ -1,34 +1,20 @@
 import textwrap
 from typing import Annotated, Generic, TypeVar
 
-from pydantic import (
-    BaseModel,
-    ConfigDict,
-    Field,
-    GetJsonSchemaHandler,
-    model_validator,
-)
+from overture.schema.system.feature import Feature
+from overture.schema.system.field_constraint import UniqueItemsConstraint
+from overture.schema.system.model_constraint import no_extra_fields
+from overture.schema.system.primitive import BBox, Geometry
+from overture.schema.system.ref import Id, Identified
+from overture.schema.system.string import CountryCodeAlpha2
+from pydantic import BaseModel, ConfigDict, Field, GetJsonSchemaHandler, model_validator
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import core_schema
 from typing_extensions import Self
 
-from overture.schema.system.feature import Feature
-from overture.schema.system.field_constraint import UniqueItemsConstraint
-from overture.schema.system.model_constraint import no_extra_fields
-from overture.schema.system.primitive import (
-    Geometry,
-)
-from overture.schema.system.ref import Id, Identified
-from overture.schema.system.string import (
-    CountryCodeAlpha2,
-)
-
 from .enums import PerspectiveMode
 from .sources import Sources
-from .types import (
-    FeatureVersion,
-    Level,
-)
+from .types import FeatureVersion, Level
 
 ThemeT = TypeVar("ThemeT", bound=str)
 TypeT = TypeVar("TypeT", bound=str)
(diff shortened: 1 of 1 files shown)


vcschapp:
Do we need to do this? My understanding is the CDP is now using the Pydantic version of the schema.

I think we can be out of the business of updating the JSON Schema now?

---

Given the shapeshifting nature of `bbox` structure across data formats, it's ideal if we can avoid codifying a single format-dependent variant. The Pydantic version tries to keep `bbox` format-agnostic.

jenningsanderson:
> Do we need to do this? My understanding is the CDP is now using the Pydantic version of the schema.
> 
> I think we can be out of the business of updating the JSON Schema now?
> 
> Given the shapeshifting nature of `bbox` structure across data formats, it's ideal if we can avoid codifying a single format-dependent variant. The Pydantic version tries to keep `bbox` format-agnostic.

Sorry for the complete lack of context on this one — let me shed some light: 

As we move away from the `conceptual GeoJSON schema` to the what-you-see-is-what-you-get mode, I think we should ensure that the `bbox` attribute actually shows up in the schema documentation — not the primitive one referenced [here](https://docs.overturemaps.org/schema/reference/system/primitive/geometry/), but the actual implementation of it as a struct of `xmin`, `ymin`, `xmax`, `ymax`. While this is really a convenience column added to the data in the central data pipeline, it is the _only_ column present in the release parquet that's not discussed in the schema docs explicitly. 

I don't think we should alter the `bbox` primitive that's part of a Geometry, as this is clearly foundational to the geometry — but instead introduce a new concept of an Overture Feature's bbox, built from this bbox primitive where we can define things like precision, data type, etc. 

If defined in Pydantic, this would be accessible to the data pipeline validations and prevent future schema-related errors. 

Bonus: This also gives us an opportunity to define what makes a bounding box valuable across themes. Since it's primarily for data-filtering, does it makes sense to clamp it to some arbitrary precision? In many cases, it's probably the most-read column, which means going across the wire — maybe it doesn't need centimeter level precision? Does that impact query time?

vcschapp:
> > Do we need to do this? My understanding is the CDP is now using the Pydantic version of the schema.
> > I think we can be out of the business of updating the JSON Schema now?
> > Given the shapeshifting nature of `bbox` structure across data formats, it's ideal if we can avoid codifying a single format-dependent variant. The Pydantic version tries to keep `bbox` format-agnostic.
> 
> Sorry for the complete lack of context on this one — let me shed some light:
> 
> As we move away from the `conceptual GeoJSON schema` to the what-you-see-is-what-you-get mode, I think we should ensure that the `bbox` attribute actually shows up in the schema documentation — not the primitive one referenced [here](https://docs.overturemaps.org/schema/reference/system/primitive/geometry/), but the actual implementation of it as a struct of `xmin`, `ymin`, `xmax`, `ymax`. While this is really a convenience column added to the data in the central data pipeline, it is the _only_ column present in the release parquet that's not discussed in the schema docs explicitly.
> 
> I don't think we should alter the `bbox` primitive that's part of a Geometry, as this is clearly foundational to the geometry — but instead introduce a new concept of an Overture Feature's bbox, built from this bbox primitive where we can define things like precision, data type, etc.
> 
> If defined in Pydantic, this would be accessible to the data pipeline validations and prevent future schema-related errors.
> 
> Bonus: This also gives us an opportunity to define what makes a bounding box valuable across themes. Since it's primarily for data-filtering, does it makes sense to clamp it to some arbitrary precision? In many cases, it's probably the most-read column, which means going across the wire — maybe it doesn't need centimeter level precision? Does that impact query time?

`bbox` is defined in Pydantic [as a primitive](https://github.com/OvertureMaps/schema/blob/dev/packages/overture-schema-system/src/overture/schema/system/primitive/bbox.py).

Because the aim is to keep the schema portable across data formats (including GeoJSON), the "primitive" `Bbox` explicitly does not state what its format-dependent structure is, same as `Geometry`. If we want people to know the Parquet structure, I think what we can do is write up the canned content of the docs page for `Bbox` in a way that explains what it looks like in different formats.

vcschapp:
@jenningsanderson any objections to closing this? Let me know if you want to discuss.

vcschapp:
Closing this PR as stale.

PR #505, opened 2026-05-04
[CHORE](ci) Test dep floors via lowest-direct matrix

### LINKED ISSUE #504: Test declared dependency floors via lowest-direct CI matrix
## Problem

The Python packages in this repo declare dependency floors (e.g.
\`pydantic>=2.0\`, \`pyspark>=3.4\`), but CI never exercises those
floors. \`uv.lock\` pins to the newest version permitted by each
range, and \`uv sync\` follows the lock, so contributors and CI test
against the latest versions only.

That makes the declared floors aspirational rather than verified. As
the code evolves, the actual minimum drifts upward — a feature added
to \`pydantic\` 2.12 gets used while \`pyproject.toml\` still claims
\`>=2.0\` is supported. A user installing with \`--resolution=lowest-direct\`
or with the lower bound otherwise pinned will hit ImportErrors.

## Proposal

Add a CI matrix cell that re-resolves to lowest-direct before running
\`make check\`:

- \`UV_RESOLUTION=lowest-direct\` makes the existing \`make uv-sync\`
  resolve each direct dependency to its declared floor instead of
  following the committed lock.
- A failure in that cell means a declared minimum is wrong — either
  the code uses a feature newer than the floor admits, or the pin
  needs to be raised.
- Pair this with a Python minor-version matrix (3.10–3.14) on the
  default cell to catch Python-version-specific regressions.

## Verified gaps

A local run of \`UV_RESOLUTION=lowest-direct make check\` against the
declared floors today fails on:

- \`pydantic>=2.0\` — \`optionality.py\` imports
  \`pydantic.experimental.missing_sentinel\` (added in 2.12);
  \`feature.py\` imports \`ModelWrapValidatorHandler\` at the top level
  (re-exported in 2.10.4) and \`Tag\` (added in 2.5).
- \`ruff>=0.12.4\` — fires the deprecated \`UP038\` rule under
  \`select = ["UP"]\`; rule was removed in 0.13.
- \`deepdiff>=8.0\` — first version shipping a \`py.typed\` marker
  (which mypy needs) is 8.6.
- Several per-package dev deps (\`ruff\`, \`mypy\`, \`pytest\`) are
  unpinned, which uv warns about under lowest-direct resolution.

sethfitz:
## Summary

- Adds a CI matrix cell that re-resolves to `lowest-direct` before `make check`, exercising the declared dependency floors instead of whatever the lock happens to pin.
- Adds a Python minor-version matrix (3.10–3.14) for the default cell so version-specific regressions surface.
- Bumps the floors that today's code already requires (pydantic ≥2.12, ruff ≥0.13, deepdiff ≥8.6) and pins per-package dev deps that were unpinned.

Closes #504.

## How it works

`UV_RESOLUTION=lowest-direct` makes the existing `make uv-sync` re-resolve from `pyproject.toml` instead of following the committed lock. No second lockfile, no constraints file. A failure in that cell means a declared minimum is wrong.

The default cell behaves as before (committed lock, newest matching) but now spans Python 3.10 through 3.14.

## Test plan

- [x] `UV_RESOLUTION=lowest-direct make check` passes locally (1900 tests)
- [x] `make check` (default resolution) passes locally
- [ ] CI: both matrix cells green across Python 3.10–3.14


.github/workflows/check-python-code.yaml
```diff
@@ -22,9 +22,22 @@ concurrency:
...
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
+        # Default resolution exercises the committed lock against every
+        # supported Python minor version. The lowest-direct cell pins each
+        # direct dependency to its declared floor (see UV_RESOLUTION below)
+        # and runs only on the Python floor.
+        python: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
lowlydba, line 35: question: how do you feel about dropping 3.10, 3.11 since they're out of Active Support?

lowlydba (approved):
Left a question on how far back we want to support non-Active versions, otherwise LGTM

.github/workflows/check-python-code.yaml
sethfitz: 3.10, potentially. FWIW, the guidance I followed when first picking a minimum Python version (last summer, so tick forward a year/version) was that (at the time) 3.10 was appropriate for libraries, esp. since many of the potential (but only potential today) users are in managed environments with old Python versions. We don't know what this looks like for our potential users, so I'd rather be conservative and drop support once we're confident that we're not locking out some of the potential users for this.

lowlydba: Thanks for the context - makes total sense.

vcschapp (approved):

PR #518, opened 2026-05-11
[FEATURE](pyspark) PySpark validation generated from the Pydantic schema

### LINKED ISSUE #517: Validate Overture data on Spark against the Pydantic schema
## Problem

We need a validator that:

## Proposal

### CLI

```bash
```

### Conformance gate

## Non-goals

## How to experience it

## Acceptance

sethfitz:
## Summary

Adds a new runtime package (`overture-schema-pyspark`) plus a new output target in `overture-schema-codegen` that emits PySpark validation expressions and conformance tests from the same Pydantic models that define the schema. Ships an `overture-validate` CLI (to be merged with the `overture-schema` CLI later; that will be an opportunity to improve its ergonomics) for running the validation against Parquet on disk or in S3.

PySpark plugs in as a peer of the existing Markdown output target: same `ModelSpec` extraction, same four-layer architecture (Discovery -> Extraction -> Output Layout -> Rendering), new pipeline module. The codegen abstraction is model-general (`ModelSpec = RecordSpec | UnionSpec`: one record that compiles to a Spark `StructType`, or a tagged union of records); in this schema every model is an Overture feature, so the generated modules are organized by theme/feature. See `packages/overture-schema-codegen/docs/design.md` for the full picture; the "PySpark Pipeline" section there covers the new stages in detail.

## What's in the PR

**`packages/overture-schema-pyspark/`** -- runtime. Public API in `validate.py` (`validate_model`, `explain_errors`), schema comparison in `schema_check.py`, dataclasses in `check.py`, the `overture-validate` CLI in `cli.py`, and shared expression building blocks in `expressions/{constraint_expressions,column_patterns,_schema_structs}.py`. The per-feature expression modules under `expressions/generated/overture/schema/<theme>/<feature>.py` and per-feature conformance tests under `tests/generated/overture/schema/<theme>/test_<feature>.py` are emitted by codegen and confined to a `generated/` boundary that `make generate-pyspark` wipes and recreates. `_registry.py` walks that tree at import time and exposes `REGISTRY: dict[str, ModelValidation]` keyed by each module's `ENTRY_POINT` value (e.g. `"overture.schema.transportation:Segment"`), reading each module's emitted `MODEL_VALIDATION` constant, plus a parallel `PARTITION_MAP` derived from each module's `PARTITIONS` dict for partitioned features.

Validation degrades gracefully when columns are skipped (`--skip-columns` / `--suppress`) or structurally absent from the data. Every check declares the top-level columns it reads, and a check over a missing column -- top-level, nested inside an array or map (`sources[].confidence` resolves to `sources`), or model-level -- is dropped before Spark planning rather than raising a raw `AnalysisException`. `ValidationResult.absent_columns` surfaces the columns dropped for this reason, and the CLI backstop classifies any residual planning error through the structured PySpark error API so a generator bug propagates as a traceback instead of being mistaken for a missing column.

This has been tested to work with Spark versions 3.4.0 - 4.1.1 (and the `lowest-direct` CI check will verify that it continues to work with the lowest declared PySpark version, which is currently `3.4`).

@Rachmanin0xFF the public API changed since the previous version you tested (the entry point is now `validate_model`, returning a `ValidationResult`, in an attempt to simplify how it gets integrated with the CDP). If you point me to its current state, I can propose an update that uses the new API.

**`packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/`** -- new output target. Pipeline stages:

```text
ModelSpec (from extraction)
    |
constraint_dispatch.py    constraints -> ExpressionDescriptor / ModelConstraintDescriptor
    |
check_builder.py          FieldShape tree -> Check / ModelCheck IR
                          (check_ir.py; resolves array nesting, map key/value
                          projection, variant gating via
                          FieldPath = ScalarPath | ArrayPath | MapPath and Guard
                          sum types)
schema_builder.py         FieldShape tree -> SchemaField list (StructType source)
test_data/                ModelSpec -> BASE_ROW_SPARSE / BASE_ROW_POPULATED, scaffold,
                          invalid_value
    |
renderer.py               Check IR -> per-feature expression module
test_renderer.py          Check IR -> per-feature conformance test module
    |
pipeline.py               orchestrates, returns GeneratedModule list
```

`make generate-pyspark` wipes both `generated/` trees and recreates them; `make check` gates on regeneration being current.

## What's covered

Every constraint Pydantic enforces today is dispatched to a PySpark expression, which allows Spark to push down predicates and take advantage of Catalyst to avoid unnecessary deserialization:

- Field constraints: `Ge` / `Gt` / `Le` / `Lt` / `Interval`, `ArrayMinLen` / `ArrayMaxLen` / `ScalarMinLen` / `ScalarMaxLen` (typed length constraints split by attachment layer), `StrippedConstraint`, `PatternConstraint`, `UniqueItemsConstraint`, `GeometryTypeConstraint`, `JsonPointerConstraint`.
- Map key/value constraints: dict-typed fields (e.g. `names.common`, a `dict[LanguageTag, StrippedString]` present on most named features) have their per-key and per-value constraints validated by projecting `map_keys` / `map_values` and applying the same field checks. A map value that is a sub-model is descended into, validating its field and model-level constraints. `dict[K, Any]` maps (e.g. `Infrastructure.source_tags`) carry no constraint and emit no checks; container-valued map values (a list or map inside a map value) are not yet supported and raise at generation time rather than emit an unanchored check.
- NewType overrides: `LinearlyReferencedRange` (length / bounds / order).
- Base-type overrides: `HttpUrl` (format + length), `EmailStr`, `BBox` (completeness, lat ordering, lat range).
- Model constraints: `RequireAnyOfConstraint`, `RadioGroupConstraint`, `RequireIfConstraint`, `ForbidIfConstraint`, `MinFieldsSetConstraint`. `MinFieldsSetConstraint` mirrors Pydantic's `model_fields_set` semantics: required fields are always set (the constructor enforces them) and count toward the threshold alongside any explicitly-set optional fields. `NoExtraFieldsConstraint` is intentionally skipped, as its behavior is implicit when using DataFrames.

## Known semantic gaps / differences

Divergences from Pydantic:

- `UniqueItemsConstraint` uses Spark's `array_distinct`, which compares whole elements with structural equality on raw stored values. Pydantic compares normalized Python objects -- e.g., `list[HttpUrl]` is compared after URL normalization (such as ensuring that URLs like `http://example.com` contain a trailing `/`: `http://example.com/`). The PySpark check catches exact duplicates, not duplicates after normalization.
- `require_any_of` checks `isNotNull` as a proxy for Pydantic's `model_fields_set`. Parquet has no equivalent of "explicitly provided" because of the tabular nature of DataFrames; `isNotNull` is stricter (it rejects fields explicitly set to null).
- `BBox`'s PySpark expressions include checking latitude ordering and range. Longitude is skipped due to undefined behavior around how we handle features that cross the antimeridian.
- Regex checks run on the JVM, so a `PatternConstraint` is evaluated with Java's regex dialect rather than Python's -- the two agree for the schema's patterns, but ASCII-vs-Unicode shorthand classes (`\w`, `\d`) and interior-newline handling differ in general.

## CLI

```console
$ overture-validate --help
Usage: overture-validate [OPTIONS] FEATURE_TYPE PATH

  Validate Overture data at PATH and write annotated Parquet.

Options:
  -o, --output TEXT            Output path for validated Parquet.
  --head INTEGER               Error rows to display.  [default: 20]
  --conf TEXT                  Spark config key=value pairs.
  --count-only                 Report error count only; skip explain/unpivot.
  --skip-schema-check          Warn on schema mismatches instead of aborting.
  --skip-columns TEXT          Columns declared absent from data; skips their
                               checks.
  --ignore-extra-columns TEXT  Extra data columns to ignore in schema
                               comparison.
  --suppress TEXT              Suppress checks: FIELD (all checks) or
                               FIELD:CHECK (specific).
  --help                       Show this message and exit.
```

Output is one row per violation: feature ID, theme/type, failing field, check name, message, offending value.

## Testing

`overture-schema-codegen` generates conformance tests alongside the PySpark expressions as a sanity-check by evaluating expected-good and expected-bad rows within a single DataFrame (setup/teardown overhead for DataFrame-per-test-case was much too high). Each feature gets two baseline rows -- `BASE_ROW_SPARSE` (required fields only) and `BASE_ROW_POPULATED` (every optional field filled in) -- and every scenario is exercised against both, so checks that depend on optional structure being present are covered as well as the minimal-row case.

`make check` generates tests and runs the full suite over everything, which use PySpark and increase test runtime (again; they run in just over a minute for me). [`testmon`](https://www.testmon.org/) was introduced to improve the developer experience by skipping tests that weren't affected by recent edits.

Beyond the unit and conformance tests:

```bash
# Local Parquet
overture-validate segment samples/segment.parquet --count-only

# Real release prefix (expect bbox Float-vs-Double mismatch -> use --skip-schema-check)
# partitions (theme=places, type=place) will automatically be detected
# note that the s3a protocol must be used when AWS-provided Hadoop filesystem JARs are not available
overture-validate place s3a://overturemaps-us-west-2/release/<release>/ --skip-schema-check --head 50
```

## Notes for review

- This PR is intentionally large because the generated tree is large. The interesting surface is smaller: `pyspark/{constraint_dispatch,check_builder,check_ir,schema_builder,renderer,test_renderer,pipeline}.py` plus the path IR in `overture-schema-system`'s `field_path.py` and the runtime in `overture-schema-pyspark/src/overture/schema/pyspark/`. Everything under `generated/` is regenerable output -- review the codegen, but skim the output to understand the shape of what's produced.
- Bundles supporting changes (testmon for incremental test runs, `VehicleSelectorBase` extraction, Java 17 CI pin, build/CI hardening so `test-all` runs unconditionally and the check job triggers on the Makefile/workflow themselves, `typing-extensions` declared) rather than splitting them into separate PRs.
- The branch has been squashed to a single commit; expect a force-push. The review/refactoring/hardening pass is complete.

Closes #517.


vcschapp:
> This PR is intentionally **large because the generated tree is large**. The interesting surface is smaller: `pyspark/{constraint_dispatch,check_builder,check_ir,schema_builder,renderer,test_renderer,pipeline}.py` plus the runtime in `overture-schema-pyspark/src/overture/schema/pyspark/`. Everything under `generated/` is regenerable output -- review the codegen, but skim the output to understand the shape of what's produced.

This seems reasonable for as a basis for review, but for merging, I don't see the upside in committing 15-30 generated files and 15K-30K generated lines to source control—we want to source control the generator, not the generatee, right?

vcschapp:
Divergence from Python no. 1 comment.

> - `UniqueItemsConstraint` uses Spark's `array_distinct`, which compares whole elements with structural equality on raw stored values. Pydantic compares normalized Python objects -- e.g., `list[HttpUrl]` is compared after URL normalization (such as ensuring that URLs like `http://example.com` contain a trailing `/`: `http://example.com/`). The PySpark check catches exact duplicates, not duplicates after normalization.

Basically we're saying PySpark will slightly under-count duplicates in some rare edge cases. While annoying, this doesn't seem like it's a very big deal.

vcschapp:
Divergence from Pydantic no. 3 question:

> - `BBox`'s PySpark expressions include checking latitude ordering and range. Longitude is skipped due to undefined behavior around how we handle features that cross the antimeridian.

Does this mean we have *stricter* constraints in the PySpark than in Pydantic? If yes, shouldn't they be the same?

vcschapp:
Regarding `@no_extra_fields`:

> `NoExtraFieldsConstraint` is intentionally skipped, as its behavior is implicit when using DataFrames.

This is true in terms of struct-typed columns, but isn't it *not* true for a top-level dataframe? *i.e.* Doesn't it mean that if I have a model like:

```python
@no_extra_fields
class Foo(BaseModel):
    bar: int
```

And a dataframe like:

| bar | baz      | qux  |
|-----|----------|------|
| 42  | "garply" | 3.14 |

Wouldn't the `baz` and `qux` columns sneak through without causing a violation?

vcschapp:
Divergence from Pydantic no. 2 comment:

> - `require_any_of` checks `isNotNull` as a proxy for Pydantic's `model_fields_set`. Parquet has no equivalent of "explicitly provided" because of the tabular nature of DataFrames; `isNotNull` is stricter (it rejects fields explicitly set to null).

Let's see if I understand this right.

I have a model like this:

```python
@require_any_of("bar", "baz")
class Foo(BaseModel):
    bar: int | None = None
    baz: str | None = None
```

I think you're saying that Pydantic would allow `Foo(bar=None)` because `bar` is set explicitly but PySpark would reject it...

The latest version of the Pydantic should also reject it... The docs say:

> To ensure parity between Python and JSON Schema validation, a field's value must be **explicitly set to a non-`None` value** to satisfy the constraint. Fields whose value was set by Pydantic using a default value violate the constraint, **as do fields explicitly set to `None`**.

That seems to be the current behavior:

```python
>>> from overture.schema.system.model_constraint import require_any_of
>>> @require_any_of("bar", "baz")
... class Foo(BaseModel):
...     bar: int | None = None
...     baz: str | None = None
... 
>>> Foo(bar=None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ANT.AMAZON.COM/schapper/overture/schema/.venv/lib/python3.10/site-packages/pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Foo
  Value error, at least one of these fields must be set to a value other than None, but none are: bar, baz (`@require_any_of`) [type=value_error, input_value={'bar': None}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

Am I missing something, or did you actually achieve perfect parity?

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/length_constraints.py
```diff
@@ -0,0 +1,47 @@
+"""Internal typed length-constraint classes.
+
+`annotated_types.MaxLen` and `annotated_types.MinLen` are polysemous:
```
vcschapp, line 3: I'm something of a logophile myself (if not exactly a sesquipedalian) and I've never heard of polysemous!

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/length_constraints.py
```diff
@@ -0,0 +1,47 @@
...
+`MaxLen(10)` on a `str` constrains character count, while `MaxLen(10)`
+on a `list[X]` constrains cardinality. The codegen extractor splits
+them by attachment layer so each variant carries its own dispatch:
+`ArrayMinLen` / `ArrayMaxLen` for `ArrayOf` layers, `ScalarMinLen` /
+`ScalarMaxLen` for scalar layers.
+
+These are codegen-internal classes -- Pydantic users continue to write
+`Annotated[X, MinLen(n)]` in their schemas; the wrapping happens inside
```
vcschapp: For my understanding: Does `Field(max_length=n)` also produce this type of internal constraint, or is that a different code path?

packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/templates/feature_module.py.jinja2
```diff
@@ -0,0 +1,83 @@
+{% from '_check_function.py.jinja2' import check_function -%}
+# This file is auto-generated by overture-schema-codegen. Do not edit.
+
+"""{{ feature_title }} validation expression builders."""
```
vcschapp: Naming question regarding "feature": can this handle any `BaseModel` that gets pulled in via discovery, or does it have to be a "feature"—and if so, what flavor of feature?

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
```diff
@@ -219,6 +241,13 @@ def docs_url(self) -> str:
         )
 
 
+FeatureSpec: TypeAlias = ModelSpec | UnionSpec
+"""Top-level feature types passed through the extraction pipeline.
+
+Consumers narrow with `isinstance` when an arm-specific attribute
+is needed (e.g. `UnionSpec.discriminator_field`).
+"""
```
vcschapp: I didn't realize this before, but this seems to be overloading the term "feature" to mean basically, "a thing that is either derived from `BaseModel`, or a union of such things"; whereas in `system` and `common` we use "feature" to mean a *geospatial* feature, *i.e.*, a thing with geometry and related attributes.

Can we use some more neutral naming for this type and the things flowing from it?

- `RootSpec`
- `TopLevelSpec`
- `TargetSpec`

vcschapp (commented):
Nice work on this.

It's a huge PR, but I think the successfulness of it is best captured by this comment:

> Every constraint Pydantic enforces today is dispatched to a PySpark expression, which allows Spark to push down predicates and take advantage of Catalyst to avoid unnecessary deserialization ...

That's phenomenal.

A few points to discuss, with my biggest items being committing of generated code; and some minor conceptual issues around the use of the word "feature".

sethfitz:
> This seems reasonable for as a basis for review, but for merging, I don't see the upside in committing 15-30 generated files and 15K-30K generated lines to source control—we want to source control the generator, not the generatee, right?

We agreed in one of the Schema WG meetings to drop the generated code from the PR to merge (once we've reviewed it), as it can be regenerated on demand and will exist in published packages for posterity.

> Basically we're saying PySpark will slightly under-count duplicates in some rare edge cases. While annoying, this doesn't seem like it's a very big deal.

Exactly. Agreed.

> Does this mean we have stricter constraints in the PySpark than in Pydantic? If yes, shouldn't they be the same?

For `BBox`, yes. We're going to follow up with @Rachmanin0xFF and the Data Platform folks to understand what their validation requirements + desires are today (in PySpark) and eventually make sure that PySpark and Pydantic are equivalent.

> Wouldn't the baz and qux columns sneak through without causing a violation?

No, these would be rejected when checking the Spark schema (before moving into per-row validation).

>> `require_any_of` checks `isNotNull` as a proxy for Pydantic's `model_fields_set`. Parquet has no equivalent of "explicitly provided" because of the tabular nature of DataFrames; `isNotNull` is stricter (it rejects fields explicitly set to null).
> Am I missing something, or did you actually achieve perfect parity?

I did achieve parity with Pydantic, but wanted to call this out because of the *ability* to omit values there (and when using JSON).

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/length_constraints.py
sethfitz: New to me too, but it's precise, and I like it!

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/length_constraints.py
sethfitz: Good catch; that should say `Field(max_lenth=n)`.

packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/templates/feature_module.py.jinja2
sethfitz: It can handle any `BaseModel`.

packages/overture-schema-codegen/src/overture/schema/codegen/extraction/specs.py
sethfitz: True. That was probably in the back of my head because I was thinking about validating Overture Features as its primary purpose even while the implementation was more generic.

None of those names really resonate, but I'll see if I can come up with something that feels less overloaded. Something that properly encompasses `ModelSpec | UnionSpec` (where `UnionSpec` is implicitly "union of models"). If we weren't dealing with unions, `ModelSpec` could be appropriate. `PrimarySpec` (complementing `SupplementarySpec`, which encompasses enums, NewTypes, nested models, other types) feels close.

sethfitz: I went with renaming `FeatureSpec` to `ModelSpec` (in line with Pydantic conventions) and `ModelSpec` turned into `RecordSpec` to distinguish it from `UnionSpec`.

vcschapp: Thx, that seems like a good choice.

vcschapp:
> We agreed in one of the Schema WG meetings to drop the generated code from the PR to merge (once we've reviewed it), as it can be regenerated on demand and will exist in published packages for posterity.

Given how close we are to merging, it seems reasonable to update the PR to drop the generated code now.

vcschapp:
> > Does this mean we have stricter constraints in the PySpark than in Pydantic? If yes, shouldn't they be the same?
> 
> For `BBox`, yes. We're going to follow up with [Adam Lastowka (@Rachmanin0xFF)](https://github.com/Rachmanin0xFF) and the Data Platform folks to understand what their validation requirements + desires are today (in PySpark) and eventually make sure that PySpark and Pydantic are equivalent.

IMO we should update the PR to provide exact parity to Pydantic. There's nothing useful served by having diverging `BBox` validations.

The `BBox` validation issue, is conceptually separate from this PR. In my mind, this PR is about building the system that can do PySpark validation replicating the existing Pydantic validation as closely as is technically feasible. The `BBox` validation issue is a different topic: do we need more extensive `BBox` validation across all supported validation platforms? Let's have that discussion separately and resolve it separately in its on issue and PR.

packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/templates/feature_module.py.jinja2
vcschapp: As of PR viewing on 7/16/2026, `feature_module.py.jinja2` is now renamed to `model_module.py.jinja2` and the term "feature" is replaced with "module". Resolving.

packages/overture-schema-system/src/overture/schema/system/field_path.py
```diff
@@ -0,0 +1,468 @@
...
+
+    name: str
+    projection: MapProjection
+
+
+PathSegment: TypeAlias = StructSegment | ArraySegment
+
+
+@dataclass(frozen=True, slots=True)
+class ScalarPath:
```
vcschapp, line 95: Given this can address either a scalar value or a whole struct (I think, correct if wrong) I wonder if a better name here might be `SinglePath` or `DirectPath`.

I don't feel strongly about this, just an idea.

packages/overture-schema-system/src/overture/schema/system/field_path.py
```diff
@@ -0,0 +1,468 @@
...
+    `projection` selects keys or values; the projected side is iterated
+    like an array, so checks on a `MapSegment` render through the same
+    element machinery as `ArraySegment`.
+    """
+
+    name: str
+    projection: MapProjection
+
+
+PathSegment: TypeAlias = StructSegment | ArraySegment
```
vcschapp, line 91: I find this type name a bit challenging because the exclusion of `MapSegment` feels arbitrary, hard to explain, in something with such a general name as `PathSegment`.

It's not heavily used, suggest dropping the explicit type alias and just inserting `StructSegment | ArraySegment` everywhere.

I don't feel strongly about this, just an idea.

packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
```diff
@@ -0,0 +1,51 @@
...
+
+@dataclass(frozen=True)
+class Check:
+    """One validation check.
+
+    `field` identifies what the check is about (for error column naming
+    and report grouping), not how to access the data.  The expression in
+    `expr` already encodes the access pattern.
+
+    `read_columns` names every top-level schema column the expression
```
vcschapp, line 30: Would be helpful to expand on the relationship between the singular `expr: Column` and the plural `read_columns: frozenset[str]`.

packages/overture-schema-system/src/overture/schema/system/field_path.py
```diff
@@ -0,0 +1,468 @@
...
+  segments encode nested-list iteration without an intervening struct,
+  e.g. `list[list[X]]` parses as a single `ArraySegment` with
+  `iter_count=2`).
+- `MapPath` -- struct segments leading to a map column, a single
+  `MapSegment` projecting the map to its keys or values, then a struct-only
+  leaf (possibly empty). Locates a value reached by iterating a
+  `dict[K, V]`'s keys or values, encoded with a `{key}` / `{value}` marker
+  on the map column and the leaf appended after it (e.g. `names.common{key}`
+  for a scalar value, `subs{value}.label` for a field inside a
+  `dict[K, Model]` value).
```
vcschapp, line 20: The limitations on where maps can appear seem arbitrary.

It feels like the tail (limitation of the stringized path representation grammar) might be wagging the dog (of allowed representations).

In both Pydantic land and Spark land, map values can be scalars, arrays, structs, or maps; and that structure can continue deep into the schema. So this doesn't feel like a limitation based on trying to fit to the lowest common denominator, it feels more like a limitation that's either arbitrary, or caused by a limitation in the path grammar, or (unnecessarily) reflecting some other known limitation elsewhere in the program.

I have a feeling that with refactoring, this module could end up being both more generalized (supporting the structured children of map use cases) with potentially fewer lines of code...

I don't at all consider this a blocker to merging. It's a useful discussion to have but any fixing could easily be documented as a GH issue and done as a follow-up.

vcschapp:
I found a conceptual bug of omission that's minor and I don't think it should block, but I do think we should note it in a GH issue for later fixing. It's more of a general code generation issue rather than anything specific to PySpark.

# Issue

We "support" generating code for certain collection types that are supported by Pydantic, however the generated code is wrong both for Markdown and PySpark.

The important ones are:

1. `set`, *e.g.*, `set[str]`
2. variable-length tuples, which are list-equivalent, *e.g.*, `tuple[str, ...]`
3. `Sequence`, *e.g.*, `Sequence[str]` since Pydantic is happy with that

The bonus one would be fixed-length tuples where the underlying type Spark type of all the elements is the same base type, as in:

`tuple[str, str]` or `tuple[str, SnakeCaseString]`

# What happens now?

Here's how the system behaves today with the various kinds of inputs:

  | Annotation           | FieldShape                  | Spark schema type            | Markdown       | Correction                   |
  | -------------------- | --------------------------- | ---------------------------- | -------------- | ---------------------------- |
  | `list[str]`          | `ArrayOf(Primitive('str'))` | ✅ `ArrayType(StringType())` | ✅ `list<str>` | N/A                          |
  | `set[str]`           | `Primitive('set')`          | ❌ `StringType()`            | ❌ `set`       | Spark array, MD `set<str>`?  |
  | `tuple[str, ...]`    | `Primitive('tuple')`        | ❌ `StringType()`            | ❌ `tuple`     | Spary array, MD `list<str>`? |
  | `tuple[str,str,str]` | `Primitive('tuple')`        | ❌ `StringType()`            | ❌ `tuple`     | Spark array, MD ???          |
  | `tuple[str, int]`    | `Primitive('tuple')`        | ❌ `StringType()`            | ❌ `tuple`     | fail codegen                 |
  | `Sequence[str]`      | raises `TypeError`          | ❌ aborts run                | ❌ aborts run  | Spary array, MD `list<str>`? |

# Why isn't this important to fix now?

1. No Overture types use these features today.
2. Rome wasn't built in a day, we don't need a PR to be perfect in every possible way to be mergeable.

vcschapp (commented):
I've been over the PR as best as I can given the significant scope.

Same conclusion as before, I think it's ready to merge and any minor tweaks or good ideas in the comments can be done after the fact. 

Two exceptions, mentioned early in the comment stream 👆:

1. We should drop the generated code before merging.
2. We should reduce `BBox` validation to parity with Pydantic for now and offload advance validation to a separate issue.

sethfitz:
Closed in favor of #569

packages/overture-schema-system/src/overture/schema/system/field_path.py
sethfitz: Agreed — the name claimed a generality it didn't have, sitting right next to `FieldSegment` (the actual `Struct | Array | Map` one). Only two use sites, so I've inlined `StructSegment | ArraySegment` and dropped the alias. #571.

_Drafted by Claude with Seth's oversight._

packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
sethfitz: Expanded the docstring to spell out the relationship: `expr` and `read_columns` are two views of one computation — `expr` is the single composed `Column`, `read_columns` is its read-set (every top-level column that one expression dereferences). They're carried separately because the builder knows the columns as it composes `expr`, and recording them beats recovering them from the finished `Column`. #571.

_Drafted by Claude with Seth's oversight._

packages/overture-schema-system/src/overture/schema/system/field_path.py
sethfitz: There's a real imprecision here, but I think it's tangled with the map-nesting thread, so I've pulled both into #570 rather than rename now. The sharper version of the point: `ArrayPath` / `MapPath` name *how the path fans out*, while `ScalarPath` names *what's at the end* — a mixed axis. But we can't rename the members cleanly before settling the taxonomy: if a path can ever traverse both an array and a map (the maps thread below), `ArrayPath` vs `MapPath` stops being a partition and the split collapses toward `Direct` (no iteration) vs `Iterated`. So the rename rides on that call.

_Drafted by Claude with Seth's oversight._

packages/overture-schema-system/src/overture/schema/system/field_path.py
sethfitz: Filed as #570. Checked the premise while I was in there: the only map in the schema today is `CommonNames` (`dict[LanguageTag, StrippedString]`), so nothing currently reaches the guards — they're latent gap-markers, not live bugs, which is what makes deferring safe. The asymmetry is the interesting part: `MapPath` already carries a `dict[K, Model]` leaf case nothing in the schema uses, while hard-blocking `dict[K, list]` and nested maps the other way. So the real fork is generalize-fully (one iterated path type, segments mixing array + map) vs narrow-to-`dict[K, scalar]`. #570 carries that plus the `ScalarPath` naming, since they turn out to be the same decision.

_Drafted by Claude with Seth's oversight._

PR #532, opened 2026-05-19
[FEATURE] Added provider, resource, and version as required fields.

### LINKED ISSUE #530: Update sources details to provide the version
# Background

The current `sources` field of the schema contains an array of `sourcePropertyItems`. Each item contains the following elements

```
- property: which property of the feature the source is associated with
- dataset: the dataset identifier, the name of the source
- license: the license that the data falls under
- record_id: the feature source identifier (how you identify the input feature in the dataset)
- update_time: inconsistently applied, but the time that the source data was last updated
- confidence: even more inconsistently applied, but how confident we are in the quality of the data or its existence
```

The goal here is to provide provenance for our data allowing users to track back and see what Overture data was constructed from. In other words the user can look at the specific record_id within the identified dataset and see where the data really comes from. The problem with this is that our dataset identifier is not unique; we are missing version information. As such I would suggest that we update our `sourcePropertyItem` definition to include the following:

```
- property: unchanged
- dataset: unchanged, but will be put on a deprecation path
- license: unchanged
- record_id: unchanged
- update_time: unchanged
- confidence: unchanged
- provider: The name of the entity that produced the data: meta, esri, microsoft, osm, etc.
- resource: The subject or type of data given by the provider: division-names, buildings, planet, etc.
- version: The sortable identifier such as a date or number: 2026-02-13, 5.3, A5692
```

Basically the `dataset` is being broken up into the `provider` + `resource`. This is being done because a number of our providers give us multiple resources and we want to make it easy for users to filter either a single resource or all. When combined, the `provider` + `resource` + `version` uniquely identify a specific input data snapshot and the record_id can be used to find the specific entity.

ericgodwin:
# Major change release plan

This step alone is not a major change as it is simply adding new fields, but when we remove the deprecated dataset field that will be breaking.

The breaking change for this should come 6 to 12 months after the minor change is released. For this change to be rolled out we are going to have to update all data loaders to populate these required fields. That will take some time. While this pull request is out now, we may not actually be able to implement this change until the August timeframe.

- Release this non-breaking change and announce a deprecation timeline for the `dataset` field.
- Come back later and remove the outdated `dataset` field and make the three new fields non-optional.

Messaging around this change is that the current method of providing provenance is not sufficient to ensure traceability. Besides documenting the deprecation of `dataset` we will want to provide details on how the `provider`, `resource`, and `version` work together to identify a data snapshot.

# Description

The intent of this change is to update our source item field to include the information necessary for data provenance:

```
- provider: The name of the entity that produced the data: meta, esri, microsoft, osm, etc.
- resource: The subject or type of data given by the provider: division-names, buildings, planet, etc.
- version: The sortable identifier such as a date or number: 2026-02-13, 5.3, A5692
```
Together, along with the `version_id` these values allow a user to uniquely identify what raw input data was used to construct Overture data. Our current system, of providing only a `dataset` is lacking dataset version information but is also inconsistently constructed. All three new fields will be omitable which means optional (absent or non-null string) and null rejected at validation time.

# Reference
Closes #530 

# Testing

The unit tests for the schema were run.

 ## Test Results

Added 2 new tests:
- test_source_item_provider_resource_version_omittable (omission works, fields excluded from JSON)
- test_source_item_provider_resource_version_not_nullable (null raises ValidationError)

```
 All 1995 tests passed.

 uv run pytest -W error packages/ -q --tb=short
 1995 passed in 5.18s

 Lint and formatting checks also clean (`ruff check`, `ruff format --check`).
```

# Checklist

1. [ ] Add relevant examples.

# Documentation website
[Docs preview for this PR.](https://staging.overturemaps.org/schema/pr/532/schema/index.html)

### DIFF
--- a/counterexamples/base/bathymetry/bad-depth.yaml
+++ b/counterexamples/base/bathymetry/bad-depth.yaml
@@ -13,5 +13,8 @@ properties:
     - record_id: x123
       property: ""
       dataset: some source
+      provider: some_provider
+      resource: some_resource
+      version: "2087-08-23"
   ext_expected_errors:
     - "depth/minimum]: minimum: got -1, want 0"
--- a/counterexamples/buildings/bad-confidence-in-source.yaml
+++ b/counterexamples/buildings/bad-confidence-in-source.yaml
@@ -14,8 +14,14 @@ properties:
   height: 21.34
   sources:
   - property: ""
-    dataset: microsoftMLBuildings,
+    dataset: microsoftMLBuildings
+    provider: microsoft
+    resource: ml-buildings
+    version: abc
     confidence: 100
   - property: /properties/height
-    dataset:  metaLidarExtractions,
+    dataset:  metaLidarExtractions
+    provider: meta
+    resource: lidar-extractions
+    version: "2087-08-23"
     confidence: -1
--- a/counterexamples/buildings/bad-license.yaml
+++ b/counterexamples/buildings/bad-license.yaml
@@ -11,6 +11,9 @@ properties:
   version: 1
   sources:
     - dataset: MyGreatDataset
+      provider: overture
+      resource: soccer-stadiums
+      version: "2087-08-23"
       property: "/geometry"
       update_time: '2024-04-26T23:55:01-08:00'
       confidence: 0.3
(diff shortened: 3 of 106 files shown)


ericgodwin:
I have changed enough functionality at this point that it makes sense to just start over.

Issue #533, opened 2026-05-20
[Devops] Branching Strategy - Phase 2.B - Release Versions


lowlydba:
## Goal

Implement CI release-event detection and the release process for `main`.

Scope: release trigger + `vnext`/`main` merge process + changelog machinery + docs. No publish workflows, those live in Phase 3 (#509), which this phase unblocks.

Implemented in PR #557.

## Model (as implemented)

Humans own the full `<major>.<minor>.<patch>` in each package's `pyproject.toml`. Any version increase merged to `main` is a release; CI only versions the interim internal builds between releases.

1. Every PR that changes a package carries a [towncrier](https://towncrier.readthedocs.io) fragment (`packages/<pkg>/changelog.d/`), enforced by CI.
2. A release PR bumps the version and runs `towncrier build`, folding fragments into that package's `CHANGELOG.md` (notes reviewed in the PR).
3. On merge, `release-trigger` cuts one published GitHub Release per bumped package: tag `<package>-v<version>`, notes read back from `CHANGELOG.md`. The umbrella `overture-schema` release is marked Latest.
4. No-bump merges get internal builds versioned `<version>.postN+main.<sha>` (or `+vnext.<sha>`), which order after the release for `>=` consumers and are rejected by public PyPI by construction.

## Decisions resolved

| Question | Decision |
|---|---|
| Patch destination | Patch is a human bump like any other, releases to PyPI. No-bump merges are internal-only `.postN` builds. |
| Changelog approach | towncrier per-PR fragments, folded in the release PR, enforced by a CI check. |
| Release granularity | Per-package tags (`<pkg>-v<semver>`); packages version independently. |
| Release gate | None at the GH Release step; the human gate moves to a GitHub Environment approval on the Phase 3 PyPI publish job. |
| Internal build versioning | PEP 440 post-releases; `+dev.N` rejected as not orderable. `N` is a per-version sequence (`.post0` = identical to the release). Verified: `>=X.Y.Z` resolves `.postN`, `==X.Y.Z` selects the clean release, `>` never matches (pin with `>=`). |
| Workspace dependency floors | uv dual declaration: hand-maintained `>=` floors in `project.dependencies` alongside workspace sources. A major bump of a dependency must cascade to its dependents, enforced by CI. |

## Definition of done

- [x] `release-trigger` workflow live: detects per-package bumps, fails on version decreases, creates per-package Releases
[...]

Issue #509, opened 2026-05-05
[Devops] Branching Strategy - Phase 3


lowlydba:
## Goal

Build and publish packages to the right destinations using version strings computed by Phase 2. Each workflow has one trigger and one destination. No version logic lives here.

`vnext` dev builds to CodeArtifact are out of scope here, split to Phase 4 (#510): they're blocked on a CodeArtifact repo that doesn't exist yet ([ops-team#299](https://github.com/OvertureMaps/ops-team/issues/299)), so this phase can ship and close without waiting on that migration.

## Trigger x destination matrix

| Trigger | Workflow | Packages | Version | Destination |
|---|---|---|---|---|
| Push to `main`, no bump | [`main-publish.yaml`](../.github/workflows/main-publish.yaml) | affected only | `<version>.postN+main.<sha>` | CodeArtifact |
| GitHub Release published (manual bump) | [`release-publish.yaml`](../.github/workflows/release-publish.yaml) | released package | `<version>` | Public PyPI (upstreamed by all internal CA) |
| Push to `vnext` | see Phase 4 (#510) | affected only | `<version>.postN+vnext.<sha>` | ⛔ Blocked on [ops-team#299](https://github.com/OvertureMaps/ops-team/issues/299) |

CodeArtifact: domain `overture-pypi` / account `505071440022` / region `us-west-2`.

### Internal builds -> CodeArtifact
- Trigger: push to `main` with no version bumps.
- Detect affected packages, version via the `compute-version` action (`main` context), build + publish to CodeArtifact. Workspace dependency floors are declared statically in each `pyproject.toml` (dual declaration), so no metadata rewriting is needed at build time.
- Implemented in PR #638.

### Releases -> PyPI
- Trigger: GitHub Release published. `release-trigger.yaml` creates the release with the `overture-release-publisher` app's installation token (provisioned via #637) instead of `GITHUB_TOKEN`, so the release actually fires the `release: published` event this workflow needs.
- Build the released package at its released version and publish to public PyPI via [Trusted Publishing](https://docs.pypi.org/trusted-publishers/) (OIDC) + attestations.
- Human gate: GitHub Environment (`pypi-release`) approval (required reviewers) on this job.
- Pattern follows the [overturemaps-py publish workflow](https://github.com/OvertureMaps/overturemaps-py/blob/main/.github/workflows/publish-pypi.yml).
- Prereq: PyPI Trusted Publisher OIDC + pypi-release environment configured (manual, one-time; #653).
[...]

Issue #490, opened 2026-04-02
[Devops] Streamlined branching strategy for schema


lowlydba:
♻️ This top comment is evergreen, and will evolve to match the current state of this effort.

----

## Overview

Streamlined branching strategy and release automation for the schema repository. Replaces the old `dev` -> `staging` -> `main` manual flow and per-release long-lived branches (e.g. `main_2026_02_12`) with a two-branch model, automated CI safety nets, and PR-driven releases.

Live documentation: [CONTRIBUTING.md](https://github.com/OvertureMaps/schema/blob/main/CONTRIBUTING.md) (branch mechanics) and [docs/versioning.md](https://github.com/OvertureMaps/schema/blob/main/docs/versioning.md) (version scheme + release process).

## Two-branch model

- `main`: integration branch. All normal work lands here. Always releasable.
- `vnext`: accumulates major changes between releases. Continuously rebased onto `main` by CI. Merges to `main` at release time (regular merge, not squash).

## Versioning model (settled in Phase 2.B)

- Humans own the full `<major>.<minor>.<patch>` per package in `pyproject.toml`. Any bump merged to `main` cuts a per-package GitHub Release (tag `<pkg>-v<version>`) and, in Phase 3, a public PyPI publish.
- Every PR that changes a package carries a towncrier changelog fragment, enforced by CI; release PRs fold fragments into `CHANGELOG.md`.
- No-bump merges produce internal-only builds versioned `<version>.postN+main.<sha>` (or `+vnext.<sha>`) for CodeArtifact. `>=` consumers resolve them; public PyPI rejects them by construction.

## Phases

| Phase | Title | Status |
|---|---|---|
| [Phase 0](https://github.com/OvertureMaps/schema/issues/506) | Branch infrastructure | ✅ Done |
| [Phase 1](https://github.com/OvertureMaps/schema/issues/507) | CI safety net | ✅ Done |
| [Phase 2.A](https://github.com/OvertureMaps/schema/issues/508) | Internal versions & version compute | ✅ Done (#534) |
| [Phase 2.B](https://github.com/OvertureMaps/schema/issues/533) | Release versions | ✅ Done (#557) |
| [Phase 3](https://github.com/OvertureMaps/schema/issues/509) | Package publish workflows (`main` + PyPI releases) | 🔄 PR open (#638) |
[...]

PR #534, opened 2026-05-21
[FEATURE] Branching strategy Phase 2.A - version computing & baselining (merge conflict resolution)

### LINKED ISSUE #508: [Devops] Branching Strategy - Phase 2.A - Internal versions
## Goal

Build the version scheme foundation and CI compute logic for **all branch contexts**.

Scope: version computation for `vnext`, `main`, and `main-bump`. No release triggering or publishing — those are Phase 2.B (#533) and Phase 3 (#509) respectively.

## Version scheme

Each of the 12 packages in `packages/*` carries `<major>.<minor>.<patch>` in its own `pyproject.toml`. Rules:

- **`<major>` and `<minor>`** are git-controlled. Bumped manually by a human-authored commit per the [classification table](https://lf-overturemaps.atlassian.net/wiki/spaces/SCHEM/pages/14286874/Schema+versioning+and+stability). Per-package independent.
- **`<patch>`** in `pyproject.toml` is seed/informational only. CI computes the actual published `<patch>` at build time (see `compute-version`).
- **`packages/overture-schema`** is the canonical schema package.

### `p2-version-scheme`

- Audit and establish baseline `<major>.<minor>` versions for all 12 packages.
- Document version bump rules per the classification table.

### `compute-version`

Reusable CI composite action: given (package, branch context), returns version string to publish.

| Context | Output version |
|---------|----------------|
| `vnext` | `<last-published>+dev.<run_number>` (PEP 440 local) — highest version for that pkg in CA. Falls back to `<major>.<minor>.0+dev.<run_number>` if never published. |
| `main` | `<major>.<minor>.<next-patch>` — highest existing `<patch>` for that `<major>.<minor>` in CA + 1. Falls back to `<major>.<minor>.0`. |
| `main-bump` | `<major>.<minor>.0` (patch resets to 0) |

## Definition of done

- [x] Baseline `<major>.<minor>` versions established for all 12 packages
- [x] `compute-version` composite action implemented + tested for all contexts (`vnext`, `main`, `main-bump`)
- [x] PR: #534

lowlydba:
# Major change release plan

Not applicable — this PR does not introduce a MAJOR schema change.

# Description

Part of the [branching strategy rollout](https://github.com/OvertureMaps/schema/issues/490) — this is **Phase 2.A** ([#508](https://github.com/OvertureMaps/schema/issues/508)): establishes package version baselines and the CI plumbing that later phases will publish with.

**Nothing is published anywhere by this PR** — it only computes and logs what versions *would* be used.

What this adds:
- Every package pinned to a static baseline version in `pyproject.toml` (`overture-schema` at `1.17.1`, others at `0.1.1`), replacing the old `__about__.py`-driven dynamic versioning.
- `compute-version`: a composite action that computes the PEP 440 version a package would get for a given branch context (`vnext`, `main`, or `main-bump`).
- `code-artifact`: a composite action for CodeArtifact auth, replacing the old shell script.
- `compute-versions-dry-run`: a CI workflow that logs computed versions on every push to `main`/`vnext` (and on relevant PRs) — visibility only, nothing is built or published (yet)

What's still to come (out of scope here):
- **Phase 2.B** ([#533](https://github.com/OvertureMaps/schema/issues/533)): decide + implement when a `main` merge actually cuts a release.
- **Phase 3** ([#509](https://github.com/OvertureMaps/schema/issues/509)): the real publish workflows (CodeArtifact dev builds, patch builds, public PyPI releases) that will call `compute-version`

See the updated roadmap table in [`CONTRIBUTING.md`](https://github.com/OvertureMaps/schema/blob/508-devops-branching-strategy---phase-2a---internal-versions/CONTRIBUTING.md#migration-notes) for the full phase-by-phase status.

Additional update in this PR iteration:
- Resolved merge conflicts by merging latest `main` into this branch (picked up upstream cleanup/removal of obsolete `gers/examples/python` content).
- Fixed a bug where `compute-version` hard-failed instead of falling back when a package had never been published.
- Fixed a bug where `compute-version` regressed a package's baseline patch (from `pyproject.toml`) back to `.0` instead of using it as a floor.

# Testing

- Ran repository validation with `make check`. Result: `1993 passed`.
- Verified `compute-versions-dry-run` end-to-end in CI (successful run) and locally with `act` for both `workflow_dispatch` and `pull_request` events.
- Verified the `compute-version` version-resolution fixes with synthetic test cases (both `main` and `vnext` contexts).
- Ran final automated validation (Code Review + CodeQL); CodeQL reported no alerts.

# Checklist

1. [ ] Add relevant examples.


gers/examples/python/utils.py
```diff
@@ -1,212 +1,212 @@
...
+
+            if not "geometry" in feat_dict:
+                if "lat" in feat_dict["properties"] and "lon" in feat_dict["properties"]:
+                    feat_dict["geometry"] = f'POINT({feat_dict["properties"]["lon"]} {feat_dict["properties"]["lat"]})'
+                else:
+                    continue
+            
+            features.append(feat_dict)
+    return features
+
```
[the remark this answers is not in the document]
lowlydba: Same — pre-existing in `gers/examples/python/`, only touched by CRLF normalization. Out of scope for this PR.

packages/overture-schema/pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "overture-schema-cli",
 ]
 description = "Complete Overture Maps schema collection with all themes and types"
-dynamic = ["version"]
+version = "1.17.1"
 license = "MIT"
```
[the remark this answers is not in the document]
lowlydba: Fixed — typo in the PR description. Version is correctly set to `1.17.1` in `pyproject.toml`; the description has been updated to match.

gers/examples/python/match_classes.py
```diff
@@ -1,216 +1,216 @@
...
+        return j
+
+class PointSnapInfo:
+    """Snap-to-road match information corresponding to one point in a trace"""
+    def __init__(self, index: int, original_point: Point, time: str, seconds_since_prev_point: float=None, predictions:Iterable[SnappedPointPrediction]=[]) -> None:
+        self.index = index
+        self.original_point = original_point
+        self.time = time
+        self.seconds_since_prev_point = seconds_since_prev_point
+        self.predictions = predictions
```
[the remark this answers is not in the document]
lowlydba: Pre-existing bug in `gers/examples/python/` — this file was only touched by the line-ending normalization commit, not authored in this PR. Logged for the `gers` maintainers to address separately.

.github/actions/compute-version/action.yml
```diff
@@ -0,0 +1,107 @@
...
+        echo "Seed version for ${PACKAGE}: ${SEED} (major.minor: ${MAJOR_MINOR})"
+
+        # --- Query CodeArtifact for the latest published version ---
+        # uv pip compile resolves the latest matching version from the index.
+        # We constrain to the current major.minor series for `main` context.
+        resolve_latest() {
+          local constraint="$1"
+          echo "$constraint" \
+            | uv pip compile - --index-url "$INDEX_URL" --no-deps --quiet 2>/dev/null \
+            | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo ""
```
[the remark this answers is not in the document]
lowlydba: Fixed. `resolve_latest()` now captures `uv pip compile` stdout+stderr, checks the exit code explicitly, and hard-fails with the error output if `uv` itself errors (network/auth failure etc.). A clean empty result (package not yet published) still returns empty string as intended.

.github/workflows/compute-versions-dry-run.yaml
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            case "$CONTEXT" in
+              vnext)
+                LATEST=$(resolve_latest "$pkg")
+                BASE="${LATEST:-${MAJOR_MINOR}.0}"
+                VERSION="${BASE}+dev.${GITHUB_RUN_NUMBER}"
+                ;;
+              main)
+                LATEST_IN_SERIES=$(resolve_latest "${pkg}>=${MAJOR_MINOR}.0,<${MAJOR_MINOR}.99999")
```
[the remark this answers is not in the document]
lowlydba: Fixed with the same approach as the action — `uv pip compile` errors now surface and fail the step rather than silently returning an empty result.

gers/examples/python/match_classes.py
```diff
@@ -1,216 +1,216 @@
...
+        self.revisited_segments = revisited_segments
+
+    def to_json(self, diagnostic_mode=False, include_all_predictions=False):
+        points_json = list(map(lambda x: x.to_json(diagnostic_mode, include_all_predictions), self.points))
+        return {
+            "id": str(self.id),
+            "elapsed": self.elapsed,
+            "source_length": self.source_length,
+            "route_length": self.route_length,
+            "points": len(self.points),
```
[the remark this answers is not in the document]
lowlydba: Same as above — pre-existing in `gers/examples/python/`, only touched by CRLF normalization. Out of scope for this PR.

ericgodwin (dismissed):
A little bit overwhelming of a review, but the concept seems sound.

vcschapp:
> 19 legacy CRLF files fixed

Good catch, appeals to the pedant in me! I wonder how they escaped my all-seeing eye!?

vcschapp (commented):
I think we (or at least I) need to think this through a bit more:

> ### Why remove hatch dynamic versions?
> 
> Previously, each package used `dynamic = ["version"]` with `[tool.hatch.version]` pointing at an `__about__.py` file containing a hardcoded string. This meant version bumps required editing source files and committing them — tightly coupling human commits to version identity. The new model separates concerns: `<major>.<minor>` lives in `pyproject.toml` (bumped intentionally by humans), while `<patch>` is computed at build time by CI based on what's already published. `__about__.py` now reads from installed package metadata at runtime, so it always reflects whatever version was actually stamped during build — no manual sync needed.

The choice of automated version bumps for public PyPI was intentional, the idea being that there should be human intention with an explicit version bump commit to trigger (public) publishing, and that the reason for the publish would be included in the commit message.

Let me give a concrete example of a workflow that I think would have unexpected results in a fully automated model.

Steps:

1. On May 29, Alice commits a minor change to `main`, with the intention that it will go out in a schema update just before the June release.
2. On May 30, Bob commits a second minor change to `main` with the same intention.

If we do automated patch updates, one of two things will happen.

- If they omit the major.minor change from their commits, then both Alice's and Bob's changes are going to be independently and immediately published as patch version bumps.
- If they include major.minor changes then both of their changes are going to be independently and immediately published as separate minor version bumps within a day of each other.

Whereas it seems more likely the intention was to combine those changes together into a single explicit minor version release...

.github/workflows/compute-versions-dry-run.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Get CodeArtifact credentials
+        id: ca
+        uses: ./.github/actions/code-artifact
+
+      - name: Compute version for each package
+        env:
+          CONTEXT: ${{ steps.context.outputs.value }}
+          INDEX_URL: ${{ steps.ca.outputs.index_url }}
+        run: |
+          echo "## Computed versions (context: ${CONTEXT})" >> "$GITHUB_STEP_SUMMARY"
```
vcschapp: Not blocking, but for larger scripts (maybe 10+ lines to make up an arbitrary number) I think there's value in factoring them out into a separate `.py` or `.sh` script that's independently testable, *i.e.*, you can run it from the command line with appropriate arguments and it does what's expected.

lowlydba: Totally - once the dust settles I'll likely convert these to GitHub Actions + scripts for similar reasons and benefits

lowlydba:
@copilot resolve the merge conflicts in this pull request

ericgodwin (dismissed):
After talking it through we are not going to publish the patch updates to the public repo automatically, but will internally. Looks good to go.

lowlydba:
> If we do automated patch updates, one of two things will happen.


@vcschapp As discussed, not going to do public patch releases (automatically) at this time - so that should clear up this concern. Please re-review

packages/overture-schema-addresses-theme/src/overture/schema/addresses/__about__.py
sethfitz, line 1: `__version__` was in code to be programmatically accessible, but I didn't realize that `importlib.metadata` had a good way to access it.

Given that, I'd be inclined to drop `__about__.py` from all of the packages. AFAIK, nothing references `__version__`.

CONTRIBUTING.md
```diff
@@ -134,7 +140,14 @@ If your fork still references `dev` or `staging`, update your remotes accordingl
 - vnext compatibility check added: every PR to `main` verifies that `vnext` can rebase cleanly on top; posts exact fix commands on conflict.
 - Post-merge automatic rebase added: `vnext` is force-rebased onto `main` after every merge; if it fails, a GitHub issue is opened.
 
-### [Phase 2](https://github.com/OvertureMaps/schema/issues/508)
+### [Phase 2.A](https://github.com/OvertureMaps/schema/issues/508), May 2026
+
+- All packages baselined with static versions in `pyproject.toml` (`overture-schema` at `1.17.1`, others at `0.1.1`).
```
sethfitz, line 156: I _think_ we'd agreed that the first public `overture-schema` PyPI release would included synced versions on the dependency packages (so we have a clean start, even if they start diverging immediately after). Not something that needs to be done now, just flagging that these `0.1.1` placeholders will need to be updated before then.

We'd also discussed using 2.0.0 as the first published version number, but that also doesn't need to be addressed yet.

(Or maybe this is part of the plan already and I missed it. If so, sorry!)

I don't have a strong position on either and can see benefits to both / it doesn't actually matter as long as we reference the `overture-schema` package version in the data's release notes.

uv.lock
sethfitz, line 1: This looks like it's out of sync. I would expect the version numbers to match above (0.17.1 / 0.1.1).

.github/actions/compute-version/action.yml
```diff
@@ -0,0 +1,114 @@
...
+          local constraint="$1"
+          local output
+          # uv pip compile exits non-zero on network/auth errors; let those surface.
+          # "Could not find a version" is a normal empty-result — grep returns 1, which is OK.
+          output=$(echo "$constraint" \
+            | uv pip compile - --index-url "$INDEX_URL" --no-deps --quiet 2>&1) || {
+            echo "ERROR: uv pip compile failed for '${constraint}':" >&2
+            echo "$output" >&2
+            exit 1
+          }
```
sethfitz: The comment here has `uv`'s behavior backwards, and I think it takes the fallbacks down with it. `uv pip compile` exits **non-zero** when nothing matches — it doesn't return zero with empty output:

```
$ echo "overture-schema-common>=99.0,<99.99999" | uv pip compile - --no-deps --quiet; echo $?
  × No solution found when resolving dependencies
1
```

So on a no-match the `|| { … exit 1; }` fires, and since the step runs under `set -euo pipefail` the whole thing aborts before either fallback is reached — vnext's `<major>.<minor>.0+dev.N` "if never published" (L82) and main's "patch 0" (L96) are both unreachable. The theme packages are seeded at `0.1.1` and (unless I'm missing it) aren't in CodeArtifact yet, so the first push that runs this would crash on the first unpublished package rather than fall back to the seed.

We'd want to separate "no matching version" (empty result, keep going) from a real network/auth failure (surface it). One shape:

```bash
output=$(echo "$constraint" | uv pip compile - --index-url "$INDEX_URL" --no-deps --quiet 2>&1) \
  || { echo "$output" | grep -qiE 'no solution found|not found in the package registry' && return 0
       echo "ERROR: uv pip compile failed for '${constraint}':" >&2; echo "$output" >&2; return 1; }
```

— though keying on uv's wording is its own kind of brittle across versions, so there may be a cleaner signal (a `uv pip index`-style existence probe first). Same function is duplicated in `compute-versions-dry-run.yaml`, so whatever we land here has to happen there too.

_Surfaced by a Claude-assisted review pass; the repro above is against uv 0.11.27._

.github/actions/compute-version/action.yml
```diff
@@ -0,0 +1,114 @@
...
+          echo "$output" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || true
```
sethfitz, line 76: Mostly a note: this only matches three-component versions, so a PEP 440 two-component release (`1.2`, which normalizes to `1.2.0`) reads as empty even when `uv` resolved it fine. We only ever cut `X.Y.Z`, so it shouldn't bite in practice — but it's the one path that *does* reach the fallback once the no-match handling above is sorted, and in `main` context an empty result there becomes `${MAJOR_MINOR}.0`, i.e. a collision with the already-published `X.Y`. `'[0-9]+(\.[0-9]+)+'` would cover both shapes.

_Surfaced by a Claude-assisted review pass._

.github/workflows/compute-versions-dry-run.yaml
```diff
@@ -0,0 +1,124 @@
...
+            resolve_latest() {
+              local output
+              output=$(echo "$1" \
+                | uv pip compile - --index-url "$INDEX_URL" --no-deps --quiet 2>&1) || {
+                echo "ERROR: uv pip compile failed for '$1':" >&2
+                echo "$output" >&2
+                exit 1
+              }
+              echo "$output" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || true
+            }
```
sethfitz: This re-implements `resolve_latest` plus the whole `vnext`/`main`/`main-bump` formula from the `compute-version` composite action — and they've already drifted a bit (`if [ -n … ]` + logging in the action, `${LATEST:-…}` here), and both carry the same no-match bug. Could the dry-run call the action per package instead? The awkward part is that the action emits one `version` output while this wants a table, so it'd be a step (or matrix) per package rather than a bash loop — more verbose, but one copy of the formula to keep correct. Feels worth doing before Phase 3 forks these any further.

_Surfaced by a Claude-assisted review pass._

.github/actions/compute-version/action.yml
```diff
@@ -0,0 +1,114 @@
...
+            else
+              NEXT_PATCH=0
+              echo "No published version in ${MAJOR_MINOR}.x series — starting at patch 0"
+            fi
+            VERSION="${MAJOR_MINOR}.${NEXT_PATCH}"
+            ;;
+
+          main-bump)
+            VERSION="${MAJOR_MINOR}.0"
+            echo "Major/minor bump — patch resets to 0"
```
sethfitz, line 122: Forward-looking, not for this PR: the compute path assumes the number it produces is unpublished — `main-bump` hands back `${MAJOR_MINOR}.0` unconditionally, and the `main` fallback (L96) does the same. The "fail if this version already exists" guard lives over in `reusable-check-python-package-versions.yaml` and isn't wired to this action. When Phase 3 connects `compute-version` to the actual publish, is the plan to keep that existence check in front of it? Flagging so it doesn't slip between the two.

_Surfaced by a Claude-assisted review pass._

sethfitz (dismissed):

packages/overture-schema-addresses-theme/src/overture/schema/addresses/__about__.py
sethfitz: (If you don't end up dropping `__about__.py` here or in an immediate follow-up, mind opening an issue to track it?)

CONTRIBUTING.md
lowlydba: I honestly can't recall pinning pretty starting versions, but I'm a bit opposed to picking arbitrary version numbers just so they're equal across packages. I am not sure I see the value for the additional work, and its something we could do at any time in the future (artificially bumping versions) - though it is a bit opposed to the idea of a (semantic) version being meaningful in the first place. Its definitely something that can be done anytime, outside of this effort.

.github/actions/compute-version/action.yml
lowlydba: Not going to bother - we can include some sort of PEP linting if we're worried folks will mis-version things, but I don't think its worth adding validation code across several places for a well-known standard at this juncture

packages/overture-schema-addresses-theme/src/overture/schema/addresses/__about__.py
lowlydba: http://github.com/OvertureMaps/schema/issues/549

.github/actions/compute-version/action.yml
lowlydba: The main risk case here — a major/minor bump landing on `main` — is already covered pre-merge: `check-python-package-versions.yaml` runs on every PR touching `pyproject.toml` and fails if the human-edited target version already exists in CodeArtifact/PyPI. Since `main-bump` in `compute-version` just extracts `<major>.<minor>` from that same pyproject.toml value, that check already guards the exact version `compute-version` would produce.

The `vnext` and `main` (auto-patch) paths are structurally safe by construction — `vnext` appends a unique `+dev.<run_number>`, and `main` increments off the latest published patch — so collisions there would only come from a race between concurrent runs, which is a Phase 3 CI-serialization concern, not something this existence check would catch anyway.

So this is intentionally phased, not a gap — Phase 3 (#509) owns the actual publish wiring, and the existing PR-time gate already covers the one path where a stale/duplicate version could slip through. Nothing to change in this PR.

.github/actions/compute-version/action.yml
lowlydba: Good catch — confirmed your repro. 
`resolve_latest()` now checks the failure output for "no solution found"/"could not find a version"/"not found in the package registry" and treats only those as the benign empty-result case; anything else still hard-fails. Verified the fallback path is reachable again with a synthetic repro of both branches.

.github/workflows/compute-versions-dry-run.yaml
lowlydba: Went with your suggestion — refactored to a discover job (context + package list) feeding a per-package matrix that calls ./.github/actions/compute-version directly, so there's one copy of the formula. Also added a pull_request trigger scoped to the two composite actions + this workflow, so future edits to either get smoke-tested automatically.

uv.lock
lowlydba: Confirmed and fixed in 28812d4 - ran `uv lock`, which updated all 12 packages from `0.1.0` to `0.1.1` and `overture-schema` from `1.17.0` to `1.17.1` to match `pyproject.toml`. Also added `UV_LOCKED=1` to the default-resolution CI matrix cell in `check-python-code.yaml` (via the existing `make check` path) so a stale lock fails fast next time, without a dedicated check - and fixed the `Makefile`'s `uv-sync` target, which was silently swallowing the error output that would explain the failure.

.github/workflows/compute-versions-dry-run.yaml
lowlydba: You can check out the test runs for all the relevant actions: https://github.com/OvertureMaps/schema/actions/runs/28950760186?pr=534

CONTRIBUTING.md
sethfitz: The "arbitrary" starting version would be whatever the schema version is (as consumed by the data pipeline team) at the time we first publish. That way the versions start from a common point and diverge semantically by package (which wasn't possible previously).

uv.lock
sethfitz: Thanks for the `UV_LOCKED` and `Makefile` fixes.

sethfitz (dismissed):

lowlydba:
@copilot resolve the merge conflicts in this pull request

lowlydba:
Apologies for yet another review request @sethfitz , @ericgodwin  - had to fix a merge conflict on the `uv.lock` file 🙃

sethfitz (approved):

ericgodwin (approved):

PR #535, opened 2026-05-21
[ENHANCEMENT] Added provider, resource, and version parameters to sources

### LINKED ISSUE #530: Update sources details to provide the version

ericgodwin:
# Background

The intent of this change is to update our source item field to include the information necessary for data provenance:
```
- provider: The name of the entity that produced the data: meta, esri, microsoft, osm, etc.
- resource: The subject or type of data given by the provider: division-names, buildings, planet, etc.
- version: The sortable identifier such as a date or number: 2026-02-13, 5.3, A5692
```
Together, along with the version_id these values allow a user to uniquely identify what raw input data was used to construct Overture data. Our current system, of providing only a dataset is lacking dataset version information but is also inconsistently constructed. All three new fields will be nullable and optional to start as this is the first step where we are making it so the pipeline can populate these fields.

## Major change release plan
While this change in itself is not a breaking change, it is part of a larger plan with major impact. The rough timeline for these changes is:

- Update the schema to add `provider` + `resource` + `version` details as optional fields (this PR - June / July)
- Non-schema work for pipelines to populate the new fields
- Update the schema to A) make `provider` + `resource` + `version` required fields and B) mark `dataset` as deprecated and make it an optional field. (BREAKING - September)
- Update the schema and code to remove the `dataset` field. (BREAKING - March 2027 or later)

Messaging around this change is that the current method of providing provenance is not sufficient to ensure traceability. Besides documenting the deprecation of `dataset` we will want to provide details on how the `provider`,  `resource`, `version` work together to identify a data snapshot.

Closes #530

# Testing

A couple of new examples / counterexamples have been added. In particular one to check that the length of each of the provided fields is at least 1 and a second which shows what properly populated fields look like.

The tests were then run with the following results:

```
  ## Test Results
  
  | Result | Count |
  |--------|-------|
  | ✅ Passed | 1993 |
  | ❌ Failed | 0 |
  | ⚠️  Errors | 0 |

  **Duration:** 5.21s
```

# Documentation website
[Docs preview for this PR.](https://staging.overturemaps.org/schema/pr/535/schema/index.html)

### DIFF
--- a/packages/overture-schema-common/src/overture/schema/common/sources.py
+++ b/packages/overture-schema-common/src/overture/schema/common/sources.py
@@ -86,6 +86,39 @@ class SourceItem(BaseModel):
             """).strip()
         ),
     ] = None
+    provider: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^[a-z0-9][a-z0-9._-]*$",
+            description=textwrap.dedent("""
+                The provider label (lowercase with no white space) for the entity that contributed this data
+                (e.g., osm, meta, esri).
+            """).strip(),
+        ),
+    ] = None
+    resource: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^[a-z0-9][a-z0-9._-]*$",
+            description=textwrap.dedent("""
+                The subject or type of data contributed by the provider (lowercase with no white space)
+                (e.g., planet, buildings, division-names).
+            """).strip(),
+        ),
+    ] = None
+    version: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^\S+$",
+            description=textwrap.dedent("""
+                A sortable identifier for the specific snapshot of the resource
+                (e.g., 2026-02-13, 5.3, A5692).
+            """).strip(),
+        ),
+    ] = None
 
 
 Sources = NewType(
--- a/packages/overture-schema-common/src/overture/schema/common/sources.py
+++ b/packages/overture-schema-common/src/overture/schema/common/sources.py
@@ -12,7 +12,12 @@ from overture.schema.common.scoping import Scope, scoped
 from overture.schema.common.types import ConfidenceScore
 from overture.schema.system.field_constraint import UniqueItemsConstraint
 from overture.schema.system.model_constraint import no_extra_fields
-from overture.schema.system.string import JsonPointer, StrippedString
+from overture.schema.system.string import (
+    JsonPointer,
+    NoWhitespaceString,
+    SnakeCaseString,
+    StrippedString,
+)
 
 
 @no_extra_fields
@@ -87,32 +92,29 @@ class SourceItem(BaseModel):
         ),
     ] = None
     provider: Annotated[
-        str | None,
+        SnakeCaseString | None,
         Field(
             min_length=1,
-            pattern=r"^[a-z0-9][a-z0-9._-]*$",
             description=textwrap.dedent("""
-                The provider label (lowercase with no white space) for the entity that contributed this data
+                The provider label for the entity that contributed this data
                 (e.g., osm, meta, esri).
             """).strip(),
         ),
     ] = None
     resource: Annotated[
-        str | None,
+        SnakeCaseString | None,
         Field(
             min_length=1,
-            pattern=r"^[a-z0-9][a-z0-9._-]*$",
             description=textwrap.dedent("""
-                The subject or type of data contributed by the provider (lowercase with no white space)
-                (e.g., planet, buildings, division-names).
+                The subject or type of data contributed by the provider
+                (e.g., planet, buildings, division_names).
             """).strip(),
         ),
     ] = None
     version: Annotated[
-        str | None,
+        NoWhitespaceString | None,
         Field(
             min_length=1,
-            pattern=r"^\S+$",
             description=textwrap.dedent("""
                 A sortable identifier for the specific snapshot of the resource
                 (e.g., 2026-02-13, 5.3, A5692).
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -260,13 +260,14 @@ description: Common schema definitions shared by all themes
       enum: [left, right]
     sourcePropertyItem:
       description: >-
-        An object storing the source for a specificed property. The property
+        An object storing the source for a specified property. The property
         is a reference to the property element within this Feature, and will be
         referenced using JSON Pointer Notation RFC 6901
-        (https://datatracker.ietf.org/doc/rfc6901/). The source dataset for
-        that referenced property will be specified in the overture list of
-        approved sources from the Overture Data Working Group that contains
-        the relevant metadata for that dataset including license source organization.
+        (https://datatracker.ietf.org/doc/rfc6901/). The source is currently
+        identified by the dataset field, but will, in the future, be replaced
+        by the combination of provider, resource, and version fields.
+        Additional metadata such as license, record_id, update_time, and
+        confidence may also be provided.
       type: object
       required: [property, dataset]
       allOf:
@@ -292,14 +293,31 @@ description: Common schema definitions shared by all themes
           type: number
           minimum: 0
           maximum: 1
+        provider:
+          type: string
+          description: >-
+            The Provider Label for the entity that produced this data
+            (e.g. osm, meta, esri).
+          minLength: 1
+        resource:
+          type: string
+          description: >-
+            The subject or type of data provided by the provider (e.g. planet,
+            buildings, division-names).
+          minLength: 1
+        version:
+          type: string
+          description: >-
+            A sortable identifier for the specific snapshot of the resource
+            (e.g. 2026-02-13, 5.3, A5692).
+          minLength: 1
     sources:
       description: >-
         The array of source information for the properties of a
         given feature, with each entry being a source object which
-        lists the property in JSON Pointer notation and the dataset
-        that specific value came from. All features must have a root
-        level source which is the default source if a specific
-        property's source is not specified.
+        lists the property in JSON Pointer notation. All features
+        must have a root level source which is the default source
+        if a specific property's source is not specified.
       type: array
       items: {"$ref" : "#/$defs/propertyDefinitions/sourcePropertyItem"}
       minItems: 1
--- a/schema/defs.yaml
+++ b/schema/defs.yaml
@@ -294,19 +294,19 @@ description: Common schema definitions shared by all themes
           minimum: 0
           maximum: 1
         provider:
-          type: string
+          type: [string, "null"]
           description: >-
             The Provider Label for the entity that produced this data
             (e.g. osm, meta, esri).
           minLength: 1
         resource:
-          type: string
+          type: [string, "null"]
           description: >-
             The subject or type of data provided by the provider (e.g. planet,
             buildings, division-names).
           minLength: 1
         version:
-          type: string
+          type: [string, "null"]
           description: >-
             A sortable identifier for the specific snapshot of the resource
             (e.g. 2026-02-13, 5.3, A5692).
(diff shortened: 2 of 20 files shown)


schema/defs.yaml
```diff
@@ -292,14 +293,31 @@ description: Common schema definitions shared by all themes
...
+          description: >-
+            The Provider Label for the entity that produced this data
+            (e.g. osm, meta, esri).
+          minLength: 1
+        resource:
+          type: string
+          description: >-
+            The subject or type of data provided by the provider (e.g. planet,
+            buildings, division-names).
+          minLength: 1
```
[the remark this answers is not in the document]
ericgodwin: Updated the PR description to say at least 1 for the length.

schema/defs.yaml
```diff
@@ -292,14 +293,31 @@ description: Common schema definitions shared by all themes
...
       description: >-
         The array of source information for the properties of a
         given feature, with each entry being a source object which
-        lists the property in JSON Pointer notation and the dataset
-        that specific value came from. All features must have a root
-        level source which is the default source if a specific
-        property's source is not specified.
+        lists the property in JSON Pointer notation. All features
+        must have a root level source which is the default source
+        if a specific property's source is not specified.
```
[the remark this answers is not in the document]
ericgodwin: I am personally okay with the description as is. Adding that dataset is a required field within the parent field seems a bit over the top.

schema/defs.yaml
```diff
@@ -292,14 +293,31 @@ description: Common schema definitions shared by all themes
           type: number
           minimum: 0
           maximum: 1
+        provider:
+          type: [string, "null"]
+          description: >-
+            The Provider Label for the entity that produced this data
```
vcschapp: Incredibly pedantic wording nits - would "contributed" be preferable to "produced"?

I'd be tempted to lowercase provider & label.

vcschapp (changes requested):
@ericgodwin we finally switched from "JSON Schema only" to "mixed JSON Schema/Pydantic (you have to update both)" to "pure Pydantic", so the place to update is [`packages/overtures-schema-common/src/overture/schema/common/sources.py`](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-common/src/overture/schema/common/sources.py).

Should just be a matter of inserting fields into `SourceItem` [here-ish](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-common/src/overture/schema/common/sources.py#L62)...

Giving an example of one field:

```python
provider: Annotated[
    StrippedString | None,
    Field(
        min_length=1,
        description=textwrap.dedent("""
            The provider label for the entity that contributed this data (e.g., osm, meta, esri).
        """).strip()
    ),
] = None
```

- The "union with `None`" (`| None`) is doing the heavy lifting that allows these fields to be nullable for the time being.
- IMO `StrippedString` is probably a reasonable base field type for all three fields because it's hard to imagine a good reason why sourcing fields would need to begin or end with whitespace...

---

Technically all of `defs.yaml` and the `counterexamples/` and `examples/` directories are "legacy" and shouldn't need to be updated.

The new analogue to `defs.yaml` is the Python source code mostly in `packages/overtures-schema-common` and the reference examples and counterexamples under `reference/`.

We schema dwellers need to do a better job keeping the repo updated and communicating this!

schema/defs.yaml
ericgodwin: Fixed

ericgodwin:
@vcschapp I went ahead and updated the Pydantic model as you indicated with a couple of slight changes.

- Updated sources.py as suggested
- Did not used StrippedString as I explicitly called out the format with a regex. I don't want embedded spaces either
- Made provider and resource lowercase only
- Added some clarity to the descriptions.

schema/defs.yaml
ericgodwin: Changed

packages/overture-schema-common/src/overture/schema/common/sources.py
```diff
@@ -86,6 +86,39 @@ class SourceItem(BaseModel):
             """).strip()
         ),
     ] = None
+    provider: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^[a-z0-9][a-z0-9._-]*$",
```
vcschapp: RE: this pattern, would you be OK with [`SnakeCaseString`](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-system/src/overture/schema/system/string.py#L143-L146)?

The [pattern](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-system/src/overture/schema/system/field_constraint/string.py#L205) is almost the same, just a tiny bit more restrictive. We use it fairly consistently for "identifier-like" string data.

packages/overture-schema-common/src/overture/schema/common/sources.py
```diff
@@ -86,6 +86,39 @@ class SourceItem(BaseModel):
...
+                The provider label (lowercase with no white space) for the entity that contributed this data
+                (e.g., osm, meta, esri).
+            """).strip(),
+        ),
+    ] = None
+    resource: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^[a-z0-9][a-z0-9._-]*$",
```
vcschapp: This could probably also be `SnakeCaseString`...

packages/overture-schema-common/src/overture/schema/common/sources.py
```diff
@@ -86,6 +86,39 @@ class SourceItem(BaseModel):
...
+                The subject or type of data contributed by the provider (lowercase with no white space)
+                (e.g., planet, buildings, division-names).
+            """).strip(),
+        ),
+    ] = None
+    version: Annotated[
+        str | None,
+        Field(
+            min_length=1,
+            pattern=r"^\S+$",
```
vcschapp: Suggest [`NoWhitespaceString`](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-system/src/overture/schema/system/string.py#L109-L119) here.

schema/defs.yaml
```diff
@@ -292,14 +293,31 @@ description: Common schema definitions shared by all themes
           type: number
           minimum: 0
           maximum: 1
+        provider:
+          type: string
+          description: >-
+            The Provider Label for the entity that produced this data
+            (e.g. osm, meta, esri).
+          minLength: 1
```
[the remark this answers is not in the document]
vcschapp: This is true.

FWIW I'd ignored this in the previous change because I was thinking we would just drop the JSON Schema changes altogether, so it would go away. That's probably the simplest course—reset `defs.yaml` and the top-level examples/counterexamples to their HEAD state and just leave the JSON Schema changes.

If we do keep these, the simplest course is just to set `type: string` and the fact that they are not in the `required: [...]` list will do what's needed.

IMO the simplest course is just to drop the JSON Schema changes out of the CR. JSON Schema's on a deprecation path and the ship seems to have sailed on keeping it up to date with the Python.

packages/overture-schema-common/src/overture/schema/common/sources.py
ericgodwin: Changed to SnakeCaseString

packages/overture-schema-common/src/overture/schema/common/sources.py
ericgodwin: Changed to SnakeCaseString

packages/overture-schema-common/src/overture/schema/common/sources.py
ericgodwin: Changed to NoWhiteSpaceString

schema/defs.yaml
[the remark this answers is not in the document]
ericgodwin: @vcschapp if the JSON schema is on a deprecation path or already out of date shouldn't we remove it? Keeping an out of sync version of the schema really seems more confusing that anything else especially since folks may be inclined to use the old version as they are more familiar with it.

ericgodwin:
@vcschapp the big open issue here is whether we should update the JSON schema or just drop those changes. I would argue that as long as the JSON schema is included in the repo we should keep it up-to-date. At this point we have not deprecated anything and users may be using the JSON version still.

schema/defs.yaml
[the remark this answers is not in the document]
vcschapp: We should, yes. It's mostly been a resource issue so far - time has been scarce, and we wanted to announce it properly.

vcschapp:
> Victor Schappert (@vcschapp) the big open issue here is whether we should update the JSON schema or just drop those changes. I would argue that as long as the JSON schema is included in the repo we should keep it up-to-date. At this point we have not deprecated anything and users may be using the JSON version still.

I'm fine with that.

If we don't drop those changes, the action item switches to flipping the `type: [string, "null"]` to just `type: string`. 

The quoted `"null"` indicates "the string n-u-l-l" rather than "the JSON `null` value". The intention is `type: [string, null]` but in the "legacy" JSON Schema we achieve similar-enough-results by just not marking the field as required.

This is the issue [that Copilot comment](https://github.com/OvertureMaps/schema/pull/535#discussion_r3284018383) is raising, I'm just guiding to simplify down to `type: string` instead of `type: [string, null]` for consistency reasons.

The reason I didn't make this into a big issue from the start is it seemed easier to me just to remove the "legacy" JSON Schema changes so I was pushing in that direction, since that's what we'd agreed on in the WG.

ericgodwin:
> This is the issue [that Copilot comment](https://github.com/OvertureMaps/schema/pull/535#discussion_r3284018383) is raising, I'm just guiding to simplify down to `type: string` instead of `type: [string, null]` for consistency reasons.

Hello @vcschapp. I went ahead and removed the "null" values. This makes it so the only inconsistency between the Pydantic and JSON versions is that Pydantic requires SnakeCase. I think I'm okay with that given the deprecation path.

atiannicelli (dismissed):
Nice!

lowlydba (dismissed):
LGTM

vcschapp (dismissed):

sethfitz (dismissed):

lowlydba (dismissed):

lowlydba (approved):

vcschapp (dismissed):

vcschapp:
Anyone understand why merging is blocked here?

Does one of myself or @lowlydba not have write access?

RoelBollens-TomTom (approved):

RoelBollens-TomTom (approved):

RoelBollens-TomTom (dismissed):

lowlydba (dismissed):

lowlydba (approved):

RoelBollens-TomTom (approved):

lowlydba:
Changed some settings again, hoping this works properly now. 

Still trying to figure out the weird state from before, this is a first for me.

PR #546, opened 2026-06-18
[BUG] Divisions: allow both is_land and is_territorial to be true

### LINKED ISSUE #542: Divisions: allow True for both is_land and is_territorial
[division_area.py](https://github.com/OvertureMaps/schema/blob/main/packages/overture-schema-divisions-theme/src/overture/schema/divisions/division_area.py#L56) uses `@radio_group("is_land", "is_territorial")` which requires exactly one of these fields to be `True`. But the [YAML schema](https://github.com/OvertureMaps/schema/blob/main/schema/divisions/division_area.yaml#L42-L48) uses anyOf, meaning both can be `True`, which is correct since a landlocked country has the same geometry for both.

It doesn't look like an existing decorator matches this so a new decorator would be required (`@require_any_true`?). Then division_area.py would need to be updated.

atiannicelli:
## Summary

Replaces the `@radio_group` constraint (exactly one must be true) with a new `@require_any_true` constraint (at least one must be true) on `DivisionArea` and `DivisionBoundary` models. This allows both `is_land` and `is_territorial` to be `true` simultaneously.

## Changes

- **New `@require_any_true` constraint** in `overture-schema-system` — validates that at least one field in a group of `bool` fields is `True`, generating `anyOf` in JSON Schema (vs `oneOf` from `@radio_group`)
- **Updated `DivisionArea` and `DivisionBoundary`** to use `@require_any_true("is_land", "is_territorial")`
- **Updated hand-written YAML schema** (`division_boundary.yaml`) from `oneOf` → `anyOf`
- **Removed `bad-not-both` counterexamples** (now valid) and added `both_land_territorial` positive examples
- **Updated codegen** `model_constraints.py` to handle the new constraint type

Closes #542


packages/overture-schema-system/src/overture/schema/system/model_constraint/require_any_true.py
```diff
@@ -0,0 +1,152 @@
...
+        f"@{require_any_true.__name__}", *field_names
+    )
+
+    return model_constraint.decorate
+
+
+class RequireAnyTrueConstraint(FieldGroupConstraint):
+    """
+    Class implementing the `require_any_true` decorator, which can also be used standalone.
+    """
```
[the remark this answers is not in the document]
atiannicelli: fixed

packages/overture-schema-system/src/overture/schema/system/model_constraint/require_any_true.py
```diff
@@ -0,0 +1,151 @@
+"""
```
vcschapp, line 1: Can we generalize this slightly so that `require_any_true` takes a list of one or more `Condition` values? If that's workable then you could have, for this use case:

```python
@require_any_true(FieldEqCondition('is_land', True), FieldEqCondition('is_territorial', True))
class DivisionArea(...):
    ...
```

And for other use cases you can create more complex requirements this way, like "either field A should be set to 'foo' or field B should be set to 'bar'".

vcschapp (commented):
Two comments:

1. It might be useful to generalize `require_any_true`.
2. The Python code doesn't use `examples/` and `counterexamples/` it uses `reference/examples/` and `reference/counterexamples/` - apologies for this inconsistency but could these be updated also?

packages/overture-schema-system/src/overture/schema/system/model_constraint/require_any_true.py
atiannicelli: Done

lowlydba (approved):

TristanDiet-TomTom (approved):

sethfitz (approved):

stepps00 (approved):

PR #552, opened 2026-07-10
[REFACTOR](schema) Split system.primitive into numeric and geometric modules

### LINKED ISSUE #464: "Primitive Types" → "Numeric Types"
We use "primitive" as a catch-all for the numeric types (`int8`, `float32`, etc.). These manifest in the schema reference as a single page with a path and title that isn't quite right: https://docs.overturemaps.org/schema/reference/system/primitive/primitives/

This may also involve moving the numeric types from `.../primitive/__init__.py` into their own module (`numeric.py`) and re-exporting through `__all__`.

sethfitz:
Closes #464.

Splits `overture.schema.system.primitive` along its two concerns:

- The numeric NewTypes (`uint8` through `float64`) move to `overture.schema.system.numeric`
- `BBox` and `Geometry` move to `overture.schema.system.geometric`
- The `primitive` package is removed; imports updated across ~35 files

The generated reference page for the numeric types moves from `system/primitive/primitives` ("Primitive Types") to `system/numeric` ("Numeric Types"), and the geometry page from `system/primitive/geometry` to `system/geometry`. Remaining "primitive" wording in system-package docstrings is retired with it.

The BBox/Geometry move implements @vcschapp's suggestion from the issue thread; flagging it here since it wasn't settled there.


packages/overture-schema-codegen/docs/walkthrough.md
```diff
@@ -443,9 +444,9 @@ independent `if` statements, not `elif`.
 `build_placement_registry` builds the complete `dict[TypeIdentity, PurePosixPath]`
 mapping each type to its output file path. Four tiers:
 
-Aggregate pages come first. All numeric primitives point to
-`system/primitive/primitives.md`. All geometry types point to
-`system/primitive/geometry.md`. These are hardcoded paths since the types share a single
+Aggregate pages come first. All numeric types point to
+`system/numeric.md`. All geometry types point to
+`system/geometry.md`. These are hardcoded paths since the types share a single
```
vcschapp: Given the model names of `geometric`/`numeric` it would make sense to rename the output MD file to `geometric.md`.

vcschapp (commented):
1. Since a couple of MD output filenames changed (`primitive.md` - gone; `numeric.md` - new) are there any external hard-code links to these files we have to keep in sync?
2. I made one inline comment suggesting maybe `geometry.md` should become `geometric.md` for consistency.

Apart from the above minor questions/comments, fully on board and ready to ship this!

vcschapp (dismissed):

vcschapp (approved):

RoelBollens-TomTom (approved):

Issue #570, opened 2026-07-17
Generalize `field_path` to arbitrary map/array nesting


sethfitz:
`field_path` (added in #518, now #569) models a path to a value through a nested schema. Today it splits paths into three classes — `ScalarPath` (no iteration), `ArrayPath` (array iteration), `MapPath` (one map projection, struct-only leaf) — and forbids anything that mixes or nests containers. @vcschapp [flagged in review](https://github.com/OvertureMaps/schema/pull/518/changes#r3598512995) that those limits feel arbitrary, and they are: the codegen generates validation for the schema's full type space, not for the shape the schema happens to have today, and these limits bake the current shape in as a ceiling. The direction is to generalize, not to trim `field_path` down to what's currently used.

## What's restricted today

A `MapPath` requires the map to be reachable without array iteration and its leaf to be struct-only. `parse` rejects `dict[K, list[V]]` (an array marker after a map projection) and multiple map markers; `promote_terminal_map` / `promote_terminal_array` raise `NotImplementedError` for a map under an array element, a map under a map element, or an array under a map element. In both Pydantic and Spark those shapes are ordinary and nest arbitrarily deep, so nothing about the data forces the restriction — it comes from the path grammar and the three-class split.

## Why it hasn't surfaced yet

The only map in the schema right now is `CommonNames` (`dict[LanguageTag, StrippedString]`), a `dict[K, scalar]` that `MapPath` handles. So the guards above are unexercised, not validated as acceptable — the first richer map field (a `dict[K, list]`, a `dict[K, Model]` reached through an array, a nested map) hits a `NotImplementedError` at codegen time. Generalizing now keeps that from being a surprise later.

## The design is already half-generalized

`MapPath` carries a non-empty-leaf case for `dict[K, Model]` (e.g. `subs{value}.label`) that no field exercises yet, while the container axis stays stubbed with `NotImplementedError`. The leaf direction was generalized and the nesting direction wasn't; the fix is to finish the generalization, not retreat to `dict[K, scalar]`.

## The work

- Collapse the taxonomy to `Direct` (no iterating segment — today's `ScalarPath`) versus a single iterated path type whose `segments` freely mix `ArraySegment` and `MapSegment`. `ArrayPath` and `MapPath` merge, and their helper properties (`array_chunks`, `iter_struct_paths`, `map_column`, `leaf`) generalize to walk a mixed segment sequence.
[...]

PR #571, opened 2026-07-17
[REFACTOR] Inline PathSegment and expand the Check docstring

### LINKED ISSUE #586: Address remaining actionable comments from the #518 pyspark review
@vcschapp's review of #518 (superseded by #569) left two actionable inline comments that were deferred to a follow-up rather than folded into that PR. This tracks clearing them.

- Inline the `PathSegment` alias (`StructSegment | ArraySegment`) at its use sites and drop it. `FieldSegment` already names the general `Struct | Array | Map` case, so `PathSegment` is a misleading middle name for a subset that excludes `MapSegment`.
- Expand the `Check` docstring to relate `expr` (the single composed `Column`) and `read_columns` (its read-set), and to explain why they are carried separately.

No behavior change. The broader naming and map-generality threads from the same review are tracked separately in #570.

sethfitz:
Follows up @vcschapp's review of #518 (superseded by #569), addressing the two actionable inline comments. The naming and map-generality threads from that review are tracked separately in #570.

- Inline the `PathSegment` alias (`StructSegment | ArraySegment`) at its two use sites and drop it. `FieldSegment` already names the general `Struct | Array | Map` case, so `PathSegment` was a misleading middle name for a subset that excludes `MapSegment`.
- Expand the `Check` docstring to relate `expr` (the single composed `Column`) and `read_columns` (its read-set), and to say why they are carried separately.

No behavior change.

Closes #586
Related: #570

### DIFF
--- a/packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
+++ b/packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
@@ -27,12 +27,21 @@ class Check:
     and report grouping), not how to access the data.  The expression in
     `expr` already encodes the access pattern.
 
-    `read_columns` names every top-level schema column the expression
-    dereferences -- one for a plain field check, several for a model-level
-    check that spans columns, plus any discriminator a variant gate reads.
-    `validate_model` drops a check when any column it reads is skipped or
-    structurally absent, so an unresolvable `F.col()` never reaches Spark;
-    it also treats these as the columns a check can be suppressed by name.
+    `expr` and `read_columns` are two views of one computation, and each is
+    a "column" in a different sense.  `read_columns` are real columns of the
+    underlying schema model -- the top-level columns the check must read to
+    evaluate.  There is always at least one; a model-level constraint that
+    spans fields names several, plus any discriminator a variant gate reads.
+    `expr` is a *virtual column*: it is not a column of the schema model but
+    one synthesized by the generated validation machinery to hold the
+    composed expression the Spark engine evaluates.  The two travel together
+    because the builder knows the read-set as it composes `expr`; recording
+    it is surer than recovering it from the finished `Column`.
+
+    `validate_model` drops a check when any column in `read_columns` is
+    skipped or structurally absent, so an unresolvable `F.col()` never
+    reaches Spark; it also treats these as the columns a check can be
+    suppressed by name.
     """
 
     field: str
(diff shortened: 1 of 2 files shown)


packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
```diff
@@ -27,12 +27,19 @@ class Check:
     and report grouping), not how to access the data.  The expression in
```
vcschapp, line 27: uber-nit: two spaces after full stop instead of one

packages/overture-schema-pyspark/src/overture/schema/pyspark/check.py
```diff
@@ -27,12 +27,19 @@ class Check:
...
-    structurally absent, so an unresolvable `F.col()` never reaches Spark;
-    it also treats these as the columns a check can be suppressed by name.
+    `expr` and `read_columns` are two views of one computation. `expr` is
+    the single composed `Column` that evaluates the check; `read_columns`
+    is its read-set -- every top-level schema column that one expression
+    dereferences (one for a plain field check, several for a model-level
+    check that spans columns, plus any discriminator a variant gate reads).
+    They travel together because the builder knows the columns as it
+    composes `expr`; recording them is surer than recovering them from the
+    finished `Column`.
```
vcschapp: This is a big help - thanks for the tweaking here.

I still get confused reading it, which is because I can't hold the whole design in my head at once and there are two different contexts for columns: the *model* columns and the *validator* pseudo-columns. (I'm calling them pseudo-columns, to coin a term, just to emphasize that while derived from the model by the PySpark code generator they aren't actually model columns.)

Can I suggest an alternative?

> `read_columns` represents the actual columns of the underlying schema model that have to be read to complete the check; there is always at least one of these columns, but for model-level multi-column constraints there may be more than one. `expr` can be thought of as a pseudo-column generated by the generated validation machinery. It is a column that does not exist on the underlying schema model but is instead generated by the validation machinery to hold the final validation expression to be evaluated by the Spark engine.

vcschapp (commented):
LGTM, one suggestion for the ` check.py` docs. I'm not passionate about it, but IMO it'd convey the meaning better. 👇

vcschapp (approved):

RoelBollens-TomTom (approved):

PR #575, opened 2026-07-20
[CHORE](deps) Refresh uv.lock within the exclude-newer window

### LINKED ISSUE #573: Refresh uv.lock to current dependency versions
The workspace uv.lock has drifted behind the latest versions permitted by the exclude-newer = "1 week" window. Refresh it with uv lock --upgrade and confirm the suite stays green under both the default and lowest-direct CI cells.

pytest 9.1 deprecates class-scoped fixtures defined as instance methods (PytestRemovedIn10Warning), which the -W error suite treats as a collection error; the codegen test fixtures move to the @classmethod form as part of the refresh.

pyspark 4.2.0 (released 2026-07-14) sits inside the one-week guard and is left for a later refresh.

sethfitz:
Closes #573.

Refreshes uv.lock to the newest versions permitted by the committed exclude-newer = "1 week" window (via `uv lock --upgrade`; no pyproject constraint changes). Runtime: pyspark 4.1.1 to 4.1.2, pydantic 2.13.3 to 2.13.4, click 8.3.3 to 8.4.2, deepdiff 9.0.0 to 9.1.0. Dev: mypy 1.20 to 2.3, pytest 9.0.3 to 9.1.1, ruff 0.15.12 to 0.15.21, coverage 7.13.5 to 7.15.1, plus transitive deps.

pytest 9.1 deprecates class-scoped fixtures defined as instance methods (PytestRemovedIn10Warning), which the `-W error` suite treats as a collection error. The 17 `scope="class"` fixtures across the codegen test modules move to the `@classmethod` form, with their self references switched to cls. pytest 9.0 accepts the same form, so the lowest-direct floor cell stays green.

pyspark 4.2.0 (released 2026-07-14) is held back: it is newer than the one-week guard. It becomes eligible on a later refresh once it ages past the window.

Verification: default resolution 6307 passed under `-W error`, with mypy 2.3, ruff, and doctest clean; the lowest-direct cell (pytest 9.0.0 floor) exercises the adapted fixtures green; `uv lock --check` is consistent.


vcschapp (dismissed):
Approved but needs to be merge-deconflicted due to merging of PR #569.

vcschapp (approved):

TristanDiet-TomTom (approved):

PR #577, opened 2026-07-20
[TEST](pyspark) Batch per-test DataFrames into one collect

### LINKED ISSUE #576: Batch per-test DataFrames in the PySpark expression tests
The expression-helper tests (`test_column_patterns.py`, `test_constraint_expressions.py`) build a DataFrame and `collect()` once per test — roughly 200 Spark jobs between the two files. Each pair carries fixed scheduling overhead, so the files spend most of their wall time on setup rather than assertions.

Both files are single-row check assertions: an input column, a check expression over it, and a predicate on the collected result. That shape batches cleanly — one wide single-row DataFrame holding every case's input, one `select` applying every check, one `collect` — the same batch-once pattern the generated conformance harness already uses.

Expected effect: `test_column_patterns.py` ~6.4s → ~4.0s and `test_constraint_expressions.py` ~16.9s → ~5.1s, with the remaining floor being Spark session startup. No coverage change; purely a test-structure speedup.

sethfitz:
Follow-up to #572, stacked on its `generalize-field-path` branch (the column_patterns cases use the `nested_map_{keys,values}_check` helpers #572 adds). I'll rebase onto the integration branch once #572 merges.

Closes #576.

## What changed

`test_column_patterns.py` and `test_constraint_expressions.py` each built a DataFrame and `collect()`ed once per test — roughly 200 Spark jobs between them. Both now use a case registry: each `_Case` carries its input column, the check over it, and a predicate on the result; a module-scoped fixture packs every case's input into one wide single-row DataFrame, applies every check in a single `select`, and collects once. This is the batch-once pattern the generated conformance harness already uses.

## Effect

Every prior assertion is preserved — the distinct asserted literals all survive and the parametrized cases stay green.

| file | before | after | cases |
| --- | --- | --- | --- |
| `test_column_patterns.py` | 6.4s | 4.0s | 33 → 37 |
| `test_constraint_expressions.py` | 16.9s | 5.1s | 164 → 165 |

The extra cases come from splitting valid/invalid two-row tests into explicit per-input cases; no coverage is dropped. The remaining floor is Spark session startup.

### DIFF
--- a/packages/overture-schema-pyspark/tests/expressions/test_column_patterns.py
+++ b/packages/overture-schema-pyspark/tests/expressions/test_column_patterns.py
@@ -1,5 +1,20 @@
-"""Tests for column_patterns — structural PySpark composition helpers."""
+"""Tests for column_patterns -- structural PySpark composition helpers.
 
+Each helper composition is exercised as a `_Case`: a uniquely-named input
+column, the check built over it, and a predicate on the collected result. The
+`results` fixture packs every case's input into one wide single-row DataFrame,
+applies every check in one `select`, and collects once -- so the whole file
+pays for a single `createDataFrame` + `collect` instead of one pair per test
+(the same batch-once pattern the generated conformance harness uses). Cases
+needing both a violating and a clean input carry two entries (`*_invalid` /
+`*_valid`).
+"""
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any
+
+import pytest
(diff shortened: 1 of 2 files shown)


vcschapp (approved):
1.5X-3X test speed-up. I'll take it!

lowlydba (approved):

TristanDiet-TomTom (approved):

PR #578, opened 2026-07-21
[BUG](pyspark) Honor string min_length in base-row generation

### LINKED ISSUE #517: Validate Overture data on Spark against the Pydantic schema

sethfitz:
## Summary

`_value_from_check_string_min_length` (base-row test-data synthesis) ignored the descriptor's `min_length` and always returned a single character. For any string field with `min_length > 1`, the synthesized base/valid row value (`'a'`, length 1) already violates the constraint -- so the generated conformance `::valid` row would be invalid and the `::invalid` scenario's baseline wrong.

The fix reads the `min_length` bound from the descriptor and emits a filler of that length (`'a' * min_length`).

## Scope

Latent on `main` today: no live schema field declares a string `min_length > 1`, so the conformance suite stays green and generated output is unchanged. The defect surfaced while working the generalize-field-path branch, where a mixed-shape test had to use `min_length=1` rather than `3` to keep its valid row valid. Genuinely orthogonal to that branch -- it does not mask a defect in #570.

The analogous array min-length synthesizer (`_min_length_from_shape_constraints`) already reads `min_length` correctly and generates that many elements; string max-length and array max-length are not lower bounds and need no base-row value. The string min-length synthesizer was the sole offender.

Part of and partly closes #517.

### DIFF
--- a/packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/test_data/base_row.py
+++ b/packages/overture-schema-codegen/src/overture/schema/codegen/pyspark/test_data/base_row.py
@@ -683,10 +683,16 @@ def _value_from_check_enum(
 
 
 def _value_from_check_string_min_length(
-    _desc: ExpressionDescriptor, _scalar: Primitive, _cs: ConstraintSource
+    desc: ExpressionDescriptor, _scalar: Primitive, _cs: ConstraintSource
 ) -> str:
-    """Return any single character; satisfies `min_length>=1` for every schema today."""
-    return "a"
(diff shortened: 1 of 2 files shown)


vcschapp (dismissed):

vcschapp (approved):
Re-approved and I got the PR check green by editing the description with "partly closes #517".

RoelBollens-TomTom (approved):

PR #590, opened 2026-07-22
[BUG](codegen) Unwrap RootModel fields to their bare root shape

### LINKED ISSUE #583: RootModel is wrongly shaped by the PySpark codegen
## Summary

A field typed as a `RootModel` subclass validates and serializes on the Pydantic side, but the PySpark codegen extracts it as a struct with a `root` member.  Nothing fails at codegen time, the mismatch only surfaces when validating real data.

## Pydantic behavior (correct)

A hypothetical per-vehicle-type toll charge on a road segment: 

```python
from pydantic import BaseModel, RootModel

class TollChargesByVehicleType(RootModel[dict[str, int]]):
    """Toll charge keyed by vehicle type (e.g. car, hgv, bus)."""

class RoadSegment(BaseModel):
    road_class: str
    toll_charges: TollChargesByVehicleType | None = None

s = RoadSegment.model_validate(
    {"road_class": "motorway", "toll_charges": {"car": 250, "hgv": 900}}
)
s.model_dump() # {'road_class': 'motorway', 'toll_charges': {'car': 250, 'hgv': 900}}
```

Pydantic validates the *bare* root value and re-emits it without any wrapper. The `overture-schema` CLI therefore handle a RootModel-typed field with zero issues.

## PySpark Codegen behavior (wrong shape)

The pyspark codegen extraction does not treat RootModel differently. `is_model_class` is just `issubclass(obj, BaseModel)`, and `RootModel` is a `BaseModel` subclass, so extraction produces:

```
toll_charges: ModelRef(RecordSpec(name='TollChargesByVehicleType',
                fields=[FieldSpec(name='root',
                                  shape=MapOf(key=Primitive('str'), ...))]))
```

i.e. the `root` field becomes a real struct member. The generated Spark schema  would declare

```
toll_charges: struct<root: map<string,int>>
```

while actual Parquet data carries the bare `map<string,int>`.

sethfitz:
## Summary

Fixes #583. A field typed as a `pydantic.RootModel` subclass was extracted by the PySpark codegen as a struct with a synthetic `root` member, so the generated Spark schema declared `struct<root: ...>` while the Parquet data carries the bare root value. The mismatch surfaces only when validating real data.

## Root cause

`RootModel` is a `BaseModel` subclass, so in `extraction/type_analyzer.py` a RootModel-typed field reached the model terminal and resolved to a `ModelRef` — a struct whose one member is `root`.

## Fix

`analyze_type`'s `_unwrap` now intercepts a `RootModel` before terminal classification: it recurses into the `root` field's annotation and reattaches any root-level metadata via a shared `attach_field_metadata` helper, extracted here from `model_extraction` so field-level and root-level metadata share one implementation. A `RootModel[dict[str, int]]` field extracts to a bare `MapOf`, rendering as `MapType(...)` downstream — resolver or not.

No current schema model uses `RootModel`, so no generated output changes. This is a latent-bug fix that makes the codegen correct for RootModel-typed fields.

### DIFF
--- a/packages/overture-schema-codegen/docs/walkthrough.md
+++ b/packages/overture-schema-codegen/docs/walkthrough.md
@@ -144,6 +144,12 @@ outermost structural layer, which is exactly the `ArrayOf` that was just constru
 **dict** recurses separately for key and value types (with `newtype_ctx=None` for both,
 since dict keys and values are independent spines) and returns `MapOf`.
 
+**RootModel** subclasses are handled distinctly from other `BaseModel` terminals, since a
+RootModel serializes as its bare root value. `_unwrap` intercepts them just before the
+terminal, recurses into the `root` field's annotation, and reattaches any root metadata with
+`attach_field_metadata` -- exactly as a model field's own metadata reattaches. A
+`RootModel[dict[str, int]]` field yields a bare `MapOf` carrying the root type's shape.
+
 **Terminal** classification in `_terminal` handles the base case: `Any` becomes
 `AnyScalar`, `Literal` becomes `LiteralScalar`, `BaseModel` subclasses route through
 `model_resolver` (or fall back to `Primitive(source_type=cls)`), everything else becomes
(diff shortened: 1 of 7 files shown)


RoelBollens-TomTom:
Looks good, just one remark, when you have a discovered model via entry point rather than used as a field it will still produces a RecordSpec with the root field. I'll leave it up to you if you want to harden against that already as RootModels as entry points could already cause additional issues upstream.

RoelBollens-TomTom (approved):

vcschapp (approved):

sethfitz:
Good call on the entry point complication. I’ll have a follow-up PR up later this week.

PR #594, opened 2026-07-23
[BUG](codegen) Skip RootModel entry points from generation and document them as aliases

### LINKED ISSUE #593: RootModel entry points produce a spurious `root` column and go undocumented
## Summary

PR #590 (closing #583) unwrapped RootModel-typed *fields* to their bare root shape. A RootModel registered as a top-level *entry point* is still mishandled: `extract_model_spec` extracts it as a top-level `RecordSpec`, producing a spurious `root` column in the generated Spark schema, and the markdown reference omits it entirely.

## Behavior

A RootModel entry point has no record structure of its own -- it serializes as its bare root value, a named alias over that shape like a NewType. Two problems follow.

Expression generation emits a struct with a synthetic `root` column -- the same wrong shape #583 described, one level up at the entry-point boundary rather than at a field.

Markdown generation drops it entirely: a RootModel is a named alias readers still need documented, but it reaches the markdown pipeline through no field tree, since a RootModel field unwraps to its bare shape and names no type.

## Expectation

Skip RootModel entry points from expression generation -- an entry point is a contribution mechanism (a type usable as a field elsewhere), not only a top-level record to generate.

Document a RootModel entry point in markdown as a type alias over its bare root shape, reusing the NewType page since both are named aliases over an underlying shape.

## Edge case

A self-referential RootModel (`RootModel[list["Self"]]`) has no finite bare shape and no node to carry a cycle back-edge, so the bare-shape unwrap recurses forever and surfaces as an opaque `RecursionError`. It should fail with a clear error instead.

sethfitz:
## Summary

Fixes #593. Follow-up to #590, which unwrapped RootModel-typed *fields* to their bare root shape. This handles a RootModel registered as a top-level *entry point*, which was still extracted as a `RecordSpec` -- producing a spurious `root` column in the generated Spark schema -- and was omitted from the markdown reference.

## Changes

Expression generation skips a RootModel entry point silently (the same as would occur for `NewTypes` or other entry point contributions) rather than emitting a wrong schema. An entry point is a contribution mechanism, not only a generation one -- the type still resolves to its bare shape wherever it is used as a field (the #590 path).

Markdown generation documents a RootModel entry point as a type alias over its bare root value, reusing the NewType page since both are named aliases over an underlying shape. The split lives at the discovery bridge: `extract_model_spec` yields the feature/union specs both pipelines generate from (a RootModel returns `None`), and `extract_alias_spec` turns a RootModel into a `NewTypeSpec`. The markdown pipeline takes these aliases alongside the supplementary types it collects from field trees -- a RootModel is reachable no other way, since a RootModel field unwraps to its bare shape and names no type.

A self-referential RootModel (`RootModel[list["Self"]]`) now raises a clear `TypeError` instead of recursing forever into an opaque `RecursionError`. Unwrapping a RootModel erases its identity, so -- unlike a self-referential `BaseModel`, which the resolver terminates with a `starts_cycle` back-edge -- a self-referential root has no node to carry a back-edge and no finite bare-shape form. A `seen_rootmodels` frozenset threaded through `_unwrap` detects re-entry at the back-edge.

## Dependency: pydantic >= 2.13

Documenting the alias reads the RootModel's root description from `model_fields["root"].description`. Pydantic surfaces it there only from 2.13.0: [pydantic#13129](https://github.com/pydantic/pydantic/pull/13129) ("Preserve `RootModel` core metadata") carries the root field's `FieldInfo` metadata -- description included -- through to the core schema. On 2.12.0 the description stays nested in the root type's `Annotated` metadata and `model_fields["root"].description` is `None`, so the `lowest-direct` CI cell -- which resolves every direct dependency to its declared floor -- failed `test_root_field_description_used_without_docstring`.

This raises the pydantic floor from `>=2.12.0` to `>=2.13.0` across the workspace packages (with a matching `uv.lock` specifier bump). Verified under the failing configuration: an isolated `UV_RESOLUTION=lowest-direct` install resolves pydantic to 2.13.0 and the root-description test passes.


sethfitz:
> documents a RootModel entry point as a type alias over its bare root value, reusing the NewType page since both are named aliases over an underlying shape

@RoelBollens-TomTom @vcschapp this raises an interesting question (since I hadn't understood how `RootModel` worked before); should we be using `RootModel` subclasses in place of `NewType`s? I'm exploring that a bit, but wanted to flag it for you to think about as well. There might be benefits, particularly around the union handling (and the docstring workarounds, which now look like they were a smell).

RoelBollens-TomTom:
> > documents a RootModel entry point as a type alias over its bare root value, reusing the NewType page since both are named aliases over an underlying shape
> 
> Roel Bollens (@RoelBollens-TomTom) Victor Schappert (@vcschapp) this raises an interesting question (since I hadn't understood how `RootModel` worked before); should we be using `RootModel` subclasses in place of `NewType`s? I'm exploring that a bit, but wanted to flag it for you to think about as well. There might be benefits, particularly around the union handling (and the docstring workarounds, which now look like they were a smell).

I'd say probably not, but still worth discussing. When I initially saw the NewTypes (and type alias for segment, and this is before the markdown and pyspark codegen) I did wonder why RootModel wasn't used, but when internalising this, I don't think you would actually have gained anything, and in fact would have added some clunkiness (you'd need to to through the `.root` to get and set values, unless having dunder methods on the RootModel class). It does make the discovery and extensions mechanism messier but I think thats a fine trade off. The union handling I would need to think about whether that would help, but the docstring workarounds, if I assume correctly what you meant here, should have been done through a Field metadata annotation imho, as that is what Pydantic handles for JSON schema just fine.

sethfitz:
I'll be working through it for a bit longer, but here's where I think I'm landing: migrate `Segment` to a `RootModel` subclass, which preserves the discovery contract (because the union typedef will then migrate to `BaseModel`), migrate the collection containers (I don't have a clear justification, but this is what my gut says), and keep the scalar (numeric, semantic) NewTypes.

It seems like it makes the implementation of discovery and extension slightly messier, but it cleans up the interfaces a bit and allows things like `Segment`'s docstring to actually propagate correctly (I hadn't noticed that it was dropped before; this is a different workaround from what I was referring to, which was attaching them to the scalar NewTypes in a way that *also* makes Python introspection work properly, since descriptions don't surface as docstrings). I _think_ I tried to get a Field annotation to carry the description, but it was sitting at the wrong level, but that was before reworking the type analysis. I'll double check that as part of this spike.

vcschapp (approved):

lowlydba (approved):

PR #520, opened 2026-05-12
[REFACTOR] Rename `-core` package to `-common` to remove a source of confusion

vcschapp:
# Description

As discussed and agreed in the [2026-03-11 / CODING SESH 💻](https://github.com/OvertureMaps/schema-wg/issues/418) meeting, this change renames the `overture-schema-core` package to `overture-schema-common`.

The purpose of this change is to eliminate a source of confusion caused by overloading the word "core":

- Since early days, the word "core" has been used in Overture to mean *that part of the map and its schema that is of central importance and not peripheral*.
- However, in the `overture-schema-core` package, the word "core" was being used in a different sense to mean basically *shared types that are common to more than one theme schema*.

This issue has produced real confusion and consumed many minutes of unproductive discussion in meetings. It is hoped this change will eliminate the confusion going forward.

# Reviewer Notes

- Added @Rachmanin0xFF as a reviewer because this change may potentially break usage of the Pydantic schema in the data pipeline and the team should have a heads up there.

# Related

- Closes https://github.com/OvertureMaps/schema/issues/521.
- Agreed in [2026-03-11 / CODING SESH 💻](https://github.com/OvertureMaps/schema-wg/issues/418)

# Testing

```
$ make check
```

PR #469, opened 2026-03-12
[REFACTOR] Organize divisions theme Pydantic package, add missing relationships, tidy and add docs

vcschapp:
# Description

Continuing the Pydantic package polishing journey from late 2025, this PR cleans up the **divisions** theme. After this, 
only **transportation** and some final tweaks to **core** (soon to be *common*) will remain.

## tl;dr

- Aligned organization of divisions theme with addresses, base, buildings, and places.
- Added missing relationships.
- Edited and tidied docs.
- Added docs coverage.

# Merge guidance: Rebase

Two two commits in the history are intentional. This PR should be rebased.

# Journey

This PR follows a sequence of PRs I published in late 2025 that started the cleanup process, but did leave it somewhat incomplete. These are the earlier ones.

- Part 1 is in PR #397.
- Part 2 is in PR #400.
- Part 3 is in PR #403.
- Part 4 is in PR #412
- Part 5 is in PR #415.

This PR is basically episode 6.

# Documentation website

*Update the hyperlink below to put the pull request number in.*

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/469)

PR #598, opened 2026-07-24
[REFACTOR] Organize transportation theme Pydantic package, add relationships, sync JSON Schema, and tidy docs

### LINKED ISSUE #599: Refactor transportation theme Pydantic schema to match other packages
Refactor transportation rather much in the way done in PR #598.

vcschapp:
# Description
  
This brings us closer to the end of the Pydantic package organization journey: **transportation** was the last theme package still on the old layout. With this PR every theme package—addresses, base, buildings, divisions, places, and now transportation —is organized consistently.
  
Transportation differs from the earlier themes in three ways. First, `segment` is a discriminated union over road/rail/water subtypes, so it becomes a small *package* rather than a single module. Second, this PR makes a couple of *technically* backwardly-incompatible, but in practice not really, changes to the transportation segment feature to push down a few fields that have been wrongly defined at the segment level down to the level of the *road* segment: `routes` and `subclass_rules`. Third, this change back ports a few changes on to the old legacy JSON Schema to keep it in sync-ish until we delete it. The back-ported changes are the two already mentioned plus removing `level` from `segment` to close the loop on a change @sethfitz made in PR #390.
  
  ## tl;dr
  
- Aligned the organization of the transportation theme with addresses, base, buildings, divisions, and places.
- `segment` becomes a package (`road.py`, `rail.py`, `water.py` + a private `_common.py` for the shared base); `connector` becomes a flat module.
- Added missing connector/segment relationships on `DestinationRule` and `ProhibitedTransitionSequenceEntry`.
- Reconciled model ↔ JSON Schema: made `routes` and `subclass_rules` road-only (back-ported to the JSON Schema), and removed the scalar `level` from JSON Schema since #390 removed it from the Pydantic schema.
- Tidied docs: `make docformat` clean, converted `SegmentSubtype`/`AccessType` to `DocumentedEnum`, and documented the previously-undefined `designated` access type.

# Merge guidance: Squash
  
The work is captured as a single squashed commit, so this PR should be squash-merged.
  
  # Journey
  
- Part 1 — PR #397
- Part 2 — PR #400
- Part 3 — PR #403
- Part 4 — PR #412
- Part 5 — PR #415
- Divisions — PR #469
- Rename `-core` to `-common` — PR #520
- **Transportation (this PR)** — 👈 you are here
  
# Testing
  
`make check` passes (lint, mypy, doctest, and the full test suite). JSON Schema baselines were regenerated to reflect the reconciled model.
  
# Documentation website
  
[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/598)

# Related Issues

Closes #599.


examples/transportation/segment/road/road.yaml
```diff
@@ -28,7 +28,6 @@ properties:
     - value: gravel
   road_flags:
     - values: [is_link, is_tunnel] # Note: `is_link` is deprecated and will be removed in a future release in favor of the link subclass
-  level: -1
```
vcschapp, line 31: In #390, @sethfitz removed `level` (via `Stacked`) from transportation segments in the Pydantic schema. This is backporting that change into the legacy examples. 

The change to `segment.yaml` backports it into the legacy JSON Schema.

schema/transportation/segment.yaml
```diff
@@ -22,9 +22,7 @@ properties:
       - title: "Common Segment Properties"
         properties:
           # Classes are per-subtype following the addition of rail classes
-          subclass_rules: { "$ref": "#/$defs/propertyContainers/subclassRulesContainer" }
           access_restrictions: { "$ref": "#/$defs/propertyContainers/accessContainer" }
-          level: { "$ref": "../defs.yaml#/$defs/propertyDefinitions/level" }
```
vcschapp, line 27: In #390, @sethfitz removed `level` (via `Stacked`) from transportation segments in the Pydantic schema. This is backporting that change into the legacy JSON Schema.

RoelBollens-TomTom:
While reading through this I noticed `PYDANTIC_GUIDE.md` still references the old module paths (as well as old entry point key naming):

https://github.com/OvertureMaps/schema/blob/23585442c8a860ca3a6ae850c4429469c583d53c/PYDANTIC_GUIDE.md?plain=1#L852-L857

Feel free to change, however the guide has broader staleness that predates this PR such as the `models.py` convention and several examples reference paths like `buildings/building/models.py` that were likely removed in earlier changes. That feels out of scope for this PR.

RoelBollens-TomTom (dismissed):

packages/overture-schema-transportation-theme/src/overture/schema/transportation/segment/_common.py
```diff
@@ -0,0 +1,373 @@
+"""Shared, mode-independent segment models for the Overture Maps transportation theme.
+
+This private module holds the pieces of the segment feature that are common to every segment
+subtype (road, rail, water): the subtype discriminator, the base ``TransportationSegment``
```
sethfitz: You fell into the double backtick trap (in a bunch of places)! (They're consistently single everywhere else.)

sethfitz (dismissed):
Looks good other than the silly backticks (for a follow-up?).

sethfitz:
Thanks — filed as #604 so it doesn't ride along here. I confirmed both halves: the entry point keys are unqualified (`connector`/`segment`, not `transportation.connector`) and resolve off the theme package root, and the `models.py`-under-a-type-subdirectory convention the guide prescribes is already contradicted by buildings' flat layout — this PR removes the last theme-package `models.py`, which also breaks the guide's L857 link. Agreed it's out of scope here; the rewrite is better done once this lands so the transportation examples describe the settled structure. I've taken it.

vcschapp:
> While reading through this I noticed `PYDANTIC_GUIDE.md` still references the old module paths (as well as old entry point key naming):
> 
> https://github.com/OvertureMaps/schema/blob/23585442c8a860ca3a6ae850c4429469c583d53c/PYDANTIC_GUIDE.md?plain=1#L852-L857
> 
> Feel free to change, however the guide has broader staleness that predates this PR such as the `models.py` convention and several examples reference paths like `buildings/building/models.py` that were likely removed in earlier changes. That feels out of scope for this PR.

This is a good find - thanks. I think the right call is just to update it now, so I did that. `PYDANTIC_GUIDE.md` should have the right references now.

packages/overture-schema-transportation-theme/src/overture/schema/transportation/segment/_common.py
vcschapp: The dreaded double backtick trap! 🪤

I made some changes - should be fixed now.

RoelBollens-TomTom (dismissed):

examples/transportation/segment/road/road.yaml
```diff
@@ -28,7 +28,6 @@ properties:
     - value: gravel
   road_flags:
     - values: [is_link, is_tunnel] # Note: `is_link` is deprecated and will be removed in a future release in favor of the link subclass
```
danabauer, line 30: Is this deprecation notice still true?

danabauer: Ok, I see that `is_link` has been remodeled in the Pydantic schema.

danabauer (dismissed):

packages/overture-schema-transportation-theme/tests/test_segment_json_schema_baseline.py
sethfitz, line 1: Where did this move?

Thinking about the version dependencies in https://github.com/OvertureMaps/schema/pull/557#issuecomment-5119038218, these baseline tests are valuable to determine whether a type shift in an updated dependency breaks things unexpectedly (and the CI low/high versions will trigger this), in which case the likely fix will be updating the dependency version spec.

sethfitz (changes requested):

packages/overture-schema-transportation-theme/tests/test_segment_json_schema_baseline.py
vcschapp: This might've been an inadvertent delete intended to regenerate. Let me take a look.

vcschapp: Thx. Fixed and updated.

sethfitz (dismissed):

sethfitz (dismissed):

sethfitz (approved):

RoelBollens-TomTom (approved):

PR #603, opened 2026-07-27
[REFACTOR] Organize common package for consistency

### LINKED ISSUE #602: Refactor `overture-schema-common` package to make structure match other refactored Pydantic package
Scope of work:

1. ~~Drop support for `ext_*` from `OvertureFeature`. This is on a deprecation path and, since it is not supported in the Markdown and PySpark validation codegen anyway, this seems like a good time to drop it before GA (public PyPI publishing) of the schema packages.~~ **DEFERRED** because the `ext_*` stuff is being used in the reference counterexamples and I don't want to mess with that right now.
2. Drop `enums.py`, `models.py`, and `types.py` in favor of more specific filesnames.
3. Fix any `make docformat` errors in the common package.
4. Drop any unused obsolete code such as the `Theme` and `Type` types from `types.py`.

vcschapp:
Closes #602


sethfitz (dismissed):

RoelBollens-TomTom:
Like #598 (where it was fixed), this refactor also left a few broken references behind in `PYDANTIC_GUIDE.md` (broken imports at lines 57 and 181, and stale paths in comments at lines 1091 and 1134)

RoelBollens-TomTom (dismissed):

packages/overture-schema-common/src/overture/schema/common/__init__.py
```diff
@@ -1,15 +1,50 @@
-from . import cartography, names, scoping, sources
-from .models import OvertureFeature, ThemeT, TypeT
-from .scoping import Scope, scoped
+"""
+Shared building blocks common to all Overture theme packages.
+
+The `OvertureFeature` base class and the reusable feature components — naming, sources,
+cartographic hints, scoping, and more — that theme packages compose to define their
+feature types.
+
```
danabauer, line 7: Excellent! Thanks for adding this documentation

packages/overture-schema-common/src/overture/schema/common/confidence.py
```diff
@@ -0,0 +1,17 @@
+"""
+Confidence scores for machine-generated or machine-augmented data.
+"""
```
danabauer, line 3: I'd like a chance to update this explanation of confidence scores based on the place's team evolving definition of the property.

packages/overture-schema-common/src/overture/schema/common/perspectives.py
```diff
@@ -0,0 +1,40 @@
+"""
+Political perspectives.
+"""
```
danabauer, line 3: need to expand docstring

packages/overture-schema-common/src/overture/schema/common/confidence.py
vcschapp: Absolutely! Note that you can do that in a separate PR - it need not come as part of this re-organization PR.

I'll also note for the record that in terms of our published schema reference docs the *module* level docs (this) are less important than the *class* level docs.

packages/overture-schema-common/src/overture/schema/common/perspectives.py
vcschapp: Agree.

Same thoughts as I made [above](https://github.com/OvertureMaps/schema/pull/603#discussion_r3676770597):

- Can be a different PR.
- FWIW the module-level docs don't (I think) show up in the schema reference docs (@sethfitz correct me if I'm wrong).

sethfitz: Correct. They'd show up in mkdocs or whatever we end up using, but do not manifest in the generated Markdown.

danabauer: Good to know. Thanks, guys.

sethfitz (approved):

vcschapp:
> Like #598 (where it was fixed), this refactor also left a few broken references behind in `PYDANTIC_GUIDE.md` (broken imports at lines 57 and 181, and stale paths in comments at lines 1091 and 1134)

Thanks @RoelBollens-TomTom. I fixed the issue because it's better to keep the PR as accurate/tight as possible IMO. Could you re-up your approval if you don't mind?

RoelBollens-TomTom (approved):

PR #613, opened 2026-07-30
[CHORE](devops) Keep Dependabot uv updates in uv.lock

### LINKED ISSUE #612: [CHORE](devops) Keep Dependabot uv updates in uv.lock; refresh transitive deps via workflow
Dependabot's uv ecosystem currently updates both pyproject.toml and uv.lock, so routine dependency bumps rewrite the manifests. Dependency updates should flow through uv.lock only and leave every pyproject.toml untouched.

`versioning-strategy: lockfile-only` restricts Dependabot to uv.lock, but it covers only the direct deps declared in the manifests and does not refresh transitive-only entries (dependabot-core#14073).

Plan: set lockfile-only on the uv ecosystem, and add a scheduled workflow that runs `uv lock --upgrade` to refresh transitive deps into uv.lock, still with no manifest edits.

sethfitz:
Closes #612.

Restrict Dependabot's uv ecosystem to uv.lock and never edit pyproject.toml, then cover the transitive deps Dependabot leaves behind with a scheduled refresh.

## Changes

- `.github/dependabot.yml`: add `versioning-strategy: lockfile-only` to the uv ecosystem. Dependabot now updates uv.lock only, for the direct deps declared in the manifests.
- `.github/workflows/uv-lock-refresh.yml`: weekly `uv lock --upgrade` job that opens a PR refreshing transitive-only entries in uv.lock. lockfile-only does not touch these (dependabot/dependabot-core#14073). pyproject.toml files stay unchanged.

## Notes

- The refresh workflow's actions are pinned to full SHAs to match the repo convention, and its commit carries a Signed-off-by trailer for the DCO check (github-actions[bot] is not allowlisted the way dependabot[bot] is).
- PRs opened by the default `GITHUB_TOKEN` do not trigger other workflows, so CI will not gate the refresh PR. An in-file NOTE flags this; wiring a PAT/App token via `token:` is the fix if CI gating is wanted.
- lockfile-only opens no PR for a dependency capped in pyproject.toml whose only update exceeds the cap, since it refuses to widen the manifest. `uv lock --upgrade` respects the same caps. Both are inherent to the no-manifest-edits requirement.

### DIFF
--- a/.github/workflows/uv-lock-refresh.yml
+++ b/.github/workflows/uv-lock-refresh.yml
@@ -1,0 +1,54 @@
+name: Refresh uv.lock (transitive deps)
+
+# Dependabot's uv ecosystem runs lockfile-only (see .github/dependabot.yml) and
+# does not refresh transitive-only entries (dependabot-core#14073). This job does
+# that: `uv lock --upgrade` re-resolves every dependency to the newest version
+# each pyproject.toml already allows, touching only uv.lock.
+#
+# Remove this workflow once dependabot-core#14073 is fixed -- tracked in #614.
+
+on:
+  schedule:
+    - cron: "0 8 * * 4" # Thursdays -- offset from Dependabot's Monday run
+  workflow_dispatch: {}
+
+permissions: {}
+
+concurrency:
+  group: uv-lock-refresh
+  cancel-in-progress: false
+
+jobs:
+  lock:
+    name: Refresh uv.lock
+    runs-on: ubuntu-slim
+    permissions:
+      contents: write # create-pull-request pushes the refresh branch
+      pull-requests: write # create-pull-request opens/updates the PR
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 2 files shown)


.github/workflows/uv-lock-refresh.yml
```diff
@@ -0,0 +1,52 @@
...
+permissions: {}
+
+concurrency:
+  group: uv-lock-refresh
+  cancel-in-progress: false
+
+jobs:
+  lock:
+    name: Refresh uv.lock
+    runs-on: ubuntu-latest
```
lowlydba:
```suggestion
    runs-on: ubuntu-slim
```
nit: probably a good use case for the teeny tiny runner

lowlydba (dismissed):

.github/workflows/uv-lock-refresh.yml
sethfitz: Good call. Applied and re-pushed.

lowlydba (approved):

vcschapp (approved):

PR #619, opened 2026-07-31
[CHORE](annex) Place py.typed marker in the annex package, not the namespace root

### LINKED ISSUE #621: [CHORE] overture-schema-annex places py.typed at the namespace root, not the package
`overture-schema-annex` ships its PEP 561 `py.typed` marker at `src/overture/schema/py.typed` — the shared `overture.schema` namespace level — instead of `src/overture/schema/annex/py.typed`, where the package's code lives.

A `py.typed` marker applies to the directory it sits in, so at the namespace level it claims the shared namespace is typed rather than marking `overture.schema.annex`. Hatchling's `packages = ["src/overture"]` vacuum masks this today, but a module-root-scoped backend (uv_build) would drop the marker silently.

sethfitz:
overture-schema-annex shipped its PEP 561 marker at `src/overture/schema/py.typed` — the shared `overture.schema` namespace level — instead of `src/overture/schema/annex/py.typed`, where its code lives. A `py.typed` marker applies to the package directory it sits in; at the namespace level it claims the shared namespace is typed rather than marking `overture.schema.annex`.

Hatchling's `packages = ["src/overture"]` vacuum includes the whole tree, so the misplacement is masked today. It surfaces under any module-root-scoped build backend (uv_build), which would drop the marker silently — which is why this lands ahead of the uv_build migration.

Bottom of a three-PR stack migrating the workspace to uv_build / PEP 420 packaging.

Closes #621


vcschapp:
PR #629 should make this one moot.

vcschapp:
Closed as moot. PR #629 is now merged.

PR #620, opened 2026-07-31
[REFACTOR](schema) Extract overture-schema-validation package

### LINKED ISSUE #622: [REFACTOR] Extract the validation API out of the overture.schema namespace root
`validate()` and `validate_json()` live in the `overture-schema` umbrella's namespace `__init__` (`src/overture/schema/__init__.py`), so the shared `overture.schema` namespace root carries real code owned by a single distribution.

The aggregator role and the validation API should be separate, and the namespace root needs to be code-free before the uv_build migration — a distribution cannot own code at a shared namespace root under a module-root-scoped backend. Extract the validation API into its own `overture-schema-validation` package.

sethfitz:
`validate()` and `validate_json()` lived in the umbrella distribution's namespace `__init__` (`overture-schema/src/overture/schema/__init__.py`), so the shared `overture.schema` namespace root carried real code owned by one distribution. This moves them into a new `overture-schema-validation` package (`overture.schema.validation`); the umbrella now depends on it, and its namespace `__init__` becomes a bare pkgutil shim.

The immediate driver is packaging: freeing the namespace root of owned code is a precondition for the uv_build migration (PR 3), where a distribution cannot own code at a shared namespace root.

## Open design question

Whether a standalone `-validation` package is the right long-term home, or just the minimal way to vacate the root, is worth deciding here rather than assuming. The larger direction this enables — and the reason a dedicated package earns its place beyond "move code off the root":

- Pull validation-specific code out of `-common` and `-system` into `-validation`, so those packages carry model/type definitions and this one carries the validation API.
- Expose PySpark validation through a `[pyspark]` extra on this package rather than the separate `-pyspark` runtime distribution.

This PR does none of that consolidation — it takes only the packaging-forced step (vacate the root). The broader move is deliberately out of scope pending agreement on the target shape. Flagging it so the review decides the destination before more code accretes around the current split.

Targets `main` directly — the #619 layer this originally stacked on was closed when the annex package was removed (c492317b), so nothing precedes it now. Closes #622


atiannicelli (commented):
no longer stacked on #619" ?? Can you verify and update the description?

shouldn't test_schema_validation.py move to the new package too?

sethfitz:
Both addressed in 8d5a37a8.

Description fixed — #619 was closed when the annex package was removed (c492317b), so this stacks on nothing and targets `main` directly. Updated the body to say so.

Moved `test_schema_validation.py` into `overture-schema-validation/tests/`, and brought everything coupled to it along: its `pyyaml`/`yamlcore` dev-deps now sit on the new package, and the root `pythonpath` points at the new tests dir.

Two things surfaced while doing that move, both folded into the same commit:

- `deepdiff` was declared in the umbrella's dev group but is only used by `-common`'s tests — it resolved only because the workspace shares one venv. Moved it to `-common`.
- With the test and its deps gone, the umbrella is a bare namespace shim, so its leftover `dev` group and its now-vestigial runtime `pyyaml` and `pydantic` and direct `overture-schema-common` deps are removed (both still arrive transitively).

The CLI still carries its own copy of the union-adapter validation logic instead of importing `overture.schema.validation` — left alone, since that's the consolidation this PR explicitly defers.

lowlydba (approved):
Nice! Great work fixing those deps - having the main package *actually* be a true thin wrapper is a huge win for long term cleanliness

vcschapp (approved):

PR #623, opened 2026-07-31
[REFACTOR](build) Migrate workspace to uv_build and PEP 420 namespace packages

### LINKED ISSUE #618: [CHORE] Migrate workspace to uv_build with PEP 420 namespace packages
Migrate every workspace package from hatchling to the uv_build backend and convert the shared `overture.schema` namespace from pkgutil-style to PEP 420 implicit namespace packages.

uv_build is module-root-scoped: it never packages an `overture/__init__.py`, and a distribution cannot own code at a shared namespace root. The migration removes the `extend_path` pkgutil shims, deletes the namespace `__init__.py` files, stops the pyspark codegen from emitting `__init__.py` so the generated trees match, and guards the namespace roots against `__init__.py` regressing.

Depends on the annex `py.typed` placement and the validation-package extraction, which land first in the same stack.

sethfitz:
Three commits.

## `refactor(build): migrate all packages from hatchling to uv_build`

Switch every workspace package's build backend to uv_build. Because the packages share the `overture.schema` namespace, each declares an explicit `[tool.uv.build-backend] module-name` (e.g. `overture.schema.common`); name normalization would derive the wrong module. The `overture-schema` aggregator uses `module-name = "overture.schema"` with `namespace = true` and ships only `overture/schema/py.typed` — uv_build requires a module root, so a deps-only package still owns the namespace marker.

Convert the pkgutil namespaces to PEP 420: uv_build ships shared namespaces implicitly and never packages an `overture/__init__.py`, so the `extend_path` shims are removed. Wheels and editable installs both merge the namespace natively.

Deleting those `__init__.py` files removes the signal ruff's isort uses to detect `overture` as first-party (detect-same-package walks `__init__` chains), so add `src = ["packages/*/src"]`. With `overture` now first-party everywhere, imports across the test suite and a few pyspark modules recanonicalize into their own group; that reformat is folded in here because it cannot land as a separately-green commit.

## `refactor(build): emit PEP 420 generated pyspark tree`

Stop the pyspark codegen from emitting empty `__init__.py` files so the generated expression and test trees are PEP 420 namespace packages, matching the rest of the workspace. The mirrored layout (`generated/overture/schema/<theme>/`) is kept — only the init emission is dropped.

Removing the inits broke three things, each given a real fix:

- pytest collection: generated conformance tests reached `_support` via deep relative imports. Under PEP 420 the module resolves as `generated.*`, so those imports overshoot the top level. Emit absolute `from _support.X import ...`, delete the `_support_prefix` machinery, and move the hand-written tests to the same form. Set `--import-mode=importlib` and `consider_namespace_packages = true`.
- type checking: add the pyspark tests dir to `mypy_path`; keep the hand-written `tests/__init__.py` so pyspark's `test_cli` doesn't collide with codegen's under importlib.
- runtime registry: `pkgutil.walk_packages` skips directories without `__init__.py`, finding zero generated modules. Walk the namespace roots as files instead; `test_registry.py` covers this — no existing test exercised the on-disk walk, so an empty registry would have passed silently.

## `chore(build): guard PEP 420 namespace roots against __init__.py`

Add a `check-namespace` gate that fails if `overture/` or `overture/schema/` gains an `__init__.py`, which would turn a namespace root back into a regular package and shadow other workspace packages' contributions at import. Wired into the check gate so the PEP 420 invariant this stack establishes cannot silently regress.

Stacked on #620. Closes #618

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: default uv-sync clean-pyspark generate-pyspark check test-all test test-only docformat docformat-only doctest doctest-only mypy mypy-only lint-only update-baselines
+.PHONY: default uv-sync clean-pyspark generate-pyspark check check-namespace test-all test test-only docformat docformat-only doctest doctest-only mypy mypy-only lint-only update-baselines
 
 TESTMON ?= --testmon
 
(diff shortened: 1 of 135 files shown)


lowlydba (dismissed):
🔥

vcschapp (dismissed):

lowlydba (dismissed):

vcschapp (dismissed):

lowlydba (dismissed):

vcschapp (dismissed):

lowlydba (approved):

vcschapp (approved):

PR #626, opened 2026-07-31
[BUG](ci) Generate PySpark expressions before building the published wheel

### LINKED ISSUE #625: [BUG](pyspark) Published overture-schema-pyspark wheel ships without its generated validation expressions
## What

`overture-schema-pyspark`'s generated validation expressions (`overture/schema/pyspark/expressions/generated/`) are not present in git — they are regenerated on demand from the Pydantic models (`make generate-pyspark`). The publish workflow syncs the workspace and runs `uv build` with no codegen step in between, so `expressions/generated/` is absent on disk at build time and the published wheel and sdist ship without it. The build backend produces the empty wheel with no error.

At runtime `_registry.py` walks `overture.schema.pyspark.expressions.generated` at import; with the tree absent, `REGISTRY` is empty and `validate_model()` discovers no models. An installed consumer gets a package that imports cleanly and validates nothing.

## Root cause

The generated tree is gitignored with no corresponding build-time generation step. `publish-python-packages.yaml` runs `uv sync --all-packages` then `uv build --package "$PACKAGE"`; neither regenerates the tree, and `uv build` packages only what is on disk under the module root.

## Reproduction (clean checkout)

- `make clean-pyspark && uv build --package overture-schema-pyspark` → wheel contains 0 `expressions/generated/*.py`
- `make generate-pyspark && uv build --package overture-schema-pyspark` → wheel contains every generated model module

## Fix

Generate the expression tree in the publish workflow before `uv build` (only for `overture-schema-pyspark`), and add a post-build guard that aborts the publish if the wheel carries no generated expressions.

sethfitz:
Fixes #625.

## Problem

`overture-schema-pyspark` ships generated validation expressions that are not present in git; they are regenerated on demand from the Pydantic models. The publish workflow synced the workspace and ran `uv build` with no codegen step, so `expressions/generated/` was absent on disk at build time and the wheel and sdist shipped without it. `_registry.py` walks that namespace at import, so an installed consumer gets an empty `REGISTRY` and `validate_model()` discovers no models. The build produced the empty wheel with no error.

## Change

- Generate the expression tree (`make generate-pyspark`) in the publish workflow before `uv build`, gated to the `overture-schema-pyspark` matrix entry.
- Add a post-build guard: if the pyspark wheel contains no `expressions/generated/*.py`, abort the publish — a silent empty wheel becomes a loud failure.

## Verification

Built the wheel and sdist locally under uv_build:

| Path | generated in wheel | in sdist | guard |
|---|---|---|---|
| no generate step (current CI) | 0 | 0 | fires (aborts) |
| generate step (this PR) | all model modules | all | passes |

The guard checks the wheel; `uv build` builds the wheel from the sdist, so a populated wheel confirms a populated sdist.

### DIFF
--- a/.github/workflows/publish-python-packages.yaml
+++ b/.github/workflows/publish-python-packages.yaml
@@ -79,6 +79,16 @@ jobs:
         id: get-code-artifact-params
         uses: ./.github/actions/code-artifact
 
+      # overture-schema-pyspark ships generated validation expressions that are
+      # not committed to git -- they are regenerated on demand from the Pydantic
+      # models. `uv build` packages whatever is on disk under the module root, so
+      # the tree must be generated here before the build, or the published wheel
+      # ships without its `expressions/generated/` modules and validate_model()
+      # discovers nothing to run.
+      - name: Generate PySpark expressions before build
+        if: matrix.package == 'overture-schema-pyspark'
+        run: make generate-pyspark
+
       - name: Publish package ${{ matrix.package }} version ${{ matrix.after }} to PyPI
         env:
           CA_TOKEN: ${{ steps.get-code-artifact-params.outputs.token }}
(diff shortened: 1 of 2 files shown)


lowlydba:
suggestion (non-blocking): @sethfitz what about moving the codegen into the build itself instead of the workflow? This shifts us from making per-package exceptions in the monorepo to letting each package own its configuration needs.

The root cause is that hatchling packages whatever is under `src/overture` at build time, so wheel contents depend on what ran before `uv build`. A [hatchling build hook](https://hatch.pypa.io/latest/plugins/build-hook/custom/) makes the wheel correct by construction:

```toml
[tool.hatch.build.targets.wheel.hooks.custom]
path = "hatch_build.py"
```

with a hook that runs codegen if `expressions/generated/` is absent and fails the build if it comes up empty. Then any build path (this workflow, a local `uv build`, a consumer building from sdist) produces a complete wheel, and the guard lives inside the build where a new publish path can't skip it. The `unzip -l` check in YAML works, but it only protects this one workflow.

❗ Caveat: codegen imports the Pydantic models, so the hook needs `overture-schema-system` importable at build time. Either add it to `build-system.requires` or generate in the sdist hook only (CI has the workspace synced) and let wheels build from the populated sdist (feels trivial either way?)

vcschapp (dismissed):

vcschapp (approved):

RoelBollens-TomTom (approved):

Issue #637, opened 2026-08-05
Provision `overture-release-publisher` GitHub App for Phase 3 release publishing


lowlydba:
Manual prerequisite for Phase 3 (#509) `p3-release-publish`. Blocks the native `release: published` trigger from firing per-package: `release-trigger.yaml` currently creates releases with `GITHUB_TOKEN`, which by design does not fire downstream workflow runs.

## What to do

1. Create a new GitHub App named `overture-release-publisher` (org-owned, not personal).
2. Grant it repository permission `contents: write` only, nothing else.
3. Install it on `OvertureMaps/schema` (this repo only).
4. Add two repo secrets:
   - `RELEASE_PUBLISHER_APP_ID`: the app's client ID.
   - `RELEASE_PUBLISHER_APP_PEM`: the app's private key (PEM).

`release-trigger.yaml` will exchange these for a short-lived installation token via `actions/create-github-app-token`, scoped to `permission-contents: write`, and use it only for the `gh release create` step so each bumped package's release fires its own `release: published` event.

## Reference

Same pattern as the existing `overture-pull-requester` app used in [`rebase-vnext.yaml`](https://github.com/OvertureMaps/schema/blob/main/.github/workflows/rebase-vnext.yaml), which was deliberately not reused here to keep release-creation privileges (`contents: write` on the default branch's release/tag namespace) separate from that app's broader rebase/issue scope.

## Done when

- [x] App created and installed on `OvertureMaps/schema`
- [ ] `RELEASE_PUBLISHER_APP_ID` and `RELEASE_PUBLISHER_APP_PEM` secrets set on the repo
- [ ] A test release (or the first real Phase 3 release) confirms `p3-release-publish` fires on `release: published`

PR #557, opened 2026-07-14
[FEATURE] Branching strategy Phase 2.B - release trigger and versioning docs

lowlydba:
## Summary

Phase 2.B of the branching strategy (#533): release detection on `main`, the changelog machinery, and the versioning docs. No publish workflows, those are Phase 3 (#509), which this unblocks.

The versioning model settled during review (thanks @vcschapp, @sethfitz): humans own the full `<major>.<minor>.<patch>` per package, any bump merged to `main` is a release, and CI only versions the interim internal builds between releases.

### Release trigger
- `release-trigger.yaml` + `detect_version_bumps.py`: on push to `main`, diffs every `packages/*/pyproject.toml` version against the previous commit. Any increase (patch included) cuts one published GitHub Release per bumped package: tag `<pkg>-v<version>`, title `` `<pkg>` <version> ``, notes read from that package's `CHANGELOG.md`. Fails loudly on version decreases and on pre-existing tags. The umbrella `overture-schema` release is marked Latest.
- Major-bump cascade guard: a package whose direct workspace dependency takes a major bump must take one itself, since the old dependency floor would admit a breaking version. Enforced at PR time by the version check and again by `release-trigger`. Direct-dep checking covers longer chains, each unbumped link fails its own check.
- `create-package-release` composite action with a `dry-run` input for testing.

### Changelog machinery
- [towncrier](https://towncrier.readthedocs.io) fragments per package (`packages/<pkg>/changelog.d/`), config centralized in the root `pyproject.toml`.
- `require-changelog-fragment.yaml`: any PR touching a package must carry a fragment or a built `CHANGELOG.md`. Plain git diff + grep, no action dependencies.
- Release PRs run `towncrier build` so the notes are reviewed as part of the version-bump diff.

### Internal build versioning
- `compute-version` reworked: no-bump merges produce `<version>.postN+main.<sha>` (or `+vnext.<sha>`). PEP 440 post-releases order after the release, so `>=X.Y.Z` consumers resolve the freshest internal build while `==X.Y.Z` selects the clean release; local labels are rejected by public PyPI, keeping these builds internal by construction. `N` is a per-version sequence resolved from CodeArtifact: the first internal build of a version is `.post0`, later builds increment the highest published `.postN`.
- Verified against uv with a local two-wheel resolution test; the one spec gotcha (`>` never matches post-releases) is documented.
[...]

Issue #510, opened 2026-05-05
[Devops] Branching Strategy - Phase 4 - vnext dev builds to CodeArtifact


lowlydba:
## Goal

Publish `vnext` no-bump pushes to a dedicated dev CodeArtifact repository, the same way Phase 3's [`main-publish.yaml`](../.github/workflows/main-publish.yaml) does for `main`. This slot originally covered a docs-only editorial pass; that scope was folded into Phase 3 instead (docs got their update alongside the workflows they document, see PR #638), and the number is reused here since it went unbuilt.

Blocked on [ops-team#299](https://github.com/OvertureMaps/ops-team/issues/299), CloudOps' migration of CodeArtifact off the legacy Airflow environment onto Managed Core Data. That issue's proposed repo pattern includes `overture-dev-pypi` (writable by developers and some CI, upstream `overture-release-pypi`), which is the destination this phase needs. Until it exists, `vnext` pushes produce no internal build at all.

## Why this can't share `main`'s CodeArtifact repo

PEP 440 ignores local labels (`+main`/`+vnext`) in version ordering, so a `>=X.Y.Z` consumer in a shared repository could resolve a `vnext` post-build (breaking changes, mid-development) over a `main` one. The two streams must stay in physically separate repositories, not just differently-labeled versions in one, per the warning in [docs/versioning.md](../docs/versioning.md#version--destination).

### Dev builds -> CodeArtifact (vnext)
- Trigger: push to `vnext` with no version bump (same detection as `main`: affected packages whose directory changed without a `pyproject.toml` bump).
- Reuse the existing `detect-affected-packages` and `compute-version` (already takes a `context: vnext` input) composite actions; this is mostly a new trigger workflow wiring up infra Phase 3 already built, not new logic.
- Version format: `<version>.postN+vnext.<sha>`.
- Destination: `overture-dev-pypi` (or whatever ops-team#299 ultimately names it) once it exists.

## Definition of done

- [ ] Dev CodeArtifact repository exists and this repo has publish credentials to it (tracked upstream on [ops-team#299](https://github.com/OvertureMaps/ops-team/issues/299))
- [ ] `vnext` publish workflow implemented and live
- [ ] Resolver smoke test: a `>=` consumer against the dev repo picks up the freshest `vnext` `.postN` build, and the `main`/`vnext` repo isolation actually holds (a `main`-repo consumer never resolves a `vnext` build)
- [ ] `docs/versioning.md`'s "Blocked" note for this row updated once live

[...]

PR #638, opened 2026-08-05
[FEATURE] Branching strategy Phase 3 - PyPI and CodeArtifact publish workflows

### LINKED ISSUE #509: [Devops] Branching Strategy - Phase 3

lowlydba:
## Summary

Phase 3 of the branching strategy (#509): the publish side of the pipeline Phase 2.B (#557) computed versions for. Internal builds go to CodeArtifact on every push to `main`; human-cut releases go to public PyPI.

### Internal builds → CodeArtifact
`main-publish.yaml`, on push to `main`: `detect-affected-packages` (new composite action, using [`tj-actions/changed-files`](https://github.com/tj-actions/changed-files) pinned by commit SHA rather than a hand-rolled `git diff --name-only`, for correct filename quoting and its `dir_names`-to-package-name folding) diffs the push against `packages/**` to find packages touched without a version bump (bumped packages release via the path below instead). Each affected package gets a `.postN+main.<sha>` version stamped in via `compute-version`, built, and published to CodeArtifact.

The same workflow doubles as a build-only smoke test on PRs touching the composite actions or itself: it discovers every package unconditionally instead of diffing (a PR touching `.github/actions/**` never touches `packages/**`, so the diff would always return zero and skip the job it's meant to exercise), builds, and stops short of publishing. Replaces `compute-versions-dry-run.yaml`.

### Releases → PyPI
`release-publish.yaml`, on `release: published`: parses `<package>-v<version>` off the tag, builds, verifies the wheel matches (and, for `overture-schema-pyspark`, that the generated expressions are in it), and publishes via [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) with OIDC trusted publishing and attestations. The version-bump PR review is the approval gate; no separate environment reviewer step, one human approval per release is enough.

It also takes a `workflow_dispatch` (`package`, `target: test-pypi|pypi`) that always publishes a synthetic `<on-disk-version>.dev0`, never the real version. This exists to prime a package's PyPI Trusted Publisher from "pending" to "normal" ahead of its real release, working around PyPI's pending-publisher rate limit (~3 at a time, independently on pypi.org and test.pypi.org, see #653). Real-PyPI dispatch reuses the same `pypi-release-<package>` environment as the automated path; `workflow_dispatch` already requires repo write access to trigger, which is gate enough for a disposable `.dev0`. Temporary: #688 tracks removing the `pypi` target once every package is primed for the v2.0 launch.

`release-trigger.yaml` creates releases with an `overture-release-publisher` app installation token instead of `GITHUB_TOKEN`: GitHub doesn't fire `release: published` for releases created with the default token, so `release-publish.yaml` would never see them. Token scoped to `contents: write` only. App provisioning was tracked on #637 (now done).

A package's first version counts as a bump too: `detect_version_bumps.py` no longer skips `before: null` entries, so a brand-new package flows through `release-trigger` to PyPI on its first version, same as any other bump. `detect_affected_packages.py`'s `bumped` set was widened to match (any package with a non-null `after` version), so the same push doesn't also queue it for an internal `.postN` build.

<details>
<summary>Mechanical follow-through: diff-package-versions action, deleted workflows, docs</summary>

`package_versions.py diff` was inlined at three call sites (`detect-version-bumps`, the new `detect-affected-packages`, and the pre-existing `reusable-check-python-package-versions.yaml`), each reimplementing the temp-file-then-`$GITHUB_OUTPUT` plumbing slightly differently. Pulled it into its own `diff-package-versions` action exposing `count`/`diff` outputs; the three consumers now just pipe `diff` into their own filter logic instead of each knowing the script's path and output convention.

`package_versions.py` itself moved from `.github/workflows/scripts/` into `diff-package-versions/`, its only remaining caller, so every action in the repo now carries its own script(s) instead of reaching outside its directory.

Deletes `compute-versions-dry-run.yaml` and `publish-python-packages.yaml`, both superseded by the two workflows above.

`docs/versioning.md` gets a workflow-name reference table and the manual-dispatch priming mechanics under the release walkthrough. `CONTRIBUTING.md` drops the "rolling out in phases" banner (this is the last phase) and corrects two release-flow descriptions that predated the approval gate.

</details>

## Testing

`detect-affected-packages`'s exclusion logic is tested locally against real repo commit ranges plus synthetic bumped/removed/new-package cases, including through the `diff-package-versions` wiring and after the script relocation. The `uv version <post-release-string> --package <pkg> --frozen --dry-run` stamping technique is confirmed to accept PEP 440 local-label and dev-release strings. zizmor is clean on every touched/new workflow and action file. The workflows themselves can't run end to end until the external prerequisites below land.

## Still open (tracked on #653, all external/manual)

- PyPI Trusted Publisher config per package, batched to work around the pending-publisher rate limit
- `pypi-release-<package>`/`test-pypi-<package>` GitHub Environments
- a resolver smoke test against real CodeArtifact and a real end-to-end PyPI release, once the above are live

`vnext` dev builds to CodeArtifact moved out of this phase entirely, tracked on Phase 4 (#510) instead: it's blocked on a CodeArtifact repo migration ([ops-team#299](https://github.com/OvertureMaps/ops-team/issues/299)) with no committed timeline, so Phase 3 ships without waiting on it.

Closes #509
Contributes to #653, #688


.github/actions/detect-affected-packages/action.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: List changed package directories
+      id: changed-dirs
+      uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
+      with:
+        base_sha: ${{ inputs.before }}
+        sha: ${{ inputs.after }}
+        files: packages/**
+        dir_names: true
+        dir_names_max_depth: 2
+        matrix: true
```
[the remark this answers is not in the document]
lowlydba: Checked this against tj-actions/changed-files' `action.yml`: `matrix: true` is documented as *"Alias for setting inputs `json` to `true` and `escape_json` to `false`"*. So `json.loads` is already safe here without an explicit `json: true`. Leaving as-is.

.github/workflows/release-trigger.yaml
```diff
@@ -16,10 +16,14 @@ name: Publish GitHub release
...
 # .github/actions/compute-version).
 #
-# NOTE: releases are created with GITHUB_TOKEN, which by design does NOT trigger
-# further workflow runs. The Phase 3 publish workflow (#509) must therefore be
-# triggered by an app/PAT token here or via repository_dispatch, a plain
-# `on: release: published` listener will not fire for these releases.
+# Releases are created with the overture-releaser app's installation token,
+# not GITHUB_TOKEN: GITHUB_TOKEN-created releases don't fire downstream
+# `release: published` events, which release-publish.yaml needs to pick up
+# each release individually. See #637 for provisioning that app.
```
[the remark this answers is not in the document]
lowlydba: Fixed in `ee9e0f29`: `overture-releaser` → `overture-release-publisher`, matching #637.

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,118 @@
+name: Release publish
+
+# Triggered when a GitHub Release is published (created by release-trigger.yaml
+# using the overture-releaser app token -- see #637 -- so the native
+# `release: published` event actually fires; GITHUB_TOKEN would not).
+#
```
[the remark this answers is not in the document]
lowlydba: Fixed in `ee9e0f29`: `overture-releaser` → `overture-release-publisher`, matching #637.

.github/workflows/main-publish.yaml
```diff
@@ -0,0 +1,170 @@
...
+        run: |
+          set -euo pipefail
+          script="packages/${PACKAGE}/scripts/prebuild.sh"
+          if [ -f "$script" ]; then
+            echo "Running ${script}"
+            bash "$script"
+          fi
+
+      - name: Stamp computed version
+        # Rewrites only the in-memory checkout's pyproject.toml, never
```
vcschapp: nit: "in-memory checkout" is a bit of a funky term, since it's more "on-disk but temporary" rather than "in-memory".

Suggest:

> Only updates `pyproject.toml` in the action's temporary checkout, never committed.

.github/workflows/main-publish.yaml
```diff
@@ -0,0 +1,170 @@
...
+          persist-credentials: false
+
+      - name: Sync code to make packages visible to Python
+        run: uv sync --locked --all-packages
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
+        with:
+          aws-region: us-west-2
+          role-to-assume: arn:aws:iam::505071440022:role/GithubActions_Schema_CodeArtifact_Publish
```
vcschapp: This selects a pretty high-privilege role regardless of whether we're on the "real publish" (`push`) or "smoke test" (`pull_request`) branch. Possible to downscope privileges for the `pull_request` branch?

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,149 @@
...
+          EVENT_NAME: ${{ github.event_name }}
+          TAG: ${{ github.event.release.tag_name }}
+          INPUT_PACKAGE: ${{ inputs.package }}
+        run: |
+          set -euo pipefail
+          if [[ "$EVENT_NAME" == "workflow_dispatch" ]]; then
+            if [[ ! -d "packages/${INPUT_PACKAGE}" ]]; then
+              echo "::error::No such package: packages/${INPUT_PACKAGE}"
+              exit 1
+            fi
```
vcschapp: Suggest moving this test outside the outer `if` so it runs unconditionally.

Reason: currently it sanity-checks the user input for `workflow_dispatch` but there's no downside and only upside to extending the sanity check to the tag parsing. In both cases, validating that the package directory exists increases our confidence in the automation.

---

I realize we didn't check out the code for the `published` arm, but then looking closer there seem to be duplicative checkouts for the `parse` and `publish` jobs on the `workflow_dispatch` arm.

Maybe it would be simpler to merge `parse` and `publish` into one thingy and have an unconditional checkout done once before the parsing that does HEAD or whatever for `workflow_dispatch` arm and the tag for `published` arm—this gets us down to single checkout for both arms—and then we could apply the "does the package really exist?" logic unconditionally to both arms...

.github/actions/detect-affected-packages/detect_affected_packages.py
```diff
@@ -0,0 +1,64 @@
...
+import sys
+
+
+def main() -> None:
+    version_diff = json.load(sys.stdin)
+    bumped = {change["package"] for change in version_diff if change["after"] is not None}
+    removed = {change["package"] for change in version_diff if change["after"] is None}
+
+    changed_paths = json.loads(os.environ.get("CHANGED_DIRS") or "[]")
+    changed = {path.split("/", 1)[1] for path in changed_paths}
```
vcschapp: This is a bit subtle but if for some reason `CHANGED_DIRS` had a path whose split on `"/"` yielded fewer than two entries, this would give a really confusing `IndexError`. Suggest we handle the possibility with an explicit exception message that will be intelligible.

.github/actions/detect-affected-packages/detect_affected_packages.py
```diff
@@ -0,0 +1,64 @@
...
+changed without also changing their pyproject.toml version. Those need an
+internal `.postN` build (see docs/versioning.md). Packages with a version bump
+in the same range are excluded because they release via `release-trigger`
+instead, and removed packages are excluded because there is nothing left to
+build.
+
+Run from the repository root, piping `package_versions.py diff`'s output in:
+
+    python3 package_versions.py diff BEFORE AFTER \\
+      | CHANGED_DIRS='["packages/overture-schema-common"]' python3 detect_affected_packages.py
```
vcschapp: It feels weird to me that we're using both stdin and an environment variable for input.

Can we just send it all to stdin doing something more like this:

```bash
$ jq -n \
      --argjson version_diff "$(python3 package_versions.py diff BEFORE AFTER)" \
      --argjson changed_dirs '["packages/overture-schema-common"]' \
      '{version_diff: $version_diff, changed_dirs: $changed_dirs}' \
    | python3 detect_affected_packages.py
```

.github/workflows/main-publish.yaml
```diff
@@ -0,0 +1,170 @@
...
+    name: Detect affected packages${{ github.event_name == 'pull_request' && ' (test)' || '' }}
+    if: github.event.repository.full_name == github.repository
+    runs-on: ubuntu-slim
+    permissions:
+      contents: read # Read pyproject.toml/file history to detect affected packages
+    outputs:
+      count: ${{ steps.discover.outputs.count || steps.detect.outputs.count }}
+      packages: ${{ steps.discover.outputs.packages || steps.detect.outputs.packages }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
vcschapp: nit: my LLM buddy tells me this commit hash is inconsistent with the one used in `release-publish.yaml`.

vcschapp (commented):
Raised a few minor nittish review comments. Nothing big. I'm excited to merge this, once we do it seems we're in the business of publishing Python packages!

.github/workflows/main-publish.yaml
lowlydba: Fixed in `0c520087`: "in-memory" -> "temporary checkout".

.github/workflows/main-publish.yaml
lowlydba: Good catch, and there's already a read-only role for this (`GithubActions_Schema_CodeArtifact_ReadOnly`, used in `reusable-check-python-package-versions.yaml`). Fixed in `0c520087`: the PR smoke test now assumes that instead of the publish role, so a wiring mistake in the skipped publish step can't turn into a real write.

.github/workflows/release-publish.yaml
lowlydba: Applied the sanity-check part in `0c520087`: `parse` now checks out unconditionally (same tag/sha ref selection `publish` already uses) and validates `packages/<pkg>` exists for both arms before setting outputs, not just `workflow_dispatch`.

Held off on merging `parse`/`publish` into one job. That's a bigger restructure than this feedback pass, and it'd fold the `ubuntu-slim` parse into `ubuntu-latest` for a check that doesn't need it. Open to doing it as a follow-up if the duplication bugs you more than the runner-size split.

.github/actions/detect-affected-packages/detect_affected_packages.py
lowlydba: Fixed in `0c520087`: pulled the split into a `_package_name` helper that raises a clear `ValueError` (path, expected shape) instead of letting a short split hit an `IndexError`.

.github/actions/detect-affected-packages/detect_affected_packages.py
lowlydba: Agreed, fixed in `0c520087`: `action.yml` now builds one `{version_diff, changed_dirs}` object with `jq -n --argjson ... --argjson ...` and pipes that to stdin, matching your suggestion. `detect_affected_packages.py` reads both keys off the single payload; no more `os.environ` read.

.github/workflows/main-publish.yaml
lowlydba: Your LLM buddy's right, `main-publish.yaml` and `release-publish.yaml` had drifted to a mix of `actions/checkout` v7.0.0/v7.0.1 across this PR, while the rest of the repo is on v7.0.1. Standardized on v7.0.1 everywhere in `0c520087`.

vcschapp (dismissed):

sethfitz (dismissed):

lowlydba:
@copilot resolve the merge conflicts in this pull request

sethfitz (approved):

vcschapp (approved):

vcschapp:
@copilot resolve the merge conflicts in this pull request.

**Edit**: Copilot can't hear me.

PR #662, opened 2026-08-14
[BUG] Fix validation registry discovery when loaded via zipimport

### LINKED ISSUE #661: Validation registry empty when overture-schema-pyspark loaded via zipimport
### Type

Bug

### Scope

Multi-theme or Platform

### Skillset

engineering

### Description

**Acceptance criteria**:

Rachmanin0xFF:
Fixes #661

> [!NOTE]
> This is one of **two alternative fixes** for #661. The other, #663, replaces filesystem discovery with a codegen-emitted index module the registry imports. Merging either one resolves the issue.

For reference, **this** PR has the fix that was temporarily published to CA to unblock release [here](https://github.com/OvertureMaps/tf-data-platform/pull/4818).

## What

`overture.schema.pyspark._registry.REGISTRY` comes back empty whenever `overture-schema-pyspark` is loaded straight from a wheel on `sys.path` instead of being installed to a real directory (this is what AWS Glue does when using the `--extra-py-files` option like we do in data-platform). Every feature type then silently reports as unregistered and the task erroneously succeeds 🫠 

## Why

`_registry.py` discovered generated modules by walking the `expressions/generated/` PEP 420 namespace package with `pathlib.Path(root_path).rglob("*.py")`, which works... only when a real filesystem directory is present! A namespace portion backed by a zip archive has a `__path__` entry that points *inside* the zip, and `pathlib` can't traverse into one.

`importlib.resources.files()` looked like the natural fix (its `Traversable` API is meant to be zipimport-aware), but its `MultiplexedPath` implementation raises `NotADirectoryError` on Python 3.10 the moment any namespace portion isn't a real directory.

## Fix

Each namespace portion is now walked directly using `zipfile.ZipFile.namelist()` for portions that resolve to a path inside a zip file.

## Verification

- Built a wheel and added it to `sys.path` without extracting it (no `pip install`, matching Glue's `--extra-py-files`): all 15 generated modules now resolve; previously `len(REGISTRY) == 0`.
- Documented empty-registry behavior when `expressions/generated/` is absent entirely (`make clean-pyspark`): unchanged, still empty and non-crashing.

### DIFF
--- a/packages/overture-schema-pyspark/changelog.d/661.bugfix.md
+++ b/packages/overture-schema-pyspark/changelog.d/661.bugfix.md
@@ -1,0 +1,1 @@
+Fixed the validation registry coming back empty when `overture-schema-pyspark` is loaded from a wheel on `sys.path` (zipimport) rather than installed to a real directory, as happens on AWS Glue via `--extra-py-files`. Generated modules under `expressions/generated/` are now discovered whether the package lives on disk or inside a zip archive.
--- a/packages/overture-schema-pyspark/src/overture/schema/pyspark/_registry.py
+++ b/packages/overture-schema-pyspark/src/overture/schema/pyspark/_registry.py
@@ -15,7 +15,8 @@ from __future__ import annotations
 
 import importlib
 import logging
-from pathlib import Path
+import zipfile
+from pathlib import Path, PurePosixPath
 
 from .check import ModelValidation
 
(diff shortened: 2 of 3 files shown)


vcschapp:

packages/overture-schema-pyspark/changelog.d/661.bugfix.md
```diff
@@ -0,0 +1 @@
+Fixed the validation registry coming back empty when `overture-schema-pyspark` is loaded from a wheel on `sys.path` (zipimport) rather than installed to a real directory, as happens on AWS Glue via `--extra-py-files`. Generated modules under `expressions/generated/` are now discovered whether the package lives on disk or inside a zip archive.
```
vcschapp, line 1: uber-nit: `s/zip archive/ZIP archive/`

vcschapp (changes requested):
Gave my reasons in issue #661 why I think PR #666 is our best bet, [see my comment here](https://github.com/OvertureMaps/schema/issues/661).

Rachmanin0xFF:
Superseded by https://github.com/OvertureMaps/schema/pull/666

PR #663, opened 2026-08-14
[BUG] Discover validation registry via a generated index module

### LINKED ISSUE #661: Validation registry empty when overture-schema-pyspark loaded via zipimport

Rachmanin0xFF:
Fixes #661

> [!NOTE]
> This is one of **two alternative fixes** for #661. The other, #662, keeps discovery on the filesystem and teaches the walk to read inside a zip. Merging either one resolves the issue.

## What

Same bug as #662: `REGISTRY` comes back empty on AWS Glue, so every feature type silently reports as unregistered and the task erroneously succeeds 🫠. This happens when `overture-schema-pyspark` is loaded as a raw `.whl` on `sys.path` (Glue's `--extra-py-files`) instead of installed to a real directory.

## Why

`_registry.py` discovered generated modules by walking `expressions/generated/` with `pathlib.Path(root_path).rglob("*.py")`, which works... only when a real filesystem directory exists! When the package is a raw wheel on `sys.path`, those modules live *inside* the zip, and `pathlib` can't traverse into one.

## Fix

We stop discovering by filesystem. `make generate-pyspark` now emits an `_index.py` alongside the validation modules that imports each one and exposes them as `MODULES`, and the registry imports that index. An ordinary import resolves whether the package is a directory or a wheel on `sys.path`. The generated modules were always *importable* under zipimport; only enumerating them by walking the tree was broken.

The index is generated from the same specs as the modules. The registry-building policy stays in `_registry.py`; codegen only decides which modules exist. There's also a test asserts that the index covers exactly the modules on disk, guarding the codegen step.

## Approach vs #662

- **#662** keeps automatic filesystem discovery and adds `pathlib`-plus-`zipfile` walking that has to know every way the package might be loaded.
- **This PR** moves discovery to an ordinary import of a generated index. The runtime shrinks to "import the index, read `MODULES`", zipimport-safe for free, at the cost of a small addition to the codegen (one extra emitted module).

An earlier revision of this PR used a hand-maintained entry-point table; that was replaced with the codegen-emitted index so nothing has to be kept in sync by hand.

## Verification

- Built a wheel, added it to `sys.path` without extracting it (matching Glue's `--extra-py-files`): all 15 generated modules resolve; previously `len(REGISTRY) == 0`. Confirmed `_index.py` ships in the wheel.
- `make check` passes (mypy, lint, docformat, full suite across the codegen and pyspark packages).
- A present-but-broken module raises instead of being silently skipped; an absent generated tree yields a clean empty registry.

Yet-untested on Glue!


vcschapp:

vcschapp (changes requested):
Gave my reasons in issue #661 why I think PR #666 is our best bet, [see my comment here](https://github.com/OvertureMaps/schema/issues/661).

Rachmanin0xFF:
Superseded by #666

PR #666, opened 2026-08-17
[BUG] Discover validation registry through importlib.resources

### LINKED ISSUE #661: Validation registry empty when overture-schema-pyspark loaded via zipimport

sethfitz:
Fixes #661

> [!NOTE]
> A third alternative for #661, alongside #662 (teach the filesystem walk to read inside a zip) and #663 (replace discovery with a codegen-emitted index). Merging any one resolves the issue.

## What

`overture.schema.pyspark._registry.REGISTRY` comes back empty when `overture-schema-pyspark` is loaded from a wheel placed on `sys.path` rather than installed to a real directory, which is what AWS Glue does with `--extra-py-files`. Every feature type then reports as unregistered.

## Why

`_registry` walked the generated tree with `pathlib.Path.rglob`. Under zipimport the namespace portion's `__path__` entry points *inside* the archive, `pathlib` cannot traverse there, and `rglob` returns no matches without raising — so the empty registry is indistinguishable from the documented "generated tree absent" case.

## Fix

Read the tree through `importlib.resources.files`, which resolves a namespace portion whether that portion is a directory on disk or a member of an archive. This is the API the stdlib provides for exactly this problem, so discovery no longer has to know how the package was loaded: any loader exposing a resource reader works, rather than only the ones the walk was taught about.

`files()` raises `NotADirectoryError` on a namespace package with any non-directory portion through Python 3.12, which is why it isn't usable as-is (as #661 notes). CPython fixed it in 3.13 and the `importlib_resources` backport carries the same fix from 6.2, so the backport is a dependency below 3.13 and evaporates when the floor rises. Glue 4.0 runs Python 3.10 and Glue 5.0 runs Python 3.11, so on Glue the backport is always the code path taken.

Two things this preserves that are worth naming. `files()` multiplexes *every* portion of the namespace, so a generated tree assembled from more than one installed distribution is still walked whole — the plural-`__path__` behaviour the previous implementation had. And a missing `expressions/generated/` subtree still yields an empty registry rather than raising.

Discovery now takes the root package name instead of a list of portion paths. That lets `_walk` drop its own `import_module` of the root (`files()` performs it) and lets the tests point the walk at a synthetic package on `sys.path`.

## Verification

End to end, against a real wheel that is never unpacked — the deployment shape, not a simulation of it:

```
$ PYTHONPATH=/tmp/wheeltest/overture_schema_pyspark-0.1.1-py3-none-any.whl python -c "..."
registry module loaded from: .../overture_schema_pyspark-0.1.1-py3-none-any.whl/overture/schema/pyspark/_registry.py
registry entries: 15
```

Control arm, same harness and same wheel shape built from `main`: `registry entries (UNFIXED code): 0`. The check distinguishes fixed from unfixed rather than merely reporting a number.

Also confirmed the wheel carries the PEP 420 generated tree (15 modules, zero `__init__.py`), and `make check` passes — full suite, mypy, lint, docformat.

## On the tests

Three tests are added; **one of them is the gate**. `test_iter_generated_module_names_reads_a_zipimported_tree` builds a wheel-shaped zip, puts it on `sys.path`, and walks the resulting namespace portion — so the portion string under test comes from real import machinery rather than being hand-written, which is the coupling the bug turned on. Verified by isolation: with the new signature but a `pathlib` walk behind it, that test fails and the other two pass.

The other two — a directory-backed tree and an absent tree — pass under both implementations. They are behaviour-preservation guards for the rewrite, not proofs of the fix.

The zip fixture writes explicit directory entries, because `zipimport` recognises a namespace portion inside an archive only when the archive carries one, and that is what `uv_build` emits. Omitting them would test a shape no real wheel has.

## Not addressed

The issue reports that the task "erroneously succeeds". `overture-validate` already exits 1 on an unresolvable feature type, so the silent success is on the consumer side — a caller treating `feature_type in REGISTRY` as "is this one of ours". Worth a separate issue against data-platform. Switching those jobs from `--extra-py-files` to `--additional-python-modules` would also sidestep zipimport entirely and let pip resolve `click` / `overture-schema-system` instead of shipping them by hand.

### DIFF
--- a/packages/overture-schema-pyspark/src/overture/schema/pyspark/_registry.py
+++ b/packages/overture-schema-pyspark/src/overture/schema/pyspark/_registry.py
@@ -4,18 +4,36 @@ Built at import time by walking the generated `expressions.generated`
 namespace and collecting every module that exposes the
 codegen-emitted `ENTRY_POINT` and `MODEL_VALIDATION` constants.
 
-The generated tree on disk is the runtime source of truth: the
-registry contains exactly what was generated, regardless of which
-theme packages are installed alongside the pyspark package. A missing
+The generated tree is the runtime source of truth: the registry
+contains exactly what was generated, regardless of which theme
+packages are installed alongside the pyspark package. A missing
 `expressions/generated/` subtree simply yields an empty registry --
 the package still imports cleanly.
+
+The tree is read through `importlib.resources`, which resolves a
+namespace portion whether it is a directory on disk or a member of an
+archive. A wheel placed straight on `sys.path` is zipimported rather
+than installed -- AWS Glue does this with `--extra-py-files` -- and
+`pathlib` cannot traverse into one.
 """
 
 from __future__ import annotations
 
 import importlib
 import logging
-from pathlib import Path
+import sys
+from collections.abc import Iterator
+
+if sys.version_info >= (3, 13):
+    from importlib.resources import files
+    from importlib.resources.abc import Traversable
+else:
+    # `importlib.resources.files` raises `NotADirectoryError` for a namespace
+    # package with any non-directory portion through Python 3.12; the backport
+    # carries the 3.13 fix. AWS Glue 4.0 runs Python 3.10 and Glue 5.0 runs
+    # 3.11, so on Glue this is always the branch taken.
+    from importlib_resources import files
+    from importlib_resources.abc import Traversable
 
 from .check import ModelValidation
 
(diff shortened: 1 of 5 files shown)


packages/overture-schema-pyspark/tests/test_registry.py
```diff
@@ -46,3 +112,38 @@ def test_registry_discovers_generated_models() -> None:
         key for key in REGISTRY if ":" in key and key.startswith("overture.schema.")
     ]
     assert generated_entries, "registry found no generated feature modules on disk"
+
+
+def test_iter_generated_module_names_reads_a_zipimported_tree(tmp_path: Path) -> None:
+    """Modules inside a wheel on `sys.path` are discovered.
+
+    Mirrors Glue's `--extra-py-files`: the wheel is never unpacked, so the
```
sethfitz: Does `--extra-py-files` belong to Glue or to `pyspark`?

packages/overture-schema-pyspark/src/overture/schema/pyspark/_registry.py
```diff
@@ -4,18 +4,36 @@
...
+The generated tree is the runtime source of truth: the registry
+contains exactly what was generated, regardless of which theme
+packages are installed alongside the pyspark package. A missing
 `expressions/generated/` subtree simply yields an empty registry --
 the package still imports cleanly.
+
+The tree is read through `importlib.resources`, which resolves a
+namespace portion whether it is a directory on disk or a member of an
+archive. A wheel placed straight on `sys.path` is zipimported rather
+than installed -- AWS Glue does this with `--extra-py-files` -- and
```
sethfitz: Same as below. Is this Glue-specific?

packages/overture-schema-pyspark/tests/test_registry.py
```diff
@@ -5,19 +5,85 @@
 real on-disk walk -- conformance tests import expression modules directly
 and `test_validate.py` registers models through a test shim -- so an empty
 registry would otherwise pass the suite unnoticed.
+
+The walk has to work for a namespace portion inside a zip as well as one on
+disk, because a wheel placed straight on `sys.path` is zipimported rather
+than installed (AWS Glue's `--extra-py-files`). Both shapes are exercised
```
sethfitz: Glue-specific?

vcschapp:

vcschapp (approved):
Gave my reasons in issue #661 why I think this one is our best bet, [see my comment here](https://github.com/OvertureMaps/schema/issues/661).

Rachmanin0xFF (approved):

Issue #682, opened 2026-08-19
[CHORE](ci) Python under .github/ escapes every lint and type gate


sethfitz:
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

dev ops

### Description

The three Python files under `.github/` are covered by no gate:

```
.github/workflows/scripts/package_versions.py
.github/actions/detect-version-bumps/detect_version_bumps.py
.github/actions/create-package-release/extract_release_notes.py
```

`[tool.ruff] src`, mypy's `files`, and the pytest paths in the `Makefile` all point at `packages/`. Nothing has ever applied the repo's own configuration to these files.

That is not hypothetical. Running the repo's ruff config over them today reports two lint errors and two files needing reformatting:

```bash
find .github -name '*.py' -print0 | xargs -0 uv run ruff check
find .github -name '*.py' -print0 | xargs -0 uv run ruff format --check
```

These files are release automation — they compute version bumps, extract release notes, and read package versions. They are not throwaway: `detect_version_bumps.py` and `package_versions.py` feed the publish path, so a mistake in them is a mistake in a release.

### Why this is being filed now

It surfaced while justifying a new top-level `tests/` directory in the #604 / #668 documentation PR. `.github/` is the repo's existing home for repo-wide process code, so it was the natural prior art to compare against — and the comparison showed that confining process code there also places it outside every check. Flagged in that PR's description for discussion; not fixed there, as it is unrelated to the documentation work.

### Suggested approach

- Add the paths to `lint-only` (and its `format` counterpart) and to mypy's `files`.
- Check whether `[tool.ruff] src` needs entries so first-party imports classify correctly in those files.
- Fix whatever the two lint errors turn out to require — they looked like import ordering, but confirm rather than assume.
- Decide whether these scripts warrant tests. They have none, and a release-path bug is expensive; that is a larger question than wiring up the linters and can be split out.

### Acceptance

`find .github -name '*.py' -print0 | xargs -0 uv run ruff check` and `ruff format --check` both pass, and `make check` fails if a future edit breaks them.

lowlydba:
These are likely to be hoisted to generic-ized versions in our public `workflows` repo - so I am intentionally not being precious with them now. Likely many will be re-written in something other than Python (either simple bash, or using TS due to the robust js libs GitHub has available).

Generally, I don't like to apply codebase-specific lint rules to CI though - what is not acceptable in the schema packages likely varies greatly compared to the concerns of CI/Actions.

Issue #491, opened 2026-04-08
Test Python code examples in documentation


sethfitz:
## Problem

Python code blocks in READMEs and guides contain import paths, type references, and usage patterns that go stale when the codebase changes. Nothing catches this today. Errors survive until a human reads the rendered documentation and tries to follow along.

#489 is a recent example. The core/system package split left behind incorrect import paths (`core.types` instead of `system.string`, `primitives` instead of `primitive`, `model_constraints` instead of `system.model_constraint`) and stale file references (`core/ref.py` instead of `system/ref/ref.py`). These were found and fixed manually as part of a broader documentation alignment — not because a test caught them.

The repo has ~56 Python code blocks across 5 files (`PYDANTIC_GUIDE.md` alone has 47). As the schema evolves, these will drift again.

## Scope

Files with Python code examples today:

| File | Code blocks | Notes |
|------|-------------|-------|
| `PYDANTIC_GUIDE.md` | 47 | Contributor-facing guide; mix of self-contained and sequential examples |
| `packages/overture-schema/README.md` | 3 | Top-level package usage |
| `packages/overture-schema-core/README.md` | 3 | Primitive types, scoping |
| `README.pydantic.md` | 2 | Package overview |
| `packages/overture-schema-codegen/README.md` | 1 | Extraction API |

Most blocks are self-contained (each includes its own imports). Some build sequentially (defining a model in one block, using it in the next). Some reference undefined variables like `feature_data` that would need hidden setup to execute.

## Options

Three pytest-based tools are worth considering. All detect ` ```python ` fences automatically and integrate with the existing test suite.

### 1. [Sybil](https://sybil.readthedocs.io/) (v10)

All code blocks in a file share a namespace by default — sequential examples just work without annotation. Hidden setup blocks (`<!-- invisible-code-block: python -->`) handle undefined variables like `feature_data`. Skip (`<!-- skip: next -->`), namespace reset (`<!-- clear-namespace -->`), and full pytest fixture support are available. Configuration is ~5 lines in `conftest.py`.

*Trade-offs:* The shared-namespace default means a failing block can cascade failures through the rest of the file. The HTML comment directives (skip, invisible blocks) are invisible when rendered but add noise to the raw markdown source. More configuration surface than the other options.

### 2. [pytest-markdown-docs](https://github.com/modal-labs/pytest-markdown-docs) (v0.9)

[...]

PR #683, opened 2026-08-19
[BUG] Correct the Python package documentation

### LINKED ISSUE #604: [DOCS] PYDANTIC_GUIDE.md documents a package layout the repo no longer uses
`PYDANTIC_GUIDE.md` documents a package layout the repo no longer uses. Raised by @RoelBollens-TomTom while reviewing #598 (https://github.com/OvertureMaps/schema/pull/598#issuecomment-5095474393); filing separately because the staleness predates that PR and fixing it there would be out of scope.

## Entry point example is wrong on both axes

Lines 850-856 show:

```toml
"transportation.connector" = "overture.schema.transportation.connector.models:Connector"
"transportation.segment" = "overture.schema.transportation.segment.models:Segment"
```

Real declaration in `packages/overture-schema-transportation-theme/pyproject.toml`:

```toml
connector = "overture.schema.transportation:Connector"
segment = "overture.schema.transportation:Segment"
```

Keys are unqualified, not theme-prefixed, and targets resolve off the theme package root rather than a `<type>.models` submodule.

## The `models.py` / type-subdirectory convention is gone

Lines 1006-1018 prescribe a type subdirectory per feature type with `models.py` / `enums.py` / `types.py` inside it. `overture-schema-buildings-theme` already contradicts this -- flat `building.py`, `building_part.py`, `_common.py` directly under `overture/schema/buildings/`, no subdirectory and no `models.py`. #598 does the same for transportation (`connector.py`, and `segment/{_common,road,rail,water}.py`), which removes the last `models.py` in a theme package; the file name survives only in `overture-schema-common` and `overture-schema-annex`.

## Dead path references

- L857 links `packages/overture-schema-transportation-theme/src/overture/schema/transportation/segment/models.py` -- resolves today, breaks when #598 merges.
- L1098 comments `# In overture-schema-buildings-theme/src/overture/schema/buildings/building/models.py` -- already dead.
- L1139 comments `# In buildings/models.py` -- already dead.

## Not stale

Every `from overture.schema.*` import in the guide resolves on `main` (11 distinct modules, all import cleanly), including the recent `system.primitive` split into `system.numeric` / `system.geometric`. The rot is confined to file layout and entry point naming, not the import surface.

## Scope

Rewrite the "Module Organization" and entry point registration sections against the buildings / post-#598 transportation layout, and repoint the three path references. Best done after #598 lands so the transportation examples describe the settled structure.

### LINKED ISSUE #668: Python package documentation is inaccurate in several places
### Type

Bug

### Scope

Multi-theme or Platform

### Skillset

engineering

### Several issues
I worked through the Python package docs end to end found several inaccuracies that range from "will not work at all" to cosmetic. Many of these are fixed in my

1. Every `pip install` instruction fails because the packages aren't published yet
From package READMEs:
```
packages/overture-schema-cli/README.md
packages/overture-schema-common/README.md
packages/overture-schema-system/README.md
packages/overture-schema-validation/README.md
packages/overture-schema/README.md
```
Until these are published, we need to update the instructions to the local `uv` workflow: `uv sync --all-packages` or `[tool.uv.sources]` path entry. 

2. Top-level imports from `overture-schema` are broken
```
from overture.schema import Building, Place
from overture.schema import parse, discover_models, json_schema
```
Fails:
```
BROKEN   from overture.schema import Building        -> ImportError
BROKEN   from overture.schema import parse           -> ImportError
BROKEN   from overture.schema import discover_models -> ImportError
BROKEN   from overture.schema import json_schema     -> ImportError
```
`overture-schema` because a metapackage whose namespace root ships only `py.typed`. Correct paths:

Documented | Actual
-- | --
overture.schema:Building | overture.schema.buildings:Building
overture.schema:discover_models | overture.schema.system.discovery:discover_models
overture.schema:json_schema | overture.schema.system.json_schema:json_schema
overture.schema:parse | does not exist anywhere in the codebase

`(parse)` no longer exists?

3. `model_validate()` is documented as accepting GeoJSON but it does not. From `overture-schema` README:
```
# Validate Overture Maps data (supports both flat/tabular and GeoJSON formats)
building = Building.model_validate(feature_data)
place = Place.model_validate(geojson_feature)
```
`model_validate` is Python mode, which expects the flat shape. Passing a GeoJSON dict fails:

```
ValidationError: 3 validation errors for building
  theme     Field required
  type      Input should be 'building'  [input_value='Feature']
  version   Field required
```

GeoJSON requires JSON mode — `model_validate_json()`, which accepts the identical payload. The error message ("theme required") does not point at the real cause.

4. The plain `overture` tag does not exist.

The CLI's own `--help` suggests a filter that matches nothing:

```
packages/overture-schema-cli/src/overture/schema/cli/commands.py:761
packages/overture-schema-cli/src/overture/schema/cli/commands.py:824
    $ overture-schema validate --tag overture --tag feature data.json
```

`overture-schema list-types --tag overture     # 0 types`

Discovery emits seven tags in total: `feature`, plus one `overture:theme=*` per theme. Either the tag should be emitted by a provider, or the help strings should stop referring to it.

5. overture-schema-codegen/README.md documents a removed API.

The "Programmatic use" section describes `analyze_type()` returning a `TypeInfo` with `.kind`, `.base_type`, and .`newtype_name`, and imports `TypeKind`:
```
from overture.schema.codegen.extraction.type_analyzer import analyze_type, TypeKind
info = analyze_type(some_annotation)
assert info.kind == TypeKind.PRIMITIVE
```
Neither `TypeInfo` nor `TypeKind` exists. The actual signature returns a 3-tuple:

`analyze_type(...) -> tuple[FieldShape, bool, str | None]`

The architecture diagram also names `TypeInfo` as an extraction-layer concept.

Beyond the docs: PR [#452](https://github.com/OvertureMaps/stac/pull/452) (Arrow schema generation) is written entirely against `TypeInfo/TypeKind` and cannot be rebased without rewriting its renderer, because the README described the API as current long after it was replaced.

6. Malformed release path in seeral places.

packages/overture-schema-pyspark/README.md uses `s3a://overturemaps-us-west-2/release/2026-06-17.0` 

7. overture-codegen list prints a raw type expression for discriminated union (cosmetic?)

```
$ overture-codegen list
Address
Bathymetry
...
typing.Annotated[typing.Union[typing.Annotated[overture.schema.transportation.segment.road.RoadSegment, Tag(tag=<SegmentSubtype.ROAD: 'road'>)], ...
```

`Segment` loads as an `Annotated` alias with no `__name__`, so the fallback prints its `repr`. Should print `Segment` or the entry-point name.

8. CLI `--help` renders literal `\b` instead of Click's formatting escape.
```
$ overture-schema validate --help
  \b Examples:   # Validate a file   $ overture-schema validate data.json \b
  # Validate from stdin   $ overture-schema validate - < data.json \b   #
  Validate only buildings   $ overture-schema validate --tag ...
```

The docstrings in `commands.py` use raw strings (`r"""`), so `\b` stays two characters rather than becoming the escape Click looks for. Result: every example block collapses into one paragraph. Fix is to drop the `r` prefix on those docstrings, or use an explicit `"\b"`.

## Already addressed
Three items from the original draft concerned documentation structure rather than correctness, and resolved in my test-docs branch by consolidating the Python docs into a single SCHEMA_GUIDE.md linked from README.md and CONTRIBUTING.md:

- PYDANTIC_GUIDE.md and README.pydantic.md have been absorbed and removed; every top-level markdown file is now reachable from README.md or CONTRIBUTING.md.
- README.pydantic.md conflated three audiences (project rationale, consumer, contributor). The replacement separates them
- README.pydantic.md documented the `discover_models()` return shape two ways in one file — tuple keys in one place, ModelKey objects in another. Only the correct form survives.

That consolidation also removed one of the six pip install instructions, one of the two sites of the broken top-level imports, one of the two model_validate()/GeoJSON claims, and one of the two sources of the phantom overture tag.

### LINKED ISSUE #678: Code samples in package READMEs need to be tested as part of the CI/CD process
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

dev ops

### Description

Several of the READMEs in the various Pydantic packages have code examples that will very quickly get out of date. We need to test these as part of the CI/CD process, similar to the way we test queries in Docusaurus.

sethfitz:
## Summary

Fixes the documentation inaccuracies catalogued in #604 and #668, the two CLI defects that produced some of them, and adds the tag that one of them described but nothing emitted. Adds a test that keeps the whole class from recurring.

**Two things want a decision rather than a review**, both marked below: what the new [`overture` tag should mean](#worth-discussing-before-this-merges-what-should-overture-mean), and whether a [top-level `tests/`](#worth-discussing-before-this-merges-a-top-level-tests) is the right home for a repo-wide check.

Every claim below was verified by running the code, not by reading it.

## Documentation

**`packages/overture-schema/README.md`** documented `from overture.schema import Building, Place` and `from overture.schema import parse, discover_models, json_schema`. None of it resolves: `overture.schema` is a namespace root shipping only `py.typed`, and `parse()` exists nowhere in the codebase. The README predates #622, which moved the validation API out of the namespace root. Rewritten against the real surface — models from their theme packages, `validate`/`validate_json` from `overture.schema.validation`, discovery and `json_schema` from `overture.schema.system`.

**Three READMEs claimed `model_validate()` accepts GeoJSON.** It does not. Python mode reads the flat tabular shape; JSON mode reads GeoJSON; the two are exclusive, and the error a GeoJSON dict produces (`theme Field required`) does not point at the cause. Each now states the split. The round-trip examples also pass `by_alias=True`, without which `class_` is emitted instead of `class` and the output will not re-validate.

**`packages/overture-schema-codegen/README.md`** described `analyze_type()` returning a `TypeInfo` with `.kind`/`.base_type`/`.newtype_name`, and imported `TypeKind`. Neither type exists; the signature is `(FieldShape, bool, str | None)`. Replaced with a worked example whose output is transcribed from an actual run. The layer diagram and the package credited with `discover_models` are corrected too.

**`packages/overture-schema-pyspark/README.md`** used release `2026-06-17.0` in its S3 examples. The bucket retains only the current release, so that path 404s — `aws s3 ls s3://overturemaps-us-west-2/release/` returns one prefix. Bumped, with a note on reading the current identifier from the STAC catalog's `latest` field, since any hardcoded release goes stale by design.

**`PYDANTIC_GUIDE.md`** prescribed a `models.py` / `enums.py` / `types.py` split inside a per-type subdirectory (#604). No theme package has such a file. Rewritten against the settled layout: one module per feature type at the theme root, `_common.py` for what a theme shares, a subpackage only for a type large enough to split. The JSON-Schema-container-to-mixin example now uses the real `Named` / `Appearance` mixins that `Building` actually inherits, and the container names on both sides (`namesContainer`, `shapeContainer`) exist in `schema/`.

## The `overture` tag

`overture-schema --help` suggested `--tag overture --tag feature` for "official Overture types only", and discovery emitted no such tag, so the filter matched nothing (#668 item 4). Rather than delete the reference, `overture-schema-common` now ships an `overture_provider` that attaches the tag when any concrete arm of an entry point subclasses `OvertureFeature`; the tag is reserved to that package.

"Is this built on Overture's feature model?" is now answerable by tag, so a consumer asking it needs neither a dependency on `overture-schema-common` nor an `issubclass` call.

The help example is also corrected on a second axis: `--tag` is OR, so pairing `overture` with `feature` would have widened the scope rather than narrowed it. `--tag overture` alone is the filter that means what the help said.

### Worth discussing before this merges: what should `overture` mean?

The help previously read "Official Overture types only", and that wording is wrong — it has been changed to "Only types built on the Overture feature model". The distinction matters, and people may reasonably expect either.

As implemented, the tag attaches to **any** `OvertureFeature` subclass reachable from an `overture.models` entry point, including a third party's:

```python
class AcmeParkingMeter(OvertureFeature[Literal["acme"], Literal["parking_meter"]]):
    pass
# -> tags: {"feature", "overture", "overture:theme=acme"}
```

So it means *"carries the Overture feature contract"* — theme, type, id, version, sources, geometry — as distinct from `feature`, which says only that something is a `Feature`. It is not a claim of membership in the Overture schema. Reserving the tag to `overture-schema-common` governs who may *emit* it, not which models receive it; the provider that emits it does so for everyone.

If the intent is the narrower "types Overture publishes", that is a different predicate — provenance rather than base class — and would need something the base class does not carry: a distributor allowlist, a separate `overture:approved` tag, or a reserved theme set. Worth settling now, because the tag is public surface as soon as this merges and the narrow reading is not a superset of the broad one.

Note also `overture:theme=acme` in the example above: the theme tag has the same property today, and predates this PR.

## CLI defects

**`--help` rendered a literal `\b`.** The command docstrings were raw strings, so Click's no-rewrap marker stayed two characters and every example block collapsed into one paragraph. Dropped the `r` prefix. `pydocstyle`'s D301 pushes the other way and is a false positive for Click docstrings, so it is waived with the reason recorded once at the module level.

**`overture-codegen list` printed a type expression instead of a name.** The listing fell back to `str(model)` for anything without `__name__`, so `Segment` emitted several hundred characters of discriminator internals. It now reads the name off the entry point, which every registered model has.

## Keeping it fixed

`tests/test_documented_imports.py` parses every fenced Python block in every tracked Markdown file and imports each `overture.*` statement it finds — 152 today, across 100+ files. A broken one fails the suite instead of accumulating.

The detector went through four review rounds, and the findings were mostly against the detector rather than the docs. Worth knowing what it does and does not do:

- Block extraction and import parsing are separate functions, each unit-tested against inline fixtures. The corpus cannot test either: a form the matcher misses simply yields nothing, and a sweep that collects less still passes. Two rounds found exactly that failure — a control that passed with the branch it named deleted, and an indent fix that was silently revertible.
- Blocks indented inside list items are matched and dedented. Without the dedent an `IndentationError` is a `SyntaxError`, which would file the block as "expected unparseable" and hide it twice.
- REPL transcripts are unwrapped rather than skipped.
- Blocks that are deliberately not valid Python are pinned by a digest of their body, not by count or index — a count stays put when one block breaks as another is fixed. An excused block may not contain an `overture` import, so the waiver cannot swallow the API citation it is excusing. It had, once.

`make lint-only` also now checks Python embedded in Markdown, with `make format` as the counterpart. That needs ruff 0.16, so the dependency floor moves from 0.13 — at the old floor the check finds no files and passes silently, and one CI job resolves at `lowest-direct`.

### Worth discussing before this merges: a top-level `tests/`

This adds the repo's first top-level `tests/` directory, which is a convention change worth a decision rather than a reviewer's shrug.

The test admits on two conditions, both recorded in `tests/README.md` so the directory does not become a catch-all:

- **Its subject spans packages, or sits outside them.** It reads every tracked Markdown file, including root-level ones (`PYDANTIC_GUIDE.md`, `GLOSSARY.md`, `CONTRIBUTING.md`) that ship in no package.
- **It cannot run from one package's install.** It imports across every package, `overture.schema.codegen` among them.

It started in `packages/overture-schema/tests/`, and that was worse in a specific way: it made that package's suite import a package its `pyproject.toml` does not declare, passing only because the workspace installs everything. Any package-scoped home has that problem.

**The cost, plainly.** `pytest packages/` is no longer complete. Three Makefile targets and the mypy `files` list had to learn about `tests/`, and a contributor running `pytest packages/` from habit misses it. `make check` does cover it — through ruff, `ruff format`, mypy and pytest.

**Prior art, and what it argues.** The repo already has repo-wide process code in Python: `.github/workflows/scripts/package_versions.py` and two action scripts. The convention there is to confine it to `.github/`, which is a real argument against a new top-level directory.

It also shows that convention's cost. Those three files sit outside every gate — `[tool.ruff] src`, mypy's `files`, and the pytest paths all point at `packages/`. That is not hypothetical: running the repo's own ruff config over them today reports two lint errors and two files needing reformatting, purely because nothing has ever applied it. Filed separately as #682; not fixed here, since it is unrelated to the documentation work.

So the choice is between a directory that is new but gated, and a convention that is established but has a hole in it.

**The alternative**, if a new top-level directory is unwelcome: a non-shipping workspace package (`packages/overture-schema-repo-checks/` or similar) that declares dependencies on everything. That keeps `pytest packages/` complete, at the cost of the repo's only package that publishes nothing — a `pyproject.toml`, a version, a `changelog.d/`, and an explicit exclusion from the publish workflow. I think that is more ceremony than the problem earns, but it is a one-commit change if the group disagrees.

Closes #604
Closes #668

### DIFF
--- a/tests/test_documented_imports.py
+++ b/tests/test_documented_imports.py
@@ -1,0 +1,368 @@
+"""Every `overture.*` import written in repo Markdown must resolve.
+
+Documentation drifts silently: a module moves, a helper is renamed, and the
+README keeps confidently describing the old surface. This parses every fenced
+Python block in every tracked Markdown file -- including blocks indented inside
+a list item -- and imports each `overture.*` statement it finds.
+
+The subject is the repo's Markdown rather than any one package, so this lives
+outside `packages/`. It reaches across the whole workspace -- a README may cite
+`overture.schema.codegen`, which no single distribution depends on.
+"""
+
+from __future__ import annotations
+
+import ast
+import hashlib
+import importlib
+import re
+import subprocess
+import textwrap
+from enum import Enum
+from pathlib import Path
+
+import pytest
+
+# A fenced ```python block, capturing its body. The indent group matches a
+# block nested in a list item; the body is dedented before parsing, since an
+# `IndentationError` is a `SyntaxError` and would file the block under
+# "expected unparseable" rather than reporting it.
+_PYTHON_BLOCK = re.compile(
+    r"^(?P<indent> *)```python[^\n]*\n(?P<body>.*?)^(?P=indent)```",
+    re.MULTILINE | re.DOTALL,
+)
+
+# Blocks that are deliberately not valid Python: illustrations whose `...`
+# placeholders stand in for elided content. Pinned by identity, not count -- a
+# bare total stays put when one block breaks as another is fixed. The identity
+# is a digest of the block body, so reordering a document does not trip it but
+# editing one of these blocks does.
+_EXPECTED_UNPARSEABLE = {
+    "PYDANTIC_GUIDE.md:43488ef4",
+}
+
+# An import statement, matched textually -- used to police the excuse list,
+# where by definition `ast` cannot be applied.
+_OVERTURE_IMPORT = re.compile(r"^\s*(?:from|import)\s+overture\b", re.MULTILINE)
+
+# Golden files are generated fixtures, not documentation.
+_EXCLUDED = "tests/golden/"
+
+
+def _repo_root() -> Path:
+    """Locate the checkout root, or skip the module when there isn't one."""
+    for candidate in Path(__file__).resolve().parents:
+        if (candidate / ".git").exists():
+            return candidate
+    pytest.skip("not running from a git checkout", allow_module_level=True)
+
+
+def _markdown_files() -> list[Path]:
+    root = _ROOT
+    try:
+        listed = subprocess.run(
+            ["git", "ls-files", "-z", "*.md"],
+            cwd=root,
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout
+    except (OSError, subprocess.CalledProcessError) as exc:  # pragma: no cover
+        pytest.skip(f"cannot list tracked files: {exc}", allow_module_level=True)
+    return [root / n for n in listed.split("\0") if n and _EXCLUDED not in n]
+
+
+def python_blocks(text: str) -> list[str]:
+    """Yield the dedented body of every fenced Python block in `text`.
+
+    Handles a block indented inside a list item. The dedent is load-bearing:
+    without it `ast.parse` raises `IndentationError`, a `SyntaxError`, and the
+    block would be filed as "expected unparseable" rather than reported.
+    """
+    return [textwrap.dedent(match["body"]) for match in _PYTHON_BLOCK.finditer(text)]
+
+
+def _documented_enum_references(
+    paths: list[Path],
+) -> list[tuple[Path, str, str, str]]:
+    """Collect `(file, module, enum, member)` quads across the corpus."""
+    found: list[tuple[Path, str, str, str]] = []
+    for path in paths:
+        for body in python_blocks(path.read_text(encoding="utf-8")):
+            try:
+                refs = enum_references(body)
+            except SyntaxError:
+                continue
+            found += [(path, module, name, member) for module, name, member in refs]
+    return found
+
+
+def parse_imports(source: str) -> list[tuple[str, str | None]]:
+    """Extract `overture.*` imports from one block of Python source.
+
+    Returns `(module, attribute)` pairs; `attribute` is None for a plain
+    `import overture.x`. Raises `SyntaxError` if the block is not valid Python.
+    """
+    found: list[tuple[str, str | None]] = []
+    for node in ast.walk(ast.parse(source)):
+        if isinstance(node, ast.ImportFrom):
+            # A relative import (`from . import x`) has no absolute module.
+            if node.level or not node.module or not _is_overture(node.module):
+                continue
+            found += [
+                (node.module, alias.name) for alias in node.names if alias.name != "*"
+            ]
+        elif isinstance(node, ast.Import):
+            found += [
+                (alias.name, None) for alias in node.names if _is_overture(alias.name)
+            ]
+    return found
+
+
+def _is_overture(module: str) -> bool:
+    return module == "overture" or module.startswith("overture.")
+
+
+def enum_references(source: str) -> list[tuple[str, str, str]]:
+    """Find `Enum.MEMBER` uses of enums imported in the same block.
+
+    Returns `(module, enum_name, member)`. Restricted to enums on purpose:
+    their members really are class attributes, whereas a Pydantic model's
+    fields are not, so `Place.addresses` would read as missing.
+    """
+    tree = ast.parse(source)
+    imported = {
+        alias.asname or alias.name: node.module
+        for node in ast.walk(tree)
+        if isinstance(node, ast.ImportFrom)
+        and node.module
+        and not node.level
+        and _is_overture(node.module)
+        for alias in node.names
+    }
+    return [
+        (imported[node.value.id], node.value.id, node.attr)
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Attribute)
+        and isinstance(node.value, ast.Name)
+        and node.value.id in imported
+        and node.attr.isupper()  # a member, not a method
+    ]
+
+
+def _documented_imports(
+    paths: list[Path],
+) -> tuple[list[tuple[Path, str, str | None]], dict[str, str]]:
+    """Collect `(file, module, attribute)` triples, and identify unparseable blocks."""
+    found: list[tuple[Path, str, str | None]] = []
+    unparseable: dict[str, str] = {}
+    for path in paths:
+        for body in python_blocks(path.read_text(encoding="utf-8")):
+            try:
+                imports = parse_imports(body)
+            except SyntaxError:
+                digest = hashlib.sha256(body.encode()).hexdigest()[:8]
+                unparseable[f"{path.relative_to(_ROOT)}:{digest}"] = body
+                continue
+            found += [(path, module, attr) for module, attr in imports]
+    return found, unparseable
+
+
+_ROOT = _repo_root()
+_MARKDOWN = _markdown_files()
+_DOCUMENTED, _UNPARSEABLE = _documented_imports(_MARKDOWN)
+_ENUM_REFERENCES = _documented_enum_references(_MARKDOWN)
+
+
+class TestBlocks:
+    """`python_blocks` against inline fixtures.
+
+    Same argument as `TestParser` below, one layer up: a fence form the
+    matcher misses yields no block, and a sweep that collects less still
+    passes. The corpus cannot pin this.
+    """
+
+    def test_flush_block(self) -> None:
+        text = "text\n\n```python\nx = 1\n```\n\nmore\n"
+        assert python_blocks(text) == ["x = 1\n"]
+
+    def test_list_item_block(self) -> None:
+        """A block indented under a list item, dedented to parse."""
+        text = "1. Example:\n\n   ```python\n   x = 1\n   ```\n"
+        assert python_blocks(text) == ["x = 1\n"]
+
+    def test_deeply_indented_block(self) -> None:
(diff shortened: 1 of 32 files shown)


tests/test_documented_imports.py
```diff
@@ -0,0 +1,368 @@
+"""Every `overture.*` import written in repo Markdown must resolve.
```
vcschapp, line 1: Can we broaden this to: "every Python code block written in repo Markdown must execute without error"?

The broader formulation doesn't seem so hard to do, and would be a much more powerful check on the accuracy of the docs.

vcschapp (changes requested):
Can we expand the markdown check scope to running the code block and checking for no errors?

tests/test_documented_imports.py
sethfitz: No, not now. That's #491, and it's a lot less straightforward (variable, common setup between blocks) whilst having some options we can lean on.

vcschapp (approved):

danabauer (approved):

PR #672, opened 2026-08-19
[DOCS] test, update, revise, and consolidate schema guides

danabauer:
## Summary

_Note: this is a rough draft of the guide meant for early testers (pre-PyPI publication) from now until the `v2.0.0` launch in early September. Leading up to that launch I will revise this guide to reflect the feedback from testers and the new pathways for accessing the packages, i.e. `pip install`. See #680_ 

closes #670 

Replaces `PYDANTIC_GUIDE.md` and `README.pydantic.md` with a single `SCHEMA_GUIDE.md`,
linked from `README.md` and `CONTRIBUTING.md`, and expands `GLOSSARY.md` from a six-term
data-model glossary into a 22-term glossary covering both the data model and the Python
toolchain.

Every code block in the new guide has been executed against the repo. That check found
four defects in the material carried over from `PYDANTIC_GUIDE.md`, all fixed here. I filed additional bugs in this repo. 

## Motivation

Three problems, all of which the consolidation addresses:

1. **It was hard to reach the existing Pydantic docs** `PYDANTIC_GUIDE.md` and `README.pydantic.md` were
   referenced by no file in the repository, and `README.md` did not mention the Python
   packages at all. A contributor had no path to either document except by browsing the
   file listing.
2. **They conflated three audiences.** `README.pydantic.md` interleaved project rationale
   ("why Pydantic over JSON Schema"), consumer instructions, and contributor setup, so no
   part of it could be corrected or updated independently of the others.
3. **They contained errors that had never been executed.** The authoring guide's own
   templates included a `SyntaxError` and two references to an enum member that does not
   exist.

## Changes

| File | What |
| --- | --- |
| `SCHEMA_GUIDE.md` | New. 4,684 lines in four parts: rationale, using the schema, extending and authoring it, reference. Absorbs both deleted files. |
| `GLOSSARY.md` | 18 → 169 lines, 6 → 22 terms. Original six definitions preserved verbatim; adds data-model terms (theme, type, subtype, GERS) and thirteen toolchain terms. Cross-linked with the guide. |
| `README.md` | Adds a "Python packages" section and a "Schema reference" section linking the guide, the glossary, and the conventions doc. |
| `CONTRIBUTING.md` | Links the guide's authoring and development-workflow material. |
| `PYDANTIC_GUIDE.md` | Deleted. 1,362 lines absorbed into Part II and Part III. |
| `README.pydantic.md` | Deleted. 233 lines absorbed into Part 0, Part II, and the development-workflow section. |

### Structure

[...]

PR #684, opened 2026-08-19
[DOCS] Consolidate the Python docs into SCHEMA_GUIDE.md

### LINKED ISSUE #685: [DOCS] Python documentation is unreachable and conflates audiences
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

engineering

### Description

The repository's Python documentation is unreachable and conflates several audiences.

**Unreachable.** `PYDANTIC_GUIDE.md` and `README.pydantic.md` are referenced by no file in the repository, and `README.md` does not mention the Python packages at all. A contributor has no path to either document except by browsing the file listing.

**Conflated audiences.** `README.pydantic.md` interleaves project rationale (why Pydantic rather than JSON Schema), consumer instructions (install, validate data), and contributor setup. No part of it can be revised independently of the others, which is part of why it drifts.

**Unverified.** Both documents accumulated errors that were never executed. #604 and #668 catalogue the ones found so far; #683 fixes them and adds a check for the class. That check does not address structure, and structure is what makes the documents hard to keep correct: a reader who cannot find a document cannot report that it is wrong.

### Suggested approach

Consolidate into a single guide, linked from `README.md` and `CONTRIBUTING.md`, with the audiences separated at the top level rather than interleaved — rationale, using the schema, extending and authoring it, reference. Expand `GLOSSARY.md` to cover both the data model (theme, type, subtype, GERS) and the Python toolchain (entry point, workspace, discriminated union), since the two vocabularies are currently mixed in prose without definition.

### Acceptance

- Every top-level Markdown file is reachable from `README.md` or `CONTRIBUTING.md`.
- Each audience can be revised without touching the others.
- The guide's code examples pass the checks added in #683.

### Notes

@danabauer opened #672 doing this work; it predates the fixes in #683 and so carries the stale content forward. #684 is that work rebased onto #683 with the corrections applied.

sethfitz:
## Summary

Replaces #672. Same consolidation — `PYDANTIC_GUIDE.md` and `README.pydantic.md` become a single `SCHEMA_GUIDE.md`, linked from `README.md` and `CONTRIBUTING.md`, and `GLOSSARY.md` grows from 6 to 22 terms — rebased onto the documentation fixes in #683, with the corrections carried into the new guide. @danabauer's commit is preserved with its authorship; the second commit is the port.

#672 can be closed in favour of this; it does not rebase cleanly, because the files it deletes are the files #683 edits.

## Why this replaces it rather than stacking on it

#672 was written against `main` before the doc fixes landed, so it did two things this branch undoes:

1. **It carried the stale content forward.** #604's `models.py` / `enums.py` / `types.py` file-type rules, the per-type-subdirectory prescription, and `# In buildings/models.py` all survive verbatim into `SCHEMA_GUIDE.md`. Those sections are now taken from the corrected `PYDANTIC_GUIDE.md` — extracted from the file programmatically rather than retyped, so the two cannot disagree.

2. **It catalogued the defects instead of fixing them.** The guide gained a "Docs in the repo that are currently wrong" section listing eight items. Seven are fixed in #683. Only `pip install` survives, and it now points at the section explaining why. A standing list of known-wrong docs is a workaround for not having a check; #683 adds the check.

## Other corrections

The transcribed `overture-schema list-types` output (two copies) and the `--group-by` output were regenerated by running the commands, so they show the `overture` tag that #683 adds. The `ModelKey` examples were showing `BuildingModel` / `PlaceModel`, which are not the names of anything.

`GLOSSARY.md`'s Tag entry said "Seven tags exist today ... There is no plain `overture` tag, despite what some help text suggests." Accurate when written, and the observation is what prompted #683 to add the tag rather than delete the help that referenced it. Now eight, with a note on what the tag does and does not assert — see the open question in #683, which is the place to weigh in on it.

## What importing the guide exposed

`SCHEMA_GUIDE.md` broke the drift detector twice on arrival, which is a fair argument for having it:

- A `>>>` REPL block failed to parse and dropped out of the sweep whole, taking its imports with it. Transcripts are unwrapped now.
- The guide deliberately shows `from overture.schema import Building  # ✗ ImportError` as a counter-example, and the detector read it as a defect. A line carrying the counter-example glyph is exempt.

Both behaviours have unit fixtures and a mutation that fails them.

## Review notes

- **@danabauer's own checks still hold.** Every code block in the guide parses, every `overture.*` import resolves, every internal link resolves. Two of those are now enforced by `tests/test_documented_imports.py` on every run rather than checked once.
- The `SCHEMA_CONVENTIONS.md` staleness flags in `README.md` were spot-checked and are accurate: it does describe JSON Schema as the definition mechanism (line 3) and does spell `subType` (line 37).
- `CONTRIBUTING.md` still does not mention the DCO sign-off requirement. Carried over unaddressed, as in #672.

Closes #685


lowlydba:
Suggestion:

With really wide-reaching/complex docs like this (several audiences, several different use cases) I really like to cornerstone them with [Diataxis](https://diataxis.fr/), which is roughly just:

> Diátaxis identifies four distinct needs, and four corresponding forms of documentation - **tutorials, how-to guides, technical reference and explanation**. It places them in a systematic relationship, and proposes that documentation should itself be organised around the structures of those needs.

One of the bonuses of this is models seem *very* adept at following/implementing it just by name reference with "Do the docs in Diataxis", so you can almost completely offload the information theory  cognitive load of organizing big documentation, while still being consistent.

RoelBollens-TomTom (approved):

vcschapp (approved):

PR #692, opened 2026-08-24
[ENHANCEMENT](transportation) Make segment connectors required

### LINKED ISSUE #669: Segments with no connectors fail validation after a round trip
### Type

Bug

### Scope

Transportation

### Skillset

engineering

### What is happening

`connectors` on the segment models declares an empty-list default that its own
`min_length=2` constraint rejects:

`packages/overture-schema-theme-transportation/src/overture/schema/transportation/segment/_common.py:174-187`

```python
connectors: Annotated[
    list[ConnectorReference] | None,
    Field(
        min_length=2,
        description="...",
    ),
    UniqueItemsConstraint(),
] = []          # ← violates min_length=2
```

Pydantic does not validate defaults unless `validate_default=True`, so the invalid value
is accepted silently on the way in and rejected on the way back out. A document that the
library itself validates cannot be re-validated after a dump.

Affects `RoadSegment`, `RailSegment`, and `WaterSegment` — all three inherit the single
declaration in `_common.py`.

### Why it matters

1. **Round-trips fail.** Read → dump → read is not idempotent for any segment without a
   `connectors` key. This breaks the normal shape of a data pipeline.
2. **The published JSON Schema is self-contradictory.** `overture-schema json-schema
   --type segment` emits, for each of the three segment arms:

   ```json
   {
     "default": [],
     "items": { "$ref": "#/$defs/ConnectorReference" },
     "minItems": 2,
     "title": "Connectors",
     "type": "array",
     "uniqueItems": true
   }
   ```

   `"default": []` alongside `"minItems": 2`. Every downstream consumer of the JSON
   Schema inherits the contradiction — generated SDKs (`datamodel-codegen`,
   `quicktype`), validators, and documentation.
3. **The absent/empty distinction is lost.** "No connectors recorded" and "connectors
   recorded, and there were none" become the same value on the model, and neither is
   representable as valid.

### Reproduction

Against `examples/transportation/segment/road/road-indoors.yaml`, which has no
`connectors` key. The CLI accepts the file:

```bash
uv run overture-schema validate examples/transportation/segment/road/road-indoors.yaml
# ✓ Successfully validated examples/transportation/segment/road/road-indoors.yaml
```

The same content fails after a round-trip:

```python
import json, yaml
from pydantic import TypeAdapter
from overture.schema.transportation import Segment

segments = TypeAdapter(Segment)
src = yaml.safe_load(open("examples/transportation/segment/road/road-indoors.yaml"))

seg = segments.validate_json(json.dumps(src))
seg.connectors           # []   ← not None, despite the key being absent

flat = seg.model_dump(mode="python", by_alias=True, exclude_none=True)
flat["connectors"]       # []   ← exclude_none does not drop it

segments.validate_python(flat)
# pydantic_core._pydantic_core.ValidationError: 1 validation error for tagged-union[...]
# road.connectors
#   List should have at least 2 items after validation, not 0
#   [type=too_short, input_value=[], input_type=list]
```

Provider-independent minimal case:

```python
from typing import Annotated
from pydantic import BaseModel, Field

class Broken(BaseModel):
    items: Annotated[list[str] | None, Field(min_length=2)] = []

m = Broken.model_validate({})
Broken.model_validate(m.model_dump(exclude_none=True))
# ValidationError: items — List should have at least 2 items after validation, not 0
```

### Scope

An audit of every field reachable from the registered models — 29 models walked,
each field's default validated against that field's own annotation and metadata —
returns exactly three violations, all from this one declaration:

```
RoadSegment.connectors   []   List should have at least 2 items after validation, not 0
RailSegment.connectors   []   List should have at least 2 items after validation, not 0
WaterSegment.connectors  []   List should have at least 2 items after validation, not 0
```

It is also the only model field in the schema packages that defaults to `[]`:

```bash
grep -rn '^\s*\] = \[\]' packages/overture-schema-theme-*/src \
    packages/overture-schema-common/src packages/overture-schema-system/src --include='*.py'
# .../transportation/segment/_common.py:187:    ] = []
```

So this is a one-line fix, not a pattern that needs a sweep.

### Proposed fix

Change the default to `None`, matching every other optional field on these models:

```python
] = None
```

Verified on the minimal case:

```
Broken  default=[]    dump={'items': []}   FAILS: items
Fixed   default=None  dump={}              round-trips OK
```

Alternatives, if `None` is wrong for the domain:

- **Drop `min_length=2`** if a segment with zero connectors is legal data. This changes
  what the schema asserts, so it needs a schema-semantics decision rather than a code fix.
- **Make the field required** if a segment must always carry connectors. Larger breaking
  change; probably `vnext` territory.

`= None` is the recommendation: it preserves the current constraint, restores the
absent/empty distinction, and matches the surrounding field declarations.

### Constraints / side effects to check

- **Consumer semantics change.** Anything doing `for c in seg.connectors` without a
  `None` guard breaks. `seg.connectors or []` is the safe idiom, and worth grepping for
  in dependent repos.
- **JSON Schema output changes.** The `"default": []` key disappears from the three
  segment arms. This will move golden baseline files — run `make update-baselines` and
  review the diff.
- **Generated PySpark expressions** should be regenerated (`make generate-pyspark`) and
  the diff reviewed, since the field's nullability is part of the expected `StructType`.
- A changelog fragment is required under
  `packages/overture-schema-theme-transportation/changelog.d/`.

### Acceptance criteria

- [ ] `connectors` no longer declares a default that its own constraints reject.
- [ ] `overture-schema json-schema --type segment` emits no `"default": []` next to
      `"minItems": 2`.
- [ ] Round-trip holds: for every example under `examples/transportation/`,
      `validate_json(src)` → `model_dump(by_alias=True)` → `validate_python(...)`
      succeeds.
- [ ] A regression test asserts the round-trip for a segment with no `connectors` key.
- [ ] Baselines regenerated and the diff reviewed; PySpark expressions regenerated.
- [ ] Changelog fragment added.

### Suggested regression test

Worth making this generic rather than `connectors`-specific — it catches the whole class
of "default violates its own constraints" for any field added later:

```python
import pytest
from pydantic import BaseModel, TypeAdapter
from pydantic_core import PydanticUndefined
from typing import Annotated, get_args

from overture.schema.system.discovery import discover_models


def _models_in(ann, depth=0):
    if depth > 6:
        return
    if isinstance(ann, type) and issubclass(ann, BaseModel):
        yield ann
        return
    for a in get_args(ann) or ():
        yield from _models_in(a, depth + 1)


def _all_models():
    seen = set()
    stack = [m for key in discover_models() for m in _models_in(key and discover_models()[key])]
    while stack:
        model = stack.pop()
        if model in seen:
            continue
        seen.add(model)
        for f in model.model_fields.values():
            stack.extend(_models_in(f.annotation))
    return seen


@pytest.mark.parametrize("model", sorted(_all_models(), key=lambda m: m.__name__))
def test_field_defaults_satisfy_their_own_constraints(model):
    for name, f in model.model_fields.items():
        if f.default is PydanticUndefined or f.default is None:
            continue
        ann = Annotated[tuple([f.annotation, *f.metadata])] if f.metadata else f.annotation
        TypeAdapter(ann).validate_python(f.default)   # must not raise
```

(Sketch — the `_all_models()` traversal wants tidying against the real discovery API
before it lands.)

### Environment

- schema repo at commit `2a6170c4`
- pydantic 2.13.4, Python 3.10
- All claims re-verified 2026-08-19 against a fresh clone: the declaration is still at
  `_common.py:187`, it is still the only model field defaulting to `[]`, the round trip
  still fails, and `json-schema --type segment` still emits `"default": []` beside
  `"minItems": 2` on all three arms.

sethfitz:
# Major change release plan

Bundled into 2.0.0. Since it matches the published data, it doesn't strictly need to be treated as a breaking change.

- None. The published data already satisfies the tightened constraint, so there is no intermediate step where producers must be given time to comply.

The schema reference regenerates from the model, and every Docusaurus segment example in this PR now carries a `connectors` array, so the visible change is that `connectors` moves into the required-property list for all three segment subtypes. This is effectively a fix for the docs to match what's published in the data.

# Description

`connectors` was declared optional with an empty-list default that its own `min_length=2` constraint rejects:

```python
connectors: Annotated[
    list[ConnectorReference] | None,
    Field(min_length=2, description="..."),
    UniqueItemsConstraint(),
] = []          # violates min_length=2
```

Pydantic does not validate defaults unless `validate_default=True`, so the invalid value was accepted on the way in and refused on the way back out. A segment constructed without connectors dumped to `connectors: []`, and re-validating that dump failed — read → dump → read was not idempotent for any segment lacking connectors.

Rather than correct the default, this makes the field required. A segment is by definition physically connected to at least two connectors, the published data has always reflected that, and an optional field whose only legal values are arrays of two or more was never expressing the real constraint.

Changes:

- `connectors` is now `list[ConnectorReference]` with no default, declared beside `subtype` in the model's required section.
- The YAML schema gains `connectors` in its segment `required` list and drops the matching `default: []` that sat beside `minItems: 2`.
- Every valid segment example gains a two-connector array. Every segment counterexample that was not itself testing connectors gains one too, so each keeps failing only for the defect it names (checklist item 3). Both example trees are updated: `reference/` (read by the test suite) and the root `examples/`/`counterexamples/` trees (the docs build symlinks `examples/`). `theme-type-mismatch.json` is deliberately untouched — it declares `theme: buildings` with `type: segment` and never reaches a segment arm.

# Reference

1. Fixes #669
2. The same contradiction in the pre-Pydantic schema: https://github.com/OvertureMaps/schema/blob/main/schema/transportation/segment.yaml#L61-L79 (noted by @vcschapp on #669)

# Testing

**No published data is affected.** Every segment in the most recent release satisfies the tightened constraint — verified against release `2026-08-19.0` (350,469,378 segments across 128 partitions), which returned zero nulls, zero empty arrays, and zero single-element arrays for all three subtypes:

```sql
SELECT
  subtype,
  count(*)                               AS n_rows,
  count_if(connectors IS NULL)           AS null_connectors,
  count_if(cardinality(connectors) = 0)  AS empty_connectors,
  count_if(cardinality(connectors) = 1)  AS one_connector,
  count_if(cardinality(connectors) >= 2) AS two_or_more
FROM v2026_08_19_0
WHERE theme = 'transportation' AND "type" = 'segment'
GROUP BY GROUPING SETS ((subtype), ())
ORDER BY subtype;
```

**This PR adds no new test, and the example updates are not proofs of the fix** — they are the corpus catching up to a tightened schema. What pins the change is the regenerated JSON Schema baseline, which fails if `connectors` stops being required, and the round trip that motivated #669, which now succeeds because the field has no default at all rather than a corrected one.

Separately verified while investigating: across every field default in the whole model tree, the three segment arms and their shared `TransportationSegment` base were the only ones their own annotations reject. Nothing else in the schema had the defect.

**Required-ness propagated to the generated PySpark layer**, not just the JSON Schema: the regenerated `transportation/segment.py` now emits `_connectors_check` with `check_required(F.col("connectors"))`. Control: optional `access_restrictions` still gets no such check.

**JSON Schema baseline delta is exactly `connectors` joining `required`** in all three arms. The `connectors` subschema itself is byte-identical to before, because the generator already emitted a non-nullable array and expressed optionality solely through `required`.

Full suite: 6215 passed.

# Checklist

1. [x] Add relevant examples.
3. [x] Update any counterexamples that became obsolete.
4. [x] Update in-schema documentation using plain English written in complete sentences, if an update is required. — The field description was already accurate; no wording change needed.
5. [ ] Update Docusaurus documentation, if an update is required. — The embedded examples are updated; prose pages not reviewed.

# Documentation website

[Docs preview for this PR.](https://dfhx9f55j8eg5.cloudfront.net/pr/692)

### DIFF
--- a/counterexamples/common/names/names-invalid-language.yaml
+++ b/counterexamples/common/names/names-invalid-language.yaml
@@ -10,6 +10,11 @@ properties:
   version: 1
   subtype: road
   class: primary
+  connectors:
+    - connector_id: fooConnector
+      at: 0
+    - connector_id: barConnector
+      at: 1
   names:
     primary: Foo
     common:
--- a/counterexamples/common/names/names-invalid-variant.yaml
+++ b/counterexamples/common/names/names-invalid-variant.yaml
@@ -10,6 +10,11 @@ properties:
   version: 1
   subtype: road
   class: primary
+  connectors:
+    - connector_id: fooConnector
+      at: 0
+    - connector_id: barConnector
+      at: 1
   names:
     primary: Bar
     rules:
--- a/counterexamples/transportation/segment/bad-geometry-type.json
+++ b/counterexamples/transportation/segment/bad-geometry-type.json
@@ -12,6 +12,10 @@
     "theme": "transportation",
     "type": "segment",
     "subtype": "water",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ],
     "version": 15
   }
 }
--- a/counterexamples/transportation/segment/bad-lr-width-missing-items.json
+++ b/counterexamples/transportation/segment/bad-lr-width-missing-items.json
@@ -11,6 +11,10 @@
     "version": 26,
     "subtype": "road",
     "class": "primary",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ],
     "width_rules": [],
     "ext_expected_errors": [
       "minItems: got 0, want 1"
--- a/counterexamples/transportation/segment/bad-lr-width.json
+++ b/counterexamples/transportation/segment/bad-lr-width.json
@@ -11,6 +11,10 @@
     "version": 26,
     "subtype": "road",
     "class": "primary",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ],
     "width_rules": [
       {
         "between" : [0, 0.5],
--- a/counterexamples/transportation/segment/missing-subtype.json
+++ b/counterexamples/transportation/segment/missing-subtype.json
@@ -11,6 +11,10 @@
     ],
     "theme": "transportation",
     "type": "segment",
-    "version": 26
+    "version": 26,
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ]
   }
 }
--- a/counterexamples/transportation/segment/missing-version.json
+++ b/counterexamples/transportation/segment/missing-version.json
@@ -11,6 +11,10 @@
     ],
     "theme": "transportation",
     "type": "segment",
-    "subtype": "water"
+    "subtype": "water",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ]
   }
 }
--- a/counterexamples/transportation/segment/property-subtype-mismatch.yaml
+++ b/counterexamples/transportation/segment/property-subtype-mismatch.yaml
@@ -10,6 +10,11 @@ properties:
   version: 1
   # Segment of subtype rail has a road flag (should be disallowed)
   subtype: rail
+  connectors:
+    - connector_id: fooConnector
+      at: 0
+    - connector_id: barConnector
+      at: 1
   road_flags:
     - values: [is_indoor]
   ext_expected_errors:
--- a/counterexamples/transportation/segment/rail/bad-rail-class.json
+++ b/counterexamples/transportation/segment/rail/bad-rail-class.json
@@ -13,6 +13,10 @@
     "type": "segment",
     "version": 18,
     "subtype": "rail",
-    "class": "foo"
+    "class": "foo",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ]
   }
 }
--- a/counterexamples/transportation/segment/rail/bad-rail-flags-duplicate-flag.json
+++ b/counterexamples/transportation/segment/rail/bad-rail-flags-duplicate-flag.json
@@ -14,6 +14,10 @@
     "version": 19,
     "subtype": "rail",
     "class": "primary",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ],
     "rail_flags": [
       {
         "values": ["is_freight", "is_freight"]
--- a/counterexamples/transportation/segment/rail/bad-rail-flags-invalid-flag.json
+++ b/counterexamples/transportation/segment/rail/bad-rail-flags-invalid-flag.json
@@ -14,6 +14,10 @@
     "version": 20,
     "subtype": "rail",
     "class": "primary",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ],
     "rail_flags": [
       {
         "values": ["foo"]
--- a/counterexamples/transportation/segment/road/bad-road-class.json
+++ b/counterexamples/transportation/segment/road/bad-road-class.json
@@ -13,6 +13,10 @@
     "type": "segment",
     "version": 18,
     "subtype": "road",
-    "class": "foo"
+    "class": "foo",
+    "connectors": [
+      { "connector_id": "fooConnector", "at": 0 },
+      { "connector_id": "barConnector", "at": 1 }
+    ]
   }
 }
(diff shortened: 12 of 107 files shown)


vcschapp (dismissed):

vcschapp:
> ## ⚠️ PR target check (advisory)
> * This PR targets `main` but has the `change type - major 🚨` label. Major/breaking changes should typically target `vnext` instead.
> 
> _Reviewer is the source of truth for change classification. This warning is advisory only._

We should discuss this. I'm not 100% clear on how we should handle the pre-`2.0.0` breaking change collection.

vcschapp:
I think it makes sense to target this to `vnext`. Then we should merge `vnext` down to `main` prior to doing the v2.0.0 release.

@lowlydba Do we have to do anything to keep `vnext` up-to-date with `main` before merging a PR to `vnext`?

danabauer (dismissed):

lowlydba:
> Do we have to do anything to keep `vnext` up-to-date with `main` before merging a PR to `vnext`?

@vcschapp

No, that's already automated: `.github/workflows/rebase-vnext.yaml` force-rebases `vnext` onto `main` on every push to `main` (via the `overture-pull-requester` app), and skips itself on the `vnext`->`main` release merge to avoid a no-op. Right now `vnext` and `main` are at the same commit (`5dc32f39`), so nothing to do before retargeting this to `vnext`.

If a rebase ever conflicts, the workflow opens an issue assigned to the triggering PR's author for manual resolution.

RoelBollens-TomTom (approved):

vcschapp (approved):

PR #708, opened 2026-08-28
[FEATURE] Attach JSON Schema as a release artifact

### LINKED ISSUE #671: Publish the JSON schema as a release artifact
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

dev ops

### Description

See #664 - the json version of the schema is the replacement. Figure out where it should live.

vcschapp:
# Description

Updates the GitHub release automation to attach full JSON Schema as `overture-schema.json` as a release artifact on `overture-schema` release.

# References

Closes #671.

# Testing

- New unit tests, `$ make check`
- Verified that `uv run overture-schema json-schema --tag overture`  emits JSON Schema as expected.
- Validated as much of the workflow manually as was possible.

### DIFF
--- a/.github/actions/create-package-release/action.yml
+++ b/.github/actions/create-package-release/action.yml
@@ -47,6 +47,13 @@ inputs:
   token:
     description: "Token used to create the release (needs contents: write)."
     required: true
+  assets:
+    description: >
+      Optional whitespace-separated list of asset files to attach to the
+      release (e.g. overture-schema.json). Uploaded atomically as part of
+      `gh release create`. Ignored on a dry run.
+    required: false
+    default: ""
 
 outputs:
   release-url:
@@ -68,6 +75,7 @@ runs:
         LATEST: ${{ inputs.latest }}
         VANITY_TAG: ${{ inputs.vanity-tag }}
         DRY_RUN: ${{ inputs.dry-run }}
+        ASSETS: ${{ inputs.assets }}
       run: |
         set -euo pipefail
 
@@ -91,6 +99,16 @@ runs:
         latest_flag="--latest=false"
         [ "$LATEST" = "true" ] && latest_flag="--latest"
 
+        # Validate requested assets exist before the dry-run branch, so a dry
+        # run catches a missing or misnamed asset.
+        read -ra assets <<< "$ASSETS"
+        for asset in "${assets[@]}"; do
+          if [ ! -f "$asset" ]; then
+            echo "::error::Release asset not found: ${asset}"
+            exit 1
+          fi
+        done
+
         if [ "$DRY_RUN" = "true" ]; then
           [ "$tag_exists" = "true" ] && echo "::warning::Release ${TAG} already exists; a real run would fail here."
           {
@@ -99,6 +117,9 @@ runs:
             echo "Would create tag \`${TAG}\` at \`${TARGET}\` (${latest_flag})."
             if [ -n "$VANITY_TAG" ]; then
               echo "Would also create vanity tag \`${VANITY_TAG}\` (no release)."
+            fi
+            if [ ${#assets[@]} -gt 0 ]; then
+              echo "Would attach assets: \`${assets[*]}\`."
             fi
             echo ""
             echo "<details><summary>Notes</summary>"
@@ -116,7 +137,8 @@ runs:
           --target "$TARGET" \
           --title "\`${PACKAGE}\` ${VERSION}" \
           --notes-file "${RUNNER_TEMP}/notes.md" \
-          $latest_flag
+          $latest_flag \
+          "${assets[@]}"
 
         url=$(gh release view "$TAG" --repo "$GITHUB_REPOSITORY" --json url --jq .url)
         echo "release-url=${url}" >> "$GITHUB_OUTPUT"
@@ -142,5 +164,9 @@ runs:
           if [ -n "$VANITY_TAG" ]; then
             echo ""
             echo "Vanity tag \`${VANITY_TAG}\` continues the legacy series (no separate release)."
+          fi
+          if [ -n "$ASSETS" ]; then
+            echo ""
+            echo "Attached assets: \`${ASSETS}\`."
           fi
         } >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/release-trigger.yaml
+++ b/.github/workflows/release-trigger.yaml
@@ -101,6 +101,21 @@ jobs:
         with:
           persist-credentials: false
 
+      # For the `overture-schema package, `uv` is needed to generate the
+      # official JSON Schema release asset.
+      - name: Install uv
+        if: matrix.package == 'overture-schema'
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          enable-cache: false
+
+      - name: Generate unified JSON schema
+        if: matrix.package == 'overture-schema'
+        run: |
+          set -euo pipefail
+          uv sync --locked --all-packages
+          uv run overture-schema json-schema --tag overture > overture-schema.json
+
       - name: Create release
         uses: ./.github/actions/create-package-release
         with:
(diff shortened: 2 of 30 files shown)


.github/workflows/release-trigger.yaml
```diff
@@ -101,6 +101,20 @@ jobs:
         with:
           persist-credentials: false
 
+      # For the `overture-schema package, `uv` is needed to generate the
+      # official JSON Schema release asset.
+      - name: Install uv
```
lowlydba, line 106: issue: this hardcodes `matrix.package == 'overture-schema'` into the generic release-trigger workflow. `main-publish.yaml` already has a package-owned hook for this shape of problem: `packages/<pkg>/scripts/prebuild.sh`, run unconditionally and a no-op if absent (see `overture-schema-pyspark`'s use of it). Worth extending that convention with a same-shaped release-asset hook instead of teaching this workflow specific package names. Every future per-package release asset otherwise adds another `if` here.

.github/workflows/release-trigger.yaml
```diff
@@ -101,6 +101,20 @@ jobs:
...
+        if: matrix.package == 'overture-schema'
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          enable-cache: false
+
+      - name: Generate unified JSON schema
+        if: matrix.package == 'overture-schema'
+        run: |
+          uv sync --locked --all-packages
+          uv run overture-schema json-schema --tag overture > overture-schema.json
```
lowlydba, line 117: nitpick (non-blocking): worth a `set -euo pipefail` here to match the composite action right after it. GitHub's default bash steps already run with `-eo pipefail`, so this is really just about `-u`, but a broken `uv sync` or CLI error here should fail loud rather than risk a partial `overture-schema.json` slipping through to the asset check.

.github/workflows/release-trigger.yaml
```diff
@@ -101,6 +101,20 @@ jobs:
...
+
       - name: Create release
```
lowlydba, line 119: suggestion: attach a GitHub Attestation (`actions/attest-build-provenance`) for `overture-schema.json` before it's handed to `create-package-release`. Everything else in this pipeline has some provenance guarantee: PyPI Trusted Publishing, CodeArtifact behind an OIDC-scoped role. This is the first artifact shipped as a raw unsigned blob with no verifiable link back to the run that built it.

.github/actions/create-package-release/action.yml
```diff
@@ -111,12 +122,23 @@ runs:
           exit 0
         fi
 
+        # Fail early with a clear message if a requested asset is missing. Here
+        # we deliberately leave `$ASSETS` unquoted so the shell splits it into
+        # whitespace-delimited words.
+        for asset in $ASSETS; do
```
lowlydba: issue: this existence check runs after the `DRY_RUN` branch already exits above, so a dry run never catches a missing or misnamed asset. That's the one part of this action that's actually new and fragile, and dry run exists specifically to catch this class of mistake before a real release.

.github/actions/create-package-release/action.yml
```diff
@@ -111,12 +122,23 @@ runs:
...
+        done
+
         gh release create "$TAG" \
           --repo "$GITHUB_REPOSITORY" \
           --target "$TARGET" \
           --title "\`${PACKAGE}\` ${VERSION}" \
           --notes-file "${RUNNER_TEMP}/notes.md" \
-          $latest_flag
+          $latest_flag \
+          $ASSETS
```
lowlydba: question: is the whitespace-separated list actually needed? There's exactly one caller today, and it passes exactly one file. The unquoted splat trades shell safety for generality nobody's using yet. If `overture-schema.json` grows a checksum or attestation sidecar, a directory glob would read cleaner than accumulating filenames in one string.

vcschapp: I'll meet you in the middle and update it to read the assets into an array so glob expansion of the input is skipped. That's the formulation shellcheck tends to push hard on.

As for the directory glob I'm happy to wait 'til we need to implement.

.github/workflows/release-trigger.yaml
vcschapp: Done.

.github/actions/create-package-release/action.yml
vcschapp: Done.

.github/actions/create-package-release/action.yml
vcschapp: Fixed.

.github/workflows/release-trigger.yaml
vcschapp: True. I thought about this and got halfway into impl, but in the end landed on:

- We already have `matrix.package == 'overture-schema'` check for the vanity tag and the argument's equally applicable there.
- For now it feels like a YAGNI thing - I'd rather refactor when we need to.

.github/workflows/release-trigger.yaml
vcschapp: Done.

.github/actions/create-package-release/action.yml
vcschapp: Done.

.github/actions/create-package-release/action.yml
vcschapp: Refactored to read it into an array so there's no implicit globbing being executed.

.github/workflows/release-trigger.yaml
vcschapp: Agree it'd be good but this feels like an enhancement. Suggest we defer the attestation.

lowlydba (approved):

.github/actions/create-package-release/action.yml
```diff
@@ -47,6 +47,13 @@ inputs:
   token:
     description: "Token used to create the release (needs contents: write)."
     required: true
+  assets:
+    description: >
+      Optional whitespace-separated list of asset files to attach to the
```
danabauer, line 52: why would we need a list of assets here?

danabauer: ah, ok, this is a way for users to request which assets they want, yes?

.github/workflows/release-trigger.yaml
```diff
@@ -112,6 +127,8 @@ jobs:
           # The umbrella package continues the legacy bare tag series as a
           # vanity tag (no second release); see docs/versioning.md.
           vanity-tag: ${{ matrix.package == 'overture-schema' && format('v{0}', matrix.version) || '' }}
+          # Only the umbrella release carries the unified JSON schema artifact.
```
danabauer, line 130: otherwise users would get a separate json file for each package?

docs/versioning.md
```diff
@@ -88,6 +88,15 @@ attached. The umbrella package is the primary entrypoint for most consumers,
...
 use only the package-prefixed scheme.
 
+The umbrella `overture-schema` release additionally carries **`overture-schema.json`**
+as a release asset. This is the Pydantic-generated unified JSON Schema for the
+whole official Overture schema, generated at release time by
+`$ overture-schema json-schema --tag overture`. Including `overture-schema.json`
+as a release artifact provides a bridge for any consumers who were using the
+now-deprecated hand-authored YAML-mastered JSON Schema under `schema/` and want
+to continue consuming JSON Schema. We attach this artifact only to the umbrella
+`overture-schema` release, not for any other package.
```
danabauer, line 98: ok, so, there's no mechanism for getting a per-package json schema

.github/workflows/release-trigger.yaml
vcschapp: That's an alternative implementation, but we decided against it at the last coding sesh.

docs/versioning.md
vcschapp: Exactly.

.github/actions/create-package-release/action.yml
vcschapp: No, this is driven by the release automation.

The `assets` parameter is general-purpose meaning we could do more with it later, but for now it works like the diagram below. Basically, if the release automation is creating a release for the `overture-schema` package, then it generates `overture-schema.json` and attaches it to the release; otherwise `assets` is blank and nothing gets attached to the release.

2026-09-01-overture-schema-PR-708-assets-populate-flow

danabauer: Got it. Thanks.

danabauer (approved):

PR #714, opened 2026-09-02
chore: detect vnext release merge by ancestry, not branch name

### LINKED ISSUE #713: Generalize vnext release-merge detection past head_ref == 'vnext'
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

dev ops

### Description

OvertureMaps/schema#709 stages the 2.0.0 release on a `2_0_0` branch instead of merging `vnext` itself, so it merges to `main` with `head_ref == '2_0_0'`. Both `.github/workflows/vnext-compat.yaml` and `.github/workflows/rebase-vnext.yaml` only special-case `head_ref == 'vnext'`, so neither recognizes this as the release merge: `vnext-compat` runs its normal compat check against a PR that's already past that point, and `rebase-vnext` would try to rebase `vnext` onto a `main` it's already an ancestor of.

Per the discussion on OvertureMaps/schema#709, formalize this as a supported path: detect the release merge by ancestry (`git merge-base --is-ancestor origin/vnext <head>`) instead of by branch name, so any staging branch built on top of `vnext`'s tip is recognized, not just a PR literally named `vnext`. Update both workflows and document the branch + merge-commit requirement in `CONTRIBUTING.md`.

Squash stays the default everywhere else; only the release path itself needs the ancestry check and a real merge commit.

lowlydba:
## Context

OvertureMaps/schema#709 discussed two ways to make its `2_0_0` staging branch a supported release path instead of a one-off. This picks option 2: keep the staging branch and formalize the detection.

Fixes OvertureMaps/schema#713

## What changed

`vnext-compat.yaml` and `rebase-vnext.yaml` both special-cased `head_ref == 'vnext'` to skip their normal checks for the vnext release merge. That misses OvertureMaps/schema#709, whose head is `2_0_0`, a branch built on top of `vnext`'s tip rather than `vnext` itself.

Both workflows now detect the release merge by ancestry instead: `git merge-base --is-ancestor origin/vnext <head>`. If `vnext` is already an ancestor of the PR head (`vnext-compat`) or the new `main` HEAD (`rebase-vnext`), the branch already carries everything on `vnext`, so it's treated as the release merge regardless of its name.

```mermaid
%%{init: {"theme": "dark", "flowchart": {"padding": 14}, "themeVariables": {"fontSize": "14px", "mainBkg": "#21262d", "nodeBorder": "#4493f8"}}}%%
flowchart TD
    classDef default rx:8,ry:8,stroke-width:0.75px
    subgraph compat["vnext-compat.yaml (on PR)"]
        A["PR targets main"] --> B{"origin/vnext ancestor<br/>of PR head?"}
        B -- yes --> C["noop:<br/>release merge, skip"]
        B -- no --> D["compat-check:<br/>simulate squash + rebase vnext"]
    end
    subgraph rebase["rebase-vnext.yaml (on push to main)"]
        E["Push to main"] --> F{"origin/vnext ancestor<br/>of new main HEAD?"}
        F -- yes --> G["skip:<br/>release merge, no rebase"]
        F -- no --> H["rebase vnext onto main,<br/>force-push"]
    end
```

`vnext-compat.yaml` needed a new `detect-release` job up front since a job's `if:` can't run shell commands; `compat-check` and `noop` now key off its `is_release` output. `rebase-vnext.yaml` just moved its checkout earlier so the existing detection step can run the ancestry check locally instead of only inspecting the merged PR's `head.ref` from the GitHub API.

`CONTRIBUTING.md` now documents the staging-branch pattern as a supported alternative to merging `vnext` directly.

## Testing

No CI-only path here to exercise outside the real workflows: I validated both YAML files parse (`yaml.safe_load`) and traced the `if:`/`needs:` wiring by hand, but haven't watched the ancestry check run against a real PR. Worth watching `vnext-status` on this PR itself once it opens against `main`, and watching `rebase-vnext` on the next `2_0_0`-style release merge.

### DIFF
--- a/.github/workflows/rebase-vnext.yaml
+++ b/.github/workflows/rebase-vnext.yaml
@@ -3,8 +3,11 @@ name: Rebase vnext onto main
 # Triggered on every push to main. Force-rebases vnext onto the new main HEAD
 # using the overture-pull-requester GitHub App (which has branch-protection bypass).
 #
-# Skipped for vnext→main release merges — vnext is already equal to main at that
-# point so a rebase would be a no-op, and the GitHub API check catches this.
+# Skipped for a vnext release merge — detected by ancestry (origin/vnext is
+# already an ancestor of the new main HEAD), not by branch name, so a staging
+# branch built on top of vnext's tip (e.g. `2_0_0`) is skipped the same as a
+# PR literally named `vnext`. vnext is already equal to main at that point so
+# a rebase would be a no-op.
 #
 # If the rebase fails a GitHub issue is opened and assigned to the PR author
 # so the conflict can be resolved manually.
@@ -41,37 +44,40 @@ jobs:
           permission-pull-requests: read    # read PR on merge commit to detect vnext→main release
           permission-workflows: write       # vnext commits may touch .github/workflows/**
 
-      # Detect whether this push was a vnext→main release merge.
-      # The GitHub API returns the PR(s) associated with the merge commit.
-      - name: Detect vnext→main release merge
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          # Use the app token so the subsequent force-push is authenticated.
+          token: ${{ steps.app-token.outputs.token }}
+          persist-credentials: true
+
+      # Detect whether this push was a vnext release merge, by ancestry rather
+      # than branch name: if origin/vnext is already an ancestor of the new
+      # main HEAD, all of vnext's commits are in main and rebasing is a no-op.
+      # The GitHub API additionally returns the PR associated with the merge
+      # commit, so a failure issue can still be filed against its author.
+      - name: Detect vnext release merge
         id: skip
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           MAIN_SHA: ${{ github.sha }}
         run: |
+          git fetch origin vnext
+
           PR_JSON=$(gh api "repos/${GITHUB_REPOSITORY}/commits/${MAIN_SHA}/pulls" \
             --jq '.[0] // empty' 2>/dev/null || true)
 
-          HEAD_REF=$(echo "$PR_JSON"  | jq -r '.head.ref   // empty' 2>/dev/null || true)
           PR_NUMBER=$(echo "$PR_JSON" | jq -r '.number     // empty' 2>/dev/null || true)
           PR_AUTHOR=$(echo "$PR_JSON" | jq -r '.user.login // empty' 2>/dev/null || true)
 
-          if [ "$HEAD_REF" = "vnext" ]; then
-            echo "Skipping: this is a vnext→main release merge."
+          if git merge-base --is-ancestor origin/vnext "$MAIN_SHA"; then
+            echo "Skipping: origin/vnext is already an ancestor of main, this is a release merge."
             echo "skip=true" >> "$GITHUB_OUTPUT"
           else
             echo "skip=false"      >> "$GITHUB_OUTPUT"
             echo "pr_number=${PR_NUMBER}" >> "$GITHUB_OUTPUT"
             echo "pr_author=${PR_AUTHOR}" >> "$GITHUB_OUTPUT"
           fi
-
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        if: steps.skip.outputs.skip != 'true'
-        with:
-          fetch-depth: 0
-          # Use the app token so the subsequent force-push is authenticated.
-          token: ${{ steps.app-token.outputs.token }}
-          persist-credentials: true
 
       - name: Rebase vnext onto new main HEAD
         if: steps.skip.outputs.skip != 'true'
--- a/.github/workflows/rebase-vnext.yaml
+++ b/.github/workflows/rebase-vnext.yaml
@@ -71,7 +71,7 @@ jobs:
           PR_AUTHOR=$(echo "$PR_JSON" | jq -r '.user.login // empty' 2>/dev/null || true)
 
           if git merge-base --is-ancestor origin/vnext "$MAIN_SHA"; then
-            echo "Skipping: origin/vnext is already an ancestor of main, this is a release merge."
+            echo "::notice::Skipping rebase: origin/vnext is already an ancestor of main, this is a release merge."
             echo "skip=true" >> "$GITHUB_OUTPUT"
           else
             echo "skip=false"      >> "$GITHUB_OUTPUT"
--- a/.github/workflows/vnext-compat.yaml
+++ b/.github/workflows/vnext-compat.yaml
@@ -194,7 +194,7 @@ jobs:
     if: github.event.pull_request.base.ref != 'main' || needs.detect-release.outputs.is_release == 'true'
     runs-on: ubuntu-slim
     steps:
-      - run: echo "Base isn't main, or this is the vnext release merge — nothing to check here."
+      - run: echo "::notice::Base isn't main, or this is the vnext release merge — nothing to check here."
 
   vnext-status:
     name: vnext compatibility status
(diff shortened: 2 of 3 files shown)


.github/workflows/vnext-compat.yaml
```diff
@@ -23,9 +26,36 @@ concurrency:
...
+        id: check
+        env:
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          git fetch origin vnext
+          if git merge-base --is-ancestor origin/vnext "$PR_HEAD_SHA"; then
+            echo "is_release=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_release=false" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
lowlydba: Good catch, fixed. `detect-release` now runs unconditionally and short-circuits to `is_release=false` for non-`main` bases (skipping the checkout entirely), so `noop`/`compat-check` never lose their dependency to skip propagation.

sethfitz (approved):

.github/workflows/rebase-vnext.yaml
```diff
@@ -41,38 +44,41 @@ jobs:
...
             --jq '.[0] // empty' 2>/dev/null || true)
 
-          HEAD_REF=$(echo "$PR_JSON"  | jq -r '.head.ref   // empty' 2>/dev/null || true)
           PR_NUMBER=$(echo "$PR_JSON" | jq -r '.number     // empty' 2>/dev/null || true)
           PR_AUTHOR=$(echo "$PR_JSON" | jq -r '.user.login // empty' 2>/dev/null || true)
 
-          if [ "$HEAD_REF" = "vnext" ]; then
-            echo "Skipping: this is a vnext→main release merge."
+          if git merge-base --is-ancestor origin/vnext "$MAIN_SHA"; then
+            echo "::notice::Skipping rebase: origin/vnext is already an ancestor of main, this is a release merge."
```
vcschapp, line 74: I suggest we drop the `skip` step because if `vnext` is an ancestor of `main`, the rebase should always be a clean fast-forward. Skipping doesn't really buy us anything and adds complexity.

----

I think it's probably wrong also, in that after this change, the rebase won't run on every merge to `main`, it'll only run when `vnext` contains at least one commit that's not on `main`. This will mean that if a sequence of PRs target `main` then `main` will drift further and further ahead of `vnext`.

Basically on the old system if you pushed a commit `B` on `main`  after a release (`r`), `vnext` would be rebased up to `A`:

```
Starting point after release `r` (old system):

   A ─── r        main
    \
     v            vnext  (behind r)

Push a new commit `B` to `main` and `vnext` is fast-forwarded up to `B`:

   A ─── r ─── B     main
              ╱
     v ──────╯      vnext
```

In the new system if this happens, then `vnext` is an ancestor of `B` so there's no fast-forward...

```
   A ─── r ─── B     main
    \
     v               vnext  (behind r)
```

And this would continue if you keep tacking on new commits `C`, `D`, *etc.*

```
   A ─── r ─── B ─── C ─── D  main
    \
     v                        vnext (drifting further behind, one push at a time...)

```

---

If we just get rid of the skip and always run the fast-forward I think we get rid of this problem, yes?


(*Note: This is a Claude-detected bug but it makes sense to me when I play it through.*)

.github/workflows/vnext-compat.yaml
```diff
@@ -160,10 +199,11 @@ jobs:
...
-    if: github.event.pull_request.base.ref != 'main' || github.head_ref == 'vnext'
+    needs: detect-release
+    if: github.event.pull_request.base.ref != 'main' || needs.detect-release.outputs.is_release == 'true'
     runs-on: ubuntu-slim
     steps:
-      - run: echo "Base isn't main, or this is the vnext release PR — nothing to check here."
+      - run: echo "::notice::Base isn't main, or this is the vnext release merge — nothing to check here."
 
   vnext-status:
     name: vnext compatibility status
```
vcschapp, line 209: I am informed by my LLM overlords that this will fail open if `detect-release` fails resulting in an otherwise non-mergeable change being allowed to merge.

The mechanism is apparently this dependency graph

```
vnext-status ──▶ are-we-good ──▶ compat-check ──▶ detect-release
                      │                                 ▲
                      └────────▶ noop ──────────────────┘
```

The claim is that GitHub action dependency propagation causes `compat-check` to be skipped if `detect-release` fails; and that `are-we-good` treats a skip (`compat-check` in this case) as if it was a pass.

Can you double-check this?

.github/workflows/vnext-compat.yaml
```diff
@@ -23,9 +26,45 @@ concurrency:
...
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          PR_BASE_REF: ${{ github.event.pull_request.base.ref }}
+        run: |
+          if [ "$PR_BASE_REF" != "main" ]; then
+            echo "is_release=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          git fetch origin vnext
+          if git merge-base --is-ancestor origin/vnext "$PR_HEAD_SHA"; then
```
vcschapp, line 58: [2/2] LLM overlord is reporting that `origin/vnext` may not be populated after `git fetch origin vnext` due to Git quirks. Apparently an alternative that is certain to work/can't break unexpectedly is to refer to `FETCH_HEAD` instead of `origin/vnext`.

.github/workflows/rebase-vnext.yaml
```diff
@@ -41,38 +44,41 @@ jobs:
...
           PR_JSON=$(gh api "repos/${GITHUB_REPOSITORY}/commits/${MAIN_SHA}/pulls" \
             --jq '.[0] // empty' 2>/dev/null || true)
 
-          HEAD_REF=$(echo "$PR_JSON"  | jq -r '.head.ref   // empty' 2>/dev/null || true)
           PR_NUMBER=$(echo "$PR_JSON" | jq -r '.number     // empty' 2>/dev/null || true)
           PR_AUTHOR=$(echo "$PR_JSON" | jq -r '.user.login // empty' 2>/dev/null || true)
 
-          if [ "$HEAD_REF" = "vnext" ]; then
-            echo "Skipping: this is a vnext→main release merge."
+          if git merge-base --is-ancestor origin/vnext "$MAIN_SHA"; then
```
vcschapp, line 73: [1/2] LLM overlord is reporting that `origin/vnext` may not be populated after `git fetch origin vnext` due to Git quirks. Apparently an alternative that is certain to work/can't break unexpectedly is to refer to `FETCH_HEAD` instead of `origin/vnext`.

.github/workflows/vnext-compat.yaml
```diff
@@ -23,9 +26,45 @@ concurrency:
...
+            echo "is_release=true" >> "$GITHUB_OUTPUT"
```
vcschapp, line 59: Same root cause as [the comment on the skip issue](https://github.com/OvertureMaps/schema/pull/714/changes#r3919021495), the `--is-ancestor` check will result in `is_release=true` during any period when `vnext` hasn't diverged from `main`, even though most of those occurrences wouldn't be considered releases.

My understanding of what we're trying to do here is that `is_release` is the workflow's way of answering the question, *is this PR the one that ships accumulated breaking changes into the main line, rather than an ordinary change?* And we want to know this because in the case where we were merging in a bunch of breaking changes from `vnext`, the compatibility check really would fail because it'd be replaying `vnext`'s commits on top of itself.

Current check is saying "if `vnext` is an ancestor of `main` it's a release" which includes the case where `main` is just organically ahead of `vnext` because there are no new breaking changes, so it'll be true too often. If we exclude that case then I think it works:

```bash
git fetch --no-tags origin \
  '+refs/heads/main:refs/remotes/origin/main' \ # This +refs business is meant to ensure
  '+refs/heads/vnext:refs/remotes/origin/vnext' # you can validly refer to `origin/vnext and origin/main`

if   git merge-base --is-ancestor origin/vnext "$PR_HEAD_SHA" \
&& ! git merge-base --is-ancestor origin/vnext origin/main; then
  echo "is_release=true"  >> "$GITHUB_OUTPUT"
else
  echo "is_release=false" >> "$GITHUB_OUTPUT"
fi
```

vcschapp (commented):
Claude found a few issues. It took me a while to understand them and recast them in words that are more human-oriented and understandable, but here they are!

Issue #674, opened 2026-08-19
[FEATURE] Carry deprecation through extraction and render it in the Markdown reference


sethfitz:
Sub-issue of #639 — surface 1 of 3 ("What flows where"), and the one we need first.

## Scope

Carry the deprecation message from the Pydantic models through extraction, and render it in the generated Markdown reference. Self-contained and shippable on its own: it delivers the reference-doc value without touching validation or the changelog.

The extraction carrier lands here because Markdown is its first consumer. The PySpark surface (#676) and the deprecation manifest (#677) both read the same carrier, so this issue is their prerequisite.

## Marking mechanism

Native Python, no custom marker:

- **Fields** — Pydantic's `Field(deprecated=...)`, which accepts `True`, a message string, or a PEP 702 `deprecated(...)` marker (also usable directly in `Annotated` metadata). All forms normalize to `FieldInfo.deprecation_message` (`str | None`).
- **Models** — PEP 702's `@deprecated("message")` (`typing_extensions.deprecated` on our 3.10 floor), which sets `__deprecated__` on the class. Read with `getattr(cls, "__deprecated__", None)`.

```python
old_field: str | None = Field(
    default=None,
    deprecated="Use `new_field` instead. Deprecated in v1.18.0.",
)

@deprecated("Use `NewFeature` instead. Deprecated in v1.18.0.")
class OldFeature(Feature):
    ...
```

## Work

- Add `deprecated: str | None` to `FieldSpec` and to the model/record spec.
- Populate both in `model_extraction.py`: `field_info.deprecation_message` for fields, `getattr(model_class, "__deprecated__", None)` for the model.
- Field-level rendering: append a `**Deprecated:** <message>` note to the field's Description cell through the existing constraint-note appender, and tag the Type column with `(deprecated)` alongside `(optional)`.
- Model-level rendering: a `**Deprecated:** <message>` banner at the top of the feature page.
- Mark one field and one model deprecated in a test fixture (not in the published schema) and assert the rendered output.

## Out of scope

Enum members and bare type aliases (`Annotated[...]`, `NewType`) have no native `@deprecated` hook and would need a project-specific convention. Deferred — see the open question on #639.

## Acceptance

- A field marked `Field(deprecated=...)` in any of the three accepted forms renders its message in the reference.
- A class marked `@deprecated(...)` renders a banner on its feature page.
- `deprecated=True` (bare) renders the generic `deprecated` message rather than dropping the signal.
[...]

PR #724, opened 2026-09-04
[FEATURE](codegen) Add a stac-table-columns codegen format

### LINKED ISSUE #723: Add stac-table-columns renderer to overture-schema-codegen
### Type

Task

### Scope

Multi-theme or Platform

### Skillset

engineering

## Context

The STAC catalog at stac.overturemaps.org emits `table:columns` with only column names — no `type` or `description`. Consumers reading the catalog have no way to discover what a column means without leaving STAC and going to the schema docs. The information already lives in the Pydantic models; a new codegen renderer can surface it as machine-readable JSON that the STAC generator consumes at build time.

## Proposed change

In [packages/overture-schema-codegen/src/overture/schema/codegen/](https://github.com/OvertureMaps/schema/tree/main/packages/overture-schema-codegen/src/overture/schema/codegen), alongside [`markdown/`](https://github.com/OvertureMaps/schema/tree/main/packages/overture-schema-codegen/src/overture/schema/codegen/markdown) and [`pyspark/`](https://github.com/OvertureMaps/schema/tree/main/packages/overture-schema-codegen/src/overture/schema/codegen/pyspark):

- Add a `stac_table_columns/` renderer that emits one JSON file per (theme, type), matching the [STAC table extension](https://github.com/stac-extensions/table) `table:columns` shape: `[{"name", "type", "description"}]`.
- `type` maps from the Pydantic field's resolved primitive to STAC/parquet types (`string`, `int32`, `int64`, `double`, `boolean`, `struct`, `list`). Reuse extraction logic already used by the pyspark renderer where possible.
- `description` comes from the field's docstring / description text.
- Wire into the CLI as a new `--format` option so `overture-codegen generate --format stac-table-columns --output-dir out/` writes `buildings/building.json`, `places/place.json`, etc.

sethfitz:
# Description

Adds a `stac-table-columns` codegen format that renders the STAC Table extension's `table:columns` from the Pydantic models, one STAC properties fragment per feature type — 15 in all. It targets **table extension v1.3.0**, released 2026-09-07.

The models already hold everything a Column Object asks for, so the catalog is a projection with no authoring step. v1.3.0's Column Object is `name`, `type` and `description`, plus what it newly admits from STAC common metadata (`data_type`, `unit`) and from the Vector extension (`vector:geometry_types`). Everything else the IR carries — constraints, optionality, defaults, nested field descriptions, enum vocabularies — still has nowhere to go, so the renderer's second output is a gap log that records each of those as it is dropped.

One fragment in full, from `overture-codegen generate --format stac-table-columns --filter 'overture:theme=buildings'` (which emits `building` and `building_part`). It shows the Arrow type strings including the nested `sources` and `names` composites, a `data_type` on every column, `vector:geometry_types` read off the geometry constraint, and `table:primary_geometry` naming the one geometry column.

<details>
<summary><code>building</code> — 26 columns</summary>

```json
{
  "stac_extensions": [
    "https://stac-extensions.github.io/table/v1.3.0/schema.json",
    "https://stac-extensions.github.io/vector/v0.1.0/schema.json"
  ],
  "properties": {
    "table:columns": [
      {
        "name": "id",
        "type": "string",
        "description": "A feature ID. This may be an ID associated with the Global Entity Reference System (GERS) if—and-only-if the feature represents an entity that is part of GERS.",
        "data_type": "other"
      },
      {
        "name": "bbox",
        "type": "struct<xmin: double, ymin: double, xmax: double, ymax: double>",
        "description": "An optional bounding box for the feature",
        "data_type": "other"
      },
      {
        "name": "geometry",
        "type": "binary",
        "description": "The building's footprint or roofprint (if traced from aerial/satellite imagery).",
        "data_type": "other",
        "vector:geometry_types": [
          "MultiPolygon",
          "Polygon"
        ]
      },
      {
        "name": "theme",
        "type": "string",
        "data_type": "other"
      },
      {
        "name": "type",
        "type": "string",
        "data_type": "other"
      },
      {
        "name": "version",
        "type": "int32",
        "data_type": "int32"
      },
      {
        "name": "sources",
        "type": "list<item: struct<property: string, dataset: string, license: string, record_id: string, update_time: timestamp[us, tz=UTC], confidence: double, provider: string, resource: string, version: string, between: list<item: double>>>",
        "description": "Information about the source data used to assemble the feature.",
        "data_type": "other"
      },
      {
        "name": "subtype",
        "type": "string",
        "description": "A broad classification of the current use and purpose of the building.\n\nIf the current use of the building no longer accords with the original built\npurpose, the current use should be specified. For example, a building built as a\ntrain station but later converted into a shopping mall would have the value\n`\"commercial\"` rather than `\"transportation\"`.",
        "data_type": "other"
      },
      {
        "name": "class",
        "type": "string",
        "description": "A more specific classification of the current use and purpose of the building.\n\nIf the current use of the building no longer accords with the original built\npurpose, the current use should be specified.",
        "data_type": "other"
      },
      {
        "name": "has_parts",
        "type": "bool",
        "description": "Whether the building has associated building part features",
        "data_type": "other"
      },
      {
        "name": "names",
        "type": "struct<primary: string, common: map<string, string>, rules: list<item: struct<value: string, variant: string, language: string, perspectives: struct<mode: string, countries: list<item: string>>, between: list<item: double>, side: string>>>",
        "description": "All known names by which the feature is called",
        "data_type": "other"
      },
      {
        "name": "level",
        "type": "int32",
        "description": "Z-order of the feature where 0 is visual level. A feature without a level is at visual level.",
        "data_type": "int32"
      },
      {
        "name": "height",
        "type": "double",
        "description": "Height of the building or part in meters.\n\nThis is the distance from the lowest point to the highest point.",
        "data_type": "float64"
      },
      {
        "name": "is_underground",
        "type": "bool",
        "description": "Whether the entire building or part is completely below ground.\n\nThe underground flag is useful for display purposes. Buildings and building parts\nthat are entirely below ground can be styled differently or omitted from the\nrendered image.\n\nThis flag is conceptually different from the `level` field, which indicates\nrelative z-ordering and, notably, can be negative even if the building is entirely\nabove-ground.",
        "data_type": "other"
      },
      {
        "name": "num_floors",
        "type": "int32",
        "description": "Number of above-ground floors of the building or part.",
        "data_type": "int32"
      },
      {
        "name": "num_floors_underground",
        "type": "int32",
        "description": "Number of below-ground floors of the building or part.",
        "data_type": "int32"
      },
      {
        "name": "min_height",
        "type": "double",
        "description": "Altitude above ground where the bottom of the building or building part starts.\n\nIf present, this value indicates that the lowest part of the building or building\npart starts is above ground level.",
        "data_type": "float64"
      },
      {
        "name": "min_floor",
        "type": "int32",
        "description": "Start floor of this building or part.\n\nIf present, this value indicates that the building or part is \"floating\" and its\nbottom-most floor is above ground level, usually because it is part of a larger\nbuilding in which some parts do reach down to ground level. An example is a building\nthat has an entry road or driveway at ground level into an interior courtyard, where\npart of the building bridges above the entry road. This property may sometimes be\npopulated when `min_height` is missing and in these cases can be used as a proxy for\n`min_height`.",
        "data_type": "int32"
      },
      {
        "name": "facade_color",
        "type": "string",
        "description": "Facade color in `#rgb` or `#rrggbb` hex notation",
        "data_type": "other"
      },
      {
        "name": "facade_material",
        "type": "string",
        "description": "Outer surface material of the facade",
        "data_type": "other"
      },
      {
        "name": "roof_material",
        "type": "string",
        "description": "Outer surface material of the roof",
        "data_type": "other"
      },
      {
        "name": "roof_shape",
        "type": "string",
        "description": "Shape of the roof",
        "data_type": "other"
      },
      {
        "name": "roof_direction",
        "type": "double",
        "description": "Bearing of the roof ridge line in degrees",
        "data_type": "float64"
      },
      {
        "name": "roof_orientation",
        "type": "string",
        "description": "Orientation of the roof shape relative to the footprint shape",
        "data_type": "other"
      },
      {
        "name": "roof_color",
        "type": "string",
        "description": "The roof color in `#rgb` or `#rrggbb` hex notation",
        "data_type": "other"
      },
      {
        "name": "roof_height",
        "type": "double",
        "description": "Height of the roof in meters.\n\nThis is the distance from the base of the roof to its highest point.",
        "data_type": "float64"
      }
    ],
    "table:primary_geometry": "geometry"
  }
}
```

</details>

Four decisions are explicit in the code, because nothing downstream can check them.

**Column types use Arrow's names** — `int64`, `string`, `double`, `binary`, `struct<a: int32, b: string>`, `list<item: int32>`. `type` is an unconstrained string, and real catalogs carry whatever their generating engine's stringifier produced; stac-extensions/table#1, "Firm up the definition of `ColumnObject.type`", has been open since 2021, lists three options and picks none. Arrow is what Overture distributes — the release is GeoParquet, and these are the names an Arrow reader reports on opening it — and it matches published catalogs, which name columns `int64`, `string`, `double`, `binary`. The vocabulary and the composite grammar both come from `pyarrow`'s own stringifier rather than from memory.

Struct **member order** is the model's declaration order. For `bbox` that does not match what Overture ships — released files order it `xmin, xmax, ymin, ymax`, as does this repo's PySpark `BBOX_STRUCT` — and the discrepancy is accepted rather than encoded: a constant matching one publisher's file layout would state a fact about that publisher rather than about the schema. It is recorded in the constant's comment and pinned by a test, so it reads as a decision rather than an oversight.

`type` and `data_type` consequently disagree on floats, and that is deliberate rather than an oversight. `data_type` is bound to STAC common metadata's fixed vocabulary, which names them `float32`/`float64`; `type` is bound to Arrow, which names the same two `float`/`double`. Harmonising them would make one field lie about the vocabulary it claims to speak. A test pins the mismatch so a later reader finds it recorded rather than "fixes" it.

**`data_type` is Overture's own base-type name, not a translation.** STAC common metadata's data-type vocabulary (`int8`…`int64`, `uint8`…`uint64`, `float32`/`float64`, `other`) already matches Overture's numeric base-type names verbatim, so this is an allowlist against that vocabulary rather than a lookup table. Every column gets one — `other` is a member of the vocabulary, not an omission — so most Overture columns resolve to `other`, and only the numerics carry signal.

**Flattening a discriminated union drops information, and every drop is logged.** In `segment`, `RoadClass` and `RailClass` each contribute a `class` column; a flat column list has one slot for it, so one description is kept and the other disappears from the output. This happens seven times in `segment`. Nothing catches it on its own: the alternatives have identical physical types, so every compatibility check passes and the output looks complete. The renderer emits a `flatten-collision` gap at each site, the only record that something was dropped.

**Ambiguity is refused, and `vector:geometry_types` is the one case that looks like an exception and is not.** `table:primary_geometry` is emitted when exactly one geometry column exists and omitted, with the omission logged, when two do — nothing in the Pydantic model indicates which is primary. `table:row_count` is never emitted: a row count is a property of data, and this path has only a schema. A recursive model raises, because a type string cannot name a type and so a cycle cannot terminate. Union alternatives resolving to different physical types raise.

`vector:geometry_types` is defined as the geometry types *present in the dataset*, which by that same rule is a data fact this renderer cannot measure. It is emitted anyway, because it is read from `GeometryTypeConstraint`, which is validated on every row and is therefore an exhaustive *bound* on what a column can contain rather than a guess about what it does. A geometry column with no such constraint gets neither the field nor a claim about it — logged as a gap instead. Because the field belongs to another extension, the fragment declares the Vector extension in `stac_extensions` exactly when a column emits it, or it would not validate.

`unit` is **not** implemented, because Overture Schema has no consistent way to encode one. Where a unit varies per row it is a field — `Speed` carries a `SpeedUnit`, so a single `unit` string on that column would assert every row shares a unit when the schema allows mixing `mph` and `km/h`. Where a unit is fixed (`Height`, `Elevation`, `Depth`, roof bearing) it appears only in a free-text description. Neither is a column-level fact, and this renderer does not parse prose.

Two commits here stand on their own. The first: extraction now carries `default` and `deprecated` through to `FieldSpec`, which addresses #674. The last is a bug in the existing writer, not in this format — `generate --output-dir` wrote through `Path.write_text` with no `encoding`, so on a non-UTF-8 locale it raised `UnicodeEncodeError` on the first em-dash in a description and wrote nothing. The markdown format has been reachable by that all along; generated files are now written as UTF-8.

Closes #723.

# Reference

1. #723 — the issue this implements
2. https://github.com/stac-extensions/table — the extension being targeted
3. https://github.com/stac-extensions/table/releases/tag/v1.3.0 — the release this targets, which admits STAC common metadata on the Column Object and recommends `vector:geometry_types` on geometry columns
4. https://github.com/stac-extensions/vector — the extension `vector:geometry_types` comes from
5. https://github.com/stac-extensions/table/issues/1 — the open thread on what `type` means
6. #674 — wants the same `deprecated` carry the first commit adds

# Checklist

1. [ ] Add relevant examples. — n/a, no schema change
2. [ ] Add relevant counterexamples. — n/a, no schema change
3. [ ] Update any counterexamples that became obsolete. — n/a, no schema change
4. [ ] Update in-schema documentation using plain English written in complete sentences, if an update is required. — n/a, no schema change
5. [ ] Update Docusaurus documentation, if an update is required. — not done; would follow if this stops being a draft
6. [ ] Review change with Overture technical writer to ensure any advanced documentation needs will be taken care of, unless the change is trivial and would not affect the documentation. — n/a, codegen-internal


packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -0,0 +1,619 @@
...
+    (_dt.date, "date"),
+    (str, "varchar"),
+    (bytes, "blob"),
+    (int, "bigint"),
+    (float, "double"),
+)
+
+
+@dataclass
+class Column:
```
jjcfrancisco, line 186: With v1.3.0, we should add a `data_type` field here alongside `type`, plus a small mapping ([spec vocab](https://github.com/radiantearth/stac-spec/blob/master/commons/common-metadata.md#data-types)). Although, a lot of columns will resolve to `"other"` for Overture, only numerics like `level` carry real signal.

We should also add [`geometry_type`](https://github.com/stac-extensions/table#column-object) (new in v1.3.0) for the geometry column. Overture already knows whether each feature type is `Point` / `LineString` / `Polygon`.

  And [`unit`](https://github.com/radiantearth/stac-spec/blob/master/commons/common-metadata.md) from common metadata (now permitted on Column Object in v1.3.0). Overture carries real units like `km/h` or `mph` for speed limits

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -0,0 +1,619 @@
...
+    "TABLE_EXTENSION_URI",
+    "TableColumnsDocument",
+    "render_table_columns",
+]
+
+TABLE_EXTENSION_URI = "https://stac-extensions.github.io/table/v1.2.0/schema.json"
+
+# Lowercase DuckDB SQL, the dialect the Portolan spec repo's own generated
+# reference catalog emits. See the module docstring: this is a choice.
+_DUCKDB_TYPES: dict[str, str] = {
```
jjcfrancisco: Despite DuckDB being the only executable example in the STAC repo, I'd argue for Arrow.

  **Prior art in the wild.** I (randomly) picked three published Portolan catalogs:

  - [Madrid: defunciones-barrio-2025](https://browser.portolan-sdi.org/#/external/storage.googleapis.com/carto-open-catalogs/madrid-datos-abiertos/main/demografia/defunciones-barrio-2025/collection.json)
  - [Microsoft ML road detections](https://browser.portolan-sdi.org/#/external/data.source.coop/nlebovits/microsoft-ml-road-detections/road-detections/collection.json)
  - [TriMet district boundary](https://browser.portolan-sdi.org/#/external/data.source.coop/cholmes/trimet/district-boundary/collection.json)

  All three have Arrow flavored primitives (`int64`, `string`, `double`, `binary`) although slightly diverge on struct grammar. It seems a little bit wild at the moment.

Even the  STAC's own vocab is already Arrow(ish), the `data_type` list in [STAC common metadata](https://github.com/radiantearth/stac-spec/blob/master/commons/common-metadata.md#data-types)

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -0,0 +1,619 @@
...
+from ..extraction.specs import NO_DEFAULT, FieldSpec, ModelSpec, UnionSpec
+from .exceptions import Kind, TableColumnsGap, TableColumnsUnrepresentable
+
+__all__ = [
+    "TABLE_EXTENSION_URI",
+    "TableColumnsDocument",
+    "render_table_columns",
+]
+
+TABLE_EXTENSION_URI = "https://stac-extensions.github.io/table/v1.2.0/schema.json"
```
jjcfrancisco: This should point to the new [v1.3.0](https://github.com/stac-extensions/table/releases/tag/v1.3.0)

jjcfrancisco (changes requested):
Looks good in general. I left few comments to add functionality due to new [v1.3.0](https://github.com/stac-extensions/table/releases/tag/v1.3.0)

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/pipeline.py
```diff
@@ -42,8 +42,13 @@ def generate_table_columns_documents(
         # A bare properties fragment rather than a whole Item: the extension
         # fields are what this target emits, and an Item would need id, geometry,
         # bbox, datetime and links, none of which come from a schema.
+        extensions = [TABLE_EXTENSION_URI]
+        if rendered.uses_vector_extension:
+            # A `vector:` field without the Vector extension declared here
+            # would not validate against either extension's schema.
+            extensions.append(VECTOR_EXTENSION_URI)
```
sethfitz: When would this not be true?

packages/overture-schema-codegen/changelog.d/723.feature.md
```diff
@@ -1 +1 @@
-Added a `stac-table-columns` codegen format rendering the STAC Table extension's `table:columns` from the models, with a gap log recording what a flat column list cannot hold. Field defaults and deprecation now carry through extraction so the renderer can report both.
+Added a `stac-table-columns` codegen format rendering the STAC Table extension's (v1.3.0) `table:columns` from the models, with a gap log recording what a flat column list cannot hold. Column types are spelled in Arrow, the vocabulary a reader of Overture's released GeoParquet reports. Each Column Object also carries STAC common metadata's `data_type` and, for the geometry column, the Vector extension's `vector:geometry_types` (declared in `stac_extensions` alongside the Table extension when emitted). Field defaults and deprecation now carry through extraction so the renderer can report both.
```
sethfitz: Avoid "spelling" / "spelled" throughout.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -1,18 +1,34 @@
 """Render a model spec to the STAC Table extension's `table:columns`.
 
-The Column Object is `name` (required), `type`, `description`. That is the whole
-target. Everything else the IR carries -- constraints, enum vocabularies, named
-types, defaults, optionality, nested field descriptions -- has nowhere to go, so
-this renderer's real output is the gap log beside the columns.
+The Column Object is `name` (required), `type`, `description`, plus what v1.3.0
+newly permits from STAC common metadata (`data_type`, `unit`) and from the
```
sethfitz: Drop "newly"; this won't age well otherwise.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -1,18 +1,34 @@
...
-downstream can falsify either:
+Four decisions are stated here rather than defaulted into, because nothing
+downstream can falsify any of them:
 
 1. **The type dialect.** `type` is an unconstrained string, and real catalogs
    carry whatever their generating engine's stringifier produced -- two dialects
-   in mixed case across the 526 columns of published catalogs surveyed. This renderer
-   emits lowercase DuckDB SQL, matching the spec repo's own generated reference
-   catalog. No validator can tell a right dialect from a wrong one.
+   in mixed case across the 526 columns of published catalogs surveyed. This
```
sethfitz: "526" isn't going to age well.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -21,6 +37,30 @@
...
+   such constraint gets neither the field nor a claim about it -- logged as a
+   gap instead, the same refusal `table:primary_geometry` makes for real
+   ambiguity. `unit` is not implemented: no Overture field carries a
+   structured, column-wide fixed unit anywhere in the IR. The one case raised
+   in review -- speed limits -- turns out to carry its unit as a per-row
+   `SpeedUnit` enum nested inside `Speed`, not a schema-fixed unit; the fields
+   that really do have one fixed unit (`Height`, `Elevation`, `Depth`, roof
+   bearing) carry it only in a free-text description ("... in meters"), which
+   this renderer will not parse to fabricate a `unit` value the schema never
+   structurally asserted.
```
sethfitz: Don't reference the review. Declare that units are omitted because Overture Schema currently has no consistent way to encode them.

In general, simplify this paragraph.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -49,57 +91,106 @@
...
-_BBOX_TYPE = "struct(xmin double, ymin double, xmax double, ymax double)"
+# The type string a shape with no Arrow type of its own collapses to.
+_STRING_TYPE = "string"
+
+# Arrow has no geometry type. GeoParquet stores geometry as WKB in a plain
+# binary column and carries the semantics in file-level metadata, so `binary`
+# is what an Arrow reader reports for Overture's own released geometry column
+# -- confirmed against the published release, and against both catalogs the
+# review cited, which spell their geometry column `binary` too. The erasure is
+# logged rather than absorbed; see `_scalar_type`.
```
sethfitz: No need to talk about verification (that's assumed) or the review that occurred.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -49,57 +91,106 @@
...
+#
+# This repo's PySpark `BBOX_STRUCT` and the codegen test rows agree. The `BBox`
+# class docstring lists `xmin, ymin, xmax, ymax`, but those are constructor
+# keywords, where order carries no meaning.
+_BBOX_TYPE = "struct<xmin: double, xmax: double, ymin: double, ymax: double>"
+
+# Pydantic types with no Arrow type of their own. Keyed by NAME rather than by
 # `issubclass`: in Pydantic v2 neither `HttpUrl` nor `EmailStr` is a `str`
 # subclass, so the builtin fallback below never reaches them -- the same trap
 # the Vecorel renderer records, found here the same way, by running it.
```
sethfitz: Don't reference the Vecorel renderer

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
```diff
@@ -224,6 +336,39 @@ def _unwrap(shape: FieldShape) -> FieldShape:
...
+        return _OTHER_DATA_TYPE
+    return _STAC_DATA_TYPES.get(base_type, _OTHER_DATA_TYPE)
+
+
+def _geometry_types_of(shape: FieldShape) -> tuple[str, ...] | None:
+    """Return the GeoJSON geometry type names a geometry column is constrained to.
+
+    Reads `GeometryTypeConstraint.allowed_types` -- see decision 4 in the
+    module docstring for why this schema constraint, and not an actual data
+    scan, is a legitimate source for `vector:geometry_types`. Returns `None`
```
sethfitz: Unnecessary to declare. Pydantic models describe the shape of the data and are not necessarily linked to a concrete dataset, so the scan can't be relied on.

packages/overture-schema-codegen/tests/test_stac_table_columns.py
```diff
@@ -147,6 +166,244 @@ class M(BaseModel):
         render_table_columns(extract_model(M), strict=True)
 
 
+def test_extension_uri_is_v1_3_0() -> None:
+    """v1.2.0 is stale; the extension has since released v1.3.0."""
+
+    assert (
+        TABLE_EXTENSION_URI
+        == "https://stac-extensions.github.io/table/v1.3.0/schema.json"
+    )
```
sethfitz: Pointless test. Scan for other pointless tests.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
sethfitz: `data_type` needed no mapping table: STAC common metadata's vocabulary (`int8`…`int64`, `uint8`…`uint64`, `float32`/`float64`, `other`) already matches Overture's own numeric base-type names, so it's an allowlist against that vocabulary rather than a translation.

One distinction: our earlier reading that there was no concrete vocabulary was about `type`, which is still unconstrained (stac-extensions/table#1 is open, lists three options, picks none). v1.3.0 draws the line explicitly: `type` takes the native name in the file format, `data_type` the standardized one from STAC common metadata. The vocabulary is stac-spec's and predates this; what's new is permission to use it on a Column Object, which v1.2.0 didn't mention at all.

`type` and `data_type` now disagree on floats. `data_type` says `float32`/`float64` because that vocabulary is fixed; `type` says `float`/`double` because those are Arrow's names. There's a test pinning the mismatch so nobody later "fixes" one into the other.

On geometry: this emits `vector:geometry_types`, read from the model's `GeometryTypeConstraint`, since that's what's declared. The released v1.3.0 defines no `geometry_type` on the Column Object — it points at the Vector extension instead.

Overture Schema doesn't define units in a way that's compatible with what STAC is looking for, so `unit` isn't implemented. I have some ideas for a consistent unit definition that could then be surfaced in both the Markdown and STAC renderers:

```python
class Unit(str, DocumentedEnum): ...
class LengthUnit(Unit): ...

Height = NewType("Height", Annotated[float64, LengthUnit.M])
Width  = NewType("Width",  Annotated[float64, LengthUnit.M])

height: Height | None = Field(default=None, description="Height of the building or part.")
```

The type names the quantity and the annotation carries the unit, so a height and a width stay distinct even though both are metres — mixing them is a type error rather than a silent bug. The wire format stays a plain double and Pydantic validation is unchanged, while the unit becomes a machine-readable fact a renderer can emit.

packages/overture-schema-codegen/src/overture/schema/codegen/stac_table_columns/renderer.py
sethfitz: Agreed, and switched — `type` is Arrow now.

STAC's `data_type` vocabulary is already Arrow-shaped, so under DuckDB `type` a column carried two forms of the same physical type (`type: "integer"`, `data_type: "int32"`) that shared no vocabulary at all. Under Arrow they converge for every primitive, diverging only where Arrow's naming is finer than STAC's (`string`/`binary` against STAC's blanket `other`).

Both the primitive names and the composite grammar come from `pyarrow`'s own stringifier rather than from memory, cross-checked against the Portolan catalogs you cited and against the Arrow schema of an actual `2026-08-19.0` release file read from S3. That third source also settled `Geometry` → `binary`: it is literally what Arrow reports for Overture's own geometry column. The semantic loss is logged as a gap, and `vector:geometry_types` carries the part the schema can assert.

One thing the switch surfaced without changing: the `bbox` struct is emitted in the `BBox` class's declaration order, `xmin, ymin, xmax, ymax`, while released Overture files order it `xmin, xmax, ymin, ymax`. A member-order constant matching one publisher's file layout would be a fact about that publisher rather than about the schema, so this renderer keeps the model's order and the discrepancy is accepted deliberately. It is recorded in the constant's comment and in a test, so it reads as a decision rather than an oversight.

One place the sources disagree. `pyarrow`'s constructor stringifies list elements as `list<item: …>`, while Arrow reading Overture's own Parquet reports `list<element: …>`. This emits `item`, because the same round-trip that renames it also renders maps as `map<string, string ('common')>` — a file artifact rather than a type name — so following "what the reader reports" consistently produces something worse. Happy to flip it if you read the precedent differently; it is one line.

Separately, `datetime` maps to `timestamp[us, tz=UTC]`, which describes the model rather than the release: no shipped Overture column is actually an Arrow timestamp (`sources[].update_time` is a `string`). That is right for a schema-derived renderer, but the "what we distribute" justification does not reach that field, so it is worth a second opinion.

PR #738, opened 2026-09-11
[BUG](system) Export Feature from the package root

### LINKED ISSUE #737: [BUG](system) Feature is not exported from overture.schema.system, unlike OvertureFeature from common
## Problem

`overture.schema.common` exports its feature base class at the package root. `overture.schema.system` does not.

```python
from overture.schema.common import OvertureFeature    # works
from overture.schema.system import Feature            # ImportError: cannot import name 'Feature'
from overture.schema.system.feature import Feature    # the only working form
```

`system/__init__.py` re-exports its submodules plus `create_model`, so the package root already carries non-module names. `Feature` is the one base class left out of that set.

AUTHORING.md uses the shorthand the export would make true:

> You don't have to build on `OvertureFeature` — subclass `system.Feature` directly for a GeoJSON-serializing model with none of the Overture conventions.

Read as an import path, `common.OvertureFeature` resolves and `system.Feature` does not. The two packages play the same role for the two base classes a third party subclasses, and answer differently.

## Fix

In `packages/overture-schema-system/src/overture/schema/system/__init__.py`:

```python
from .feature import Feature
```

plus `"Feature"` in `__all__`, mirroring `common`'s `from .feature import FeatureVersion, OvertureFeature, ThemeT, TypeT`.

## Scope

`Feature` only. `Geometry` and `BBox` sit in the same position under `geometric`, and `Id` under `ref`; whether the system root should carry those is a separate question about how wide that surface ought to be. `Feature` is the name our own authoring guidance already writes as though it were exported.

sethfitz:
# Description

`overture.schema.common` exports `OvertureFeature` at its package root. `overture.schema.system` exported only its submodules and `create_model`, so the parallel name did not resolve:

```python
from overture.schema.common import OvertureFeature  # works
from overture.schema.system import Feature  # ImportError
from overture.schema.system.feature import Feature  # the only working form
```

AUTHORING.md points third-party authors at the name that did not exist: "You don't have to build on `OvertureFeature` — subclass `system.Feature` directly for a GeoJSON-serializing model with none of the Overture conventions." Anyone taking that literally hits an ImportError on their first line.

This adds `from .feature import Feature` and `"Feature"` to `__all__`, matching how `common` exports `FeatureVersion`, `OvertureFeature`, `ThemeT` and `TypeT` from its own `feature` submodule. The deep import keeps working; nothing else changes.

Scope is `Feature` alone. `Geometry` and `BBox` under `geometric`, and `Id` under `ref`, sit in the same position — how wide the system root's surface should be is a separate question, and #737 says so.

# Reference

1. Closes #737.

# Testing

`test_feature_is_exported_from_package_root` asserts both the attribute and its presence in `__all__`. Reverting the `__init__.py` change alone fails it with `AttributeError: module 'overture.schema.system' has no attribute 'Feature'`; it is the only added test, and it is a real gate rather than a behaviour-preservation guard.

`make lint-only`, `make mypy-only`, `make doctest-only` and the full suite pass.

# Checklist

1. [ ] Add relevant examples.

# Documentation website

### DIFF
--- a/packages/overture-schema-system/changelog.d/737.bugfix.md
+++ b/packages/overture-schema-system/changelog.d/737.bugfix.md
@@ -1,0 +1,1 @@
+Export `Feature` from the `overture.schema.system` package root, so `from overture.schema.system import Feature` resolves the way `from overture.schema.common import OvertureFeature` already does.
--- a/packages/overture-schema-system/src/overture/schema/system/__init__.py
+++ b/packages/overture-schema-system/src/overture/schema/system/__init__.py
@@ -163,10 +163,12 @@ from . import (
     string,
 )
 from .create_model import create_model
+from .feature import Feature
 
 __all__ = [
     "create_model",
     "doc",
+    "Feature",
     "feature",
     "field_constraint",
     "geometric",
--- a/packages/overture-schema-system/tests/test_feature.py
+++ b/packages/overture-schema-system/tests/test_feature.py
@@ -2129,3 +2129,10 @@ class TestRefactoring:
             assert expect_properties_object_schema == _properties_object_schema
         else:
             assert properties_object_schema == _properties_object_schema
+
+
+def test_feature_is_exported_from_package_root() -> None:
+    import overture.schema.system as system
+
+    assert system.Feature is Feature
+    assert "Feature" in system.__all__


RoelBollens-TomTom (approved):

lowlydba (approved):
LGTM

